PR #3, opened 2016-09-11
Properly escape inline JSON as HTML

Rob--W:
Simply replacing `</` with `<\/` is not sufficient for escaping HTML, because `<script>` and `<!--` have a special meaning in the HTML5 parser too.

Besides that, not escaping `&` leads to edge cases where the result is incorrectly interpreted.

### DIFF
--- a/router/router.py
+++ b/router/router.py
@@ -412,7 +412,7 @@ class Handler(SimpleHTTPServer.SimpleHTTPRequestHandler):
             if 'json' in self.headers.getheader('Accept', ''):
                 self.generate(j, 'application/json')
             else:
-                j = j.replace("</", "<\\/")
+                j = j.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;")
                 template = os.path.join(index_path(tree_name), 'templates/search.html')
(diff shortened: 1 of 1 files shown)


Rob--W:
Here is an example where searchfox breaks: http://searchfox.org/mozilla-central/search?q=blackbox&path=

According to my analysis, the presence of `<!--`, `<script type=\"text/javascript\">` in the search results causes the HTML5 parser to skip over the closing "real" `</script>` tag.

This happens in both Firefox and Chrome, and is actually codified in the HTML5 spec: https://dev.w3.org/html5/spec-preview/the-script-element.html#restrictions-for-contents-of-script-elements

Rob--W:
@bill-mccloskey Ping.

bill-mccloskey:
Thanks Rob. Sorry this took me so long to merge. I found some problems with it but I didn't have time to look into them. The W3 spec you linked was very helpful in understanding the issue. Please look here for the solution I came to:
https://github.com/bill-mccloskey/searchfox/commit/1f5d61cccaf7e6221ebc39d5e966ab1a3696659c
The problem is that the data you're changing is not HTML. It's JS code (really JSON) that appears in an inline <script> element in the web page. So the escaping needs to be JS escaping rather than HTML entity escaping.

PR #5, opened 2017-01-19
Add C++ raw literals to the tokenizer

mrbkap:
This is a first draft to add C++ raw literals to the tokenizer. I couldn't get the test setup working, so I added some rust-style tests to the tokenizer module. I ran them via `cargo test`. I wasn't sure how much syntax checking I should be doing here, so this corresponds very closely to [this reference](http://en.cppreference.com/w/cpp/language/string_literal). If needed, I could pare this down to assume that someone else checked for correctness if that would be preferable.

Do I need to add a test for this in `LanguageSpec`? Also, any simplifications would be appreciated. I'm especially not terribly happy with how this plugs into the main if/else structure.

Thanks.

### DIFF
--- a/tools/src/tokenize.rs
+++ b/tools/src/tokenize.rs
@@ -102,7 +102,60 @@ pub fn tokenize_c_like(string: &String, spec: &LanguageSpec) -> Vec<Token> {
 
     while cur_pos.get() < chars.len() {
         let (start, ch) = get_char();
-        if is_ident(ch) {
+
+        if ch == 'R' && peek_char() == '"' {
+            // Handle raw literals per
+            // <http://en.cppreference.com/w/cpp/language/string_literal>.
+            get_char();
+
+            // Read the delimiter.
+            let paren;
+            loop {
+                let (idx, c) = get_char();
+                if c == '(' {
+                    paren = idx;
+                    break;
+                }
+
+                if peek_pos() == string.len() {
+                    writeln!(&mut std::io::stderr(), "Expecting '(' after raw string literal").unwrap();
+                    return tokens;
+                }
+            }
+
+            let delimiter = &string[start + 2 .. paren];
+            'raw_string: loop {
+                if peek_pos() == string.len() {
+                    writeln!(&mut std::io::stderr(), "Unterminated raw string").unwrap();
+                    return tokens;
+                }
+
+                let (_, next) = get_char();
+                if next != ')' {
+                    continue;
+                }
+
+                // Find the delimiter.
+                for c in delimiter.chars() {
+                    if c != peek_char() {
+                        continue 'raw_string;
+                    }
+                    get_char();
+                }
+
+                // Is this the end quote?
+                if peek_char() != '"' {
+                    continue;
+                }
+                get_char();
+
+                // Done!
+                break;
+            }
+
+            tokens.push(Token {start: start, end: peek_pos(), kind: TokenKind::StringLiteral});
+            next_token_maybe_regexp_literal = false;
+        } else if is_ident(ch) {
             while is_ident(peek_char()) {
                 get_char();
             }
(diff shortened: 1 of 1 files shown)


tools/src/tokenize.rs
```diff
@@ -266,7 +266,77 @@ pub fn tokenize_c_like(string: &String, spec: &LanguageSpec) -> Vec<Token> {
...
+                    break;
+                }
+            }
+
+            if peek_pos() == string.len() || peek_char() != '(' {
+                writeln!(&mut std::io::stderr(), "Expecting '(' after raw string literal").unwrap();
+                return tokens;
+            }
+
+            let sentinel = &string[start + 2 .. cur_pos.get()];
```
bill-mccloskey: I think you want to use peek_pos() instead of cur_pos.get(). cur_pos gives you the index into the chars vector. If the string contains multibyte characters, then chars and string won't line up.

Or you could use the result of get_char() as I described above.

tools/src/tokenize.rs
```diff
@@ -102,7 +102,7 @@ pub fn tokenize_c_like(string: &String, spec: &LanguageSpec) -> Vec<Token> {
 
     while cur_pos.get() < chars.len() {
         let (start, ch) = get_char();
-        if is_ident(ch) {
+        if is_ident(ch) && (ch != 'R' || peek_char() != '"') {
```
bill-mccloskey: Could you reverse the order here so that the raw literal thing came first? Then you could test:
if ch == 'R' and peek_char() == '"'
And then the ident case would just be:
else if is_ident(ch)

tools/src/tokenize.rs
```diff
@@ -266,7 +266,77 @@ pub fn tokenize_c_like(string: &String, spec: &LanguageSpec) -> Vec<Token> {
...
+
+            // Read delimiter.
+            loop {
+                let c = peek_char();
+                if c == '(' {
+                    break;
+                }
+
+                if c == ')' || c == '\\' || is_whitespace(c) {
+                    writeln!(&mut std::io::stderr(), "Illegal character in raw string delimiter").unwrap();
```
bill-mccloskey: Let's assume that the input is valid C++. We're already running clang on it before this step. Same for some of the other cases below.

tools/src/tokenize.rs
```diff
@@ -266,7 +266,77 @@ pub fn tokenize_c_like(string: &String, spec: &LanguageSpec) -> Vec<Token> {
...
+                    return tokens;
+                }
+
+                get_char();
```
bill-mccloskey: Instead of using peek_char followed by get_char, you might try saving both results of get_char. I think it might simplify the code in some cases. For example, I think it's fine to consume the '(' in this loop. You could save its position(the first result of get_char)  in a variable and then use that in the slice that you take below.

tools/src/tokenize.rs
```diff
@@ -266,7 +266,77 @@ pub fn tokenize_c_like(string: &String, spec: &LanguageSpec) -> Vec<Token> {
...
+            }
+
+            if peek_pos() == string.len() || peek_char() != '(' {
+                writeln!(&mut std::io::stderr(), "Expecting '(' after raw string literal").unwrap();
+                return tokens;
+            }
+
+            let sentinel = &string[start + 2 .. cur_pos.get()];
+
+            get_char();
```
bill-mccloskey: ...and if you consume the '(' in the previous loop, you wouldn't need this get_char().

tools/src/tokenize.rs
```diff
@@ -266,7 +266,77 @@ pub fn tokenize_c_like(string: &String, spec: &LanguageSpec) -> Vec<Token> {
...
+                }
+                get_char();
+
+                // Done!
+                break;
+            }
+
+            tokens.push(Token {start: start, end: peek_pos(), kind: TokenKind::StringLiteral});
+            next_token_maybe_regexp_literal = false;
+        } else if ch == '\'' || ch == '"' || ch == 'R' {
```
bill-mccloskey: I don't think you need the extra 'R' condition here?

tools/src/tokenize.rs
mrbkap: I removed most of the validity checks for the delimiter, but do protect against running off the end of the string as we do for regular string literals.

tools/src/tokenize.rs
```diff
@@ -102,7 +102,59 @@ pub fn tokenize_c_like(string: &String, spec: &LanguageSpec) -> Vec<Token> {
 
     while cur_pos.get() < chars.len() {
         let (start, ch) = get_char();
-        if is_ident(ch) {
+
+        if ch == 'R' && peek_char() == '"' {
+            // Handle raw literals.
```
bill-mccloskey: Can you add a link to that C++ reference site in the comment? Probably easy to google, but it's still nice to have.

PR #6, opened 2017-01-28
Make nested template strings in JS work

mrbkap:
This is a little clunky, but it seems to work. With this, we don't have to try to scan ahead looking for the end of the template string and we properly tokenize regexps and strings as to not be confused by them. This also handles nested functions or object literals that could also have caused confusion with the original code.


bill-mccloskey:
Thanks! Looks perfect.

PR #7, opened 2017-01-30
Newline fixes

mrbkap:
I was reading the code over the weekend and realized that my patch to add C++ raw literals didn't handle newlines properly, which seems like a bug that's worth fixing. In doing so, I noticed that the code to split a token across a newline was duplicated a couple of times so I pulled it out -- that being said, I'm a little surprised that I had to explicitly pass `tokens` in as a `&mut Vec<>`. Do you know why that's required?

I also fixed error reporting for illegal regexps ending with an escaped newline. It probably isn't a big deal, but we handle that case correctly in the string literal code, so I figured it was worth being consistent.


bill-mccloskey:
I merged this. Thanks!

PR #11, opened 2017-02-27
Have clang indexer emit comments, expose via new UI mega-menu.

asutherland:
This is probably not suitable for deployment yet, but I think it's a promising
start.

On Comments:
- We modify clang indexer to emit "briefComment" and "rawComment" JSON fields.
  Clang is capable of parsing the comments to a doxygen-like level of
  understanding, but we don't make use of that at this point.  The assumption
  is that the raw comments are human readable and this is a good first step.
- The JSON formatter in Indexer.cpp has been further hacked up to avoid
  exploding on non-printable, non-UTF8 characters that are checked into Gecko.
  This comes at the expense of also eating UTF8 characters.  Probably the best
  course of action is to just use a real JSON encoding library that is able to
  discard corrupt characters it sees in comments.
- In order to get clang to pay attention to non-explicit doxygen comments,
  we need to pass "-fparse-all-comments.  This gets and "-Xclang" prefix right
  now which it's not clear we need, but it works.
- IDL comments are not emitted, but that would be great.
- We modify the rust logic to propagate the "briefComment" and "rawComment"
  through to the crossref file.

On the mega-menu:
- This is along the lines of what I proposed in
  https://groups.google.com/d/msg/mozilla.dev.platform/nOLNG8RwyQE/BQtaa4nsCAAJ
  Specifically, it modifies the context menu to have detailed submenus to
  display comments or inline search results.
- The jquery.menu-aim.js library is introduced to avoid accidentally
  activating a different submenu when you are moving to focus on the submenu.
  See the library overview and docs at
  https://github.com/kamens/jQuery-menu-aim for more context on how this works.
- The "declaration" jump menu item now shows the declarations/definitions'
  associated comments for the symbol in the submenu.  If there are no
  comments, the traditional line is shown.
- A hack is introduced in the UI to reintroduce the declaration menu (with a
  different string) if there is no "jump" entry for the symbol because the
  user is already at the declaration.  This is necessary to help provide
  comment context from a header when looking at the C++ implementation.
- A hack is introduced to try and filter out forward declarations from the
  forward list.  I think it might be preferable to give forward declarations
  their own kind.
- The "search" jump menu now shows the uses for the symbol if there are only
  a few of them.  Display for this could likely be improved.  If there are
  too many results, an even more poorly formatted display just says how many
  uses there are.  My gameplan for that was to build a hierarchical tree
  that summarizes use counts by directory, where clicking on a directory
  goes to the search page for that symbol with the directory filter engaged.
- These subpages get their info by doing an AJAX "symbol:NAME" search.  An
  attempt is made to prefetch the information by leveraging the existing
  hover logic to send the AJAX lookup ahead of time.  Results are cached in
  a Map using Promises.  caniuse.com says these are generally fine, but given
  that the code was already using destructing assignment and template literals,
  I'm not sure it matters.  I am interested in knowing what the targeted
  support level is.
- The nunjucks library, previously used for very minimal templating of the
  context menu, is now removed in favor of largely vanilla JS DOM creation.
  The main justification is that the symbol searches are async and nunjucks
  didn't seem particularly suited to async updates of the DOM tree.  (Although
  it can handle asynchornous data which delays the production of the result.)

I've really only tested this with C++ code.  It's possible bad/horrible things
happen with JS code.  Certainly nothing useful happens for them.


asutherland:
Here are some hopefully self-explanatory screenshots of what the results look like in action:
* https://clicky.visophyte.org/examples/searchfox/20170227/static-field-decl-comments.png
* https://clicky.visophyte.org/examples/searchfox/20170227/static-field-uses.png
* https://clicky.visophyte.org/examples/searchfox/20170227/method-def-decl-comment.png

bill-mccloskey:
Sorry it's taking me so long to get to this. It looks really cool! I should be able to take a look at it this weekend. In the meantime I'll push it to the dev site.

bill-mccloskey:
Just pushed to the dev site. If it works, you should see it at http://dev.searchfox.org in a few hours.

bill-mccloskey:
This is a really neat feature, and the UI is actually pretty smooth. I would like some improvements before it lands:
* The text it displays seems inconsistent. When there's a comment, it shows only the comment. When there's no comment, it seems to show one line of the declaration. I think we should show any comment plus all lines from the declaration (so that parameters that appear on succeeding lines are still shown).
* If a comment is shown, the indentation of all the lines except the first one is incorrect. It would be nice to strip the same amount of indentation from each line if that's possible.
* If there's no comment attached to the definition, I think it makes sense to show only the declaration.
* Sometimes I don't see anything in the right-side box. As an example, if I click on mTimeout at [1], the definitions box is empty.
* I'm a little skeptical of how useful the "uses" are that appear with "Search for ...". Most of the time it just says there are too many. The layout is kind of confusing, especially when there are multiple from the same file. I do think we can do something useful here, but it will take some work. I suggest focusing on the definitions for now and we can iterate on showing uses once the definitions part is landed.

The fact that the right side box is often partially obscured is also an issue. I wonder if there's a way that we could put the content in the same box as the context menu?

I can help out with this stuff. Just let me know which parts you'd like to work on and I can give some of the other stuff a shot.

[1] http://dev.searchfox.org/mozilla-central/source/dom/base/nsGlobalWindow.cpp#825

asutherland:
Thanks very much for the feedback!  I agree on all points.  I'd like to do a cleanup pass tonight which should address many of the points you raise, and then hand things off to you until at least mid-next-week as I focus on something I need to get up for review.

In this patch I tried to leave the existing menu unchanged.  If you ignored the stuff in the sub-menu, there was no difference.  But as your comments make clear, this leads to an inconsistent experience.  It's not helpful if the user has to actively read the results to figure out if there was useful information or not.

I think with some menu reconfiguration I can provide menu items with a consistent experience that are biased to allow [pre-attentive processing](https://en.wikipedia.org/wiki/Data_visualization#Visual_perception_and_data_visualization) to figure out if there are useful bonuses like comments without having to read the actual text.  (This is easier for the raw comments with the *'s and //'s, of course.)

asutherland:
I was overly optimistic at my cleanup pass scheduling and progress.  I've pushed the progress I've made.  I'm planning to do another enhancement/cleanup pass on the UI changes I've made tomorrow night.  Specifically, I want to:
* Handle overrides better.  Right now they're collapsed together in a lossy way that was fine for the search UI but is not useful for the more inspector-like behavior I'm using the menu for.
* Turn "search"/"uses" back on.  I've revised my previous plan of "1 menu item with its submenu doing hierarchical clustering" to be "use hierarchical clustering to generate multiple menu items", such that there could be "same file", "same directory" context-aware menu items, then split out the uses by tree hierarchy.

One thing I currently plan to punt on is extracting the entire declaration block.  Right now the code just uses "line" from the crossref file.  Ideally it would grow an "allLines" or "lines" entry as well.  If you're interested in picking that up, that would be great.  I can also get to it later if needed; I'm beginning to understand a lot more of what's going on in the rust code.

bill-mccloskey:
I did some work on this, mostly on getting the lines shown to include all function parameters. In doing so, I changed how the code works. Rather than have the C++ code store the actual line data, it just stores a range of line numbers. I prefer this because:
- It keeps the analysis data smaller.
- We can unify handling of comments and stuff, which I suspect will save effort.
- In particular, I want to do syntax highlighting of the excerpted lines. I think this will be easier if we just include a range of line numbers.
- I don't really expect to be able to use clang's Doxygen parsing code. I was thinking through how that would actually be helpful, and I can't really think of anything that important.

I'll need to fix up how the IDL comments are handled. Besides modifying it to handle the peekRange changes, only handling Doxygen-style comments isn't really sufficient. Fixing this probably will necessitate changing the IDL parser, but I think that's okay.

I also think we should try to be conservative in changing the UI. The less the UI changes, the easier it will be to land this. And I'd like to land it soon. Also, it's very important to me that the "Go to definition" option is very easy to access. Once this lands, we can explore further changes to the context menu.

The new changes are in the mega-menu branch of my repo.

asutherland:
Awesome!

* I agree line ranges are preferable, especially with syntax-highlighting.
* Similarly, I agree about the doxygen parsing code.  In previous tooling benders I attempted to provide structured understanding of complex object arguments, etc.  Although it was very neat to be able to provide syntax-highlighted code where you could click on method call arguments and understand what the specific sub-object's fields meant, the useful granularity for understanding what was happening was at the method level and higher level documentation.  At this point, I view the more detailed markup as a standardized convention to aid in reading documentation and/or something for type-analysis tools to check rather than a useful feature.

I'm absolutely agreed about minimizing UI changes.  There's an unfortunate collision here between some actually useful functionality and UI and use-case noodling on my part.  My poorly time-estimated statements about "uses" are mainly to convey that the pushes I push aren't actually useful but I have an idea about how they could be useful.  I still want to demonstrate such a thing, but it absolutely should not delay you extracting something useful[1].

I'm somewhat time-limited this week, but if you think there's a clear/easy way to modify your branch to avoid the UI feeling different, I'm happy to make the changes.  At the same time, rest assured I will not feel slighted if you see a clear path to making the changes yourself.  Searchfox is an amazing tool and your design senses are clearly a critical part of that.  (And I am very aware that I tend to try and over-feature tooling in the hope I'll be able to leverage that in the future.)

1: Noting that I do want to provide useful patches that can land as-is and not force you to extract something useful out.  Your comments here are invaluable in helping me better understand your vision so I can do so in the future.

asutherland:
Closing this as I think:
- The core idea was cleaned up and landed by :billm as target record `peekRange` and corresponding cross-ref accumulation of `peekLines`.
  - I removed `peekLines`, restoring it to propagating `peekRange` in my changes for bug 1727453 in https://github.com/mozsearch/mozsearch/pull/462 in agreement with the idea here that we can use peekLines to excerpt the already-rendered HTML (although we may have to do a variety of performance work depending on how that pans out).
- The UI prototyping here was interesting but superseded by the fancy branch and in particular the current idea is to build on :kats' proposed super nav panel, example: https://mozilla.staktrace.com/tmp/mockup.html
  - A general concern has been that any UI like was prototyped here where large amounts of content are displayed in menus in the vicinity of the click will tend to obscure what the user actually wants to look at.

PR #14, opened 2017-07-24
Add link to Raw file

mozfreddyb:
This changeset adds a link to the raw file on hg.mozilla.org. Comes in handy for SVG files, for example.

### DIFF
--- a/tools/src/bin/output-file.rs
+++ b/tools/src/bin/output-file.rs
@@ -136,6 +136,10 @@ fn main() {
                     title: "Log".to_owned(),
                     link: format!("https://hg.mozilla.org/mozilla-central/log/tip/{}", path),
                     update_link_lineno: false,
+                 }, PanelItem {
+                    title: "Raw".to_owned(),
+                    link: format!("https://hg.mozilla.org/mozilla-central/raw-file/{}/{}", oid, path),
+                    update_link_lineno: false,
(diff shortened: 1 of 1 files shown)


tools/src/bin/output-file.rs
```diff
@@ -136,6 +136,10 @@ fn main() {
                     title: "Log".to_owned(),
                     link: format!("https://hg.mozilla.org/mozilla-central/log/tip/{}", path),
                     update_link_lineno: false,
+                 }, PanelItem {
+                    title: "Raw".to_owned(),
+                    link: format!("https://hg.mozilla.org/mozilla-central/raw-file/{}/{}", oid, path),
```
bill-mccloskey: The oid here is a git revision. You might be able to link to the raw file on github (in the gecko-dev repo). Otherwise, you could link to tip (as the "Log" link does) or figure out a way to convert the Hg revision to a Git revision (which is possible to do with an HTTP API).

mozfreddyb:
Thanks for the timely review. I've switched to GitHub for the file URL, as it's much easier and provides the very same functionality.

(If you strongly prefer using hg.m.o, I'd need significant help issuing the HTTP requests to figure out the Mercurial revision id. As this is my first pull request to a rust project.)

bill-mccloskey:
Thanks!

PR #15, opened 2017-08-01
tools: Remove unneeded indirections in a few places.

emilio:
There's never a need to use `&Vec<T>` over `&[T]`, since they're always
immutable.

Similarly, there's never a reason for using `&String` over `&str`.


bill-mccloskey:
Thanks!

PR #17, opened 2017-10-17
Add syntax highlighting for java files.

thomcc:
We have a decent amount of Java code in the tree and it's a bummer that searchfox doesn't syntax highlight it. It seemed easy enough to add, so I did that.

I was going to wait until I tested this to PR, but, well, I can't actually do that, since the vagrant build seems to be broken (due to livegrep failing to build? I had to add `--incompatible_disallow_set_constructor=false` [here](https://github.com/bill-mccloskey/searchfox/blob/master/infrastructure/indexer-provision.sh#L46), which made it get a bit further, but it still failed complaining but I guess it couldn't install libgit due to some issue with the checksum?).

I'm tcsc on moz irc if you need to ask me anything about this patch, or my issue with vagrant.

### DIFF
--- a/tools/src/languages.rs
+++ b/tools/src/languages.rs
@@ -130,6 +130,17 @@ lazy_static! {
         "default", "union",
     ];
 
+    static ref RESERVED_WORDS_JAVA : Vec<&'static str> = vec![
+        "abstract", "continue", "for", "new", "switch", "assert", "default",
+        "goto", "package", "synchronized", "boolean", "do", "if", "private",
+        "this", "break", "double", "implements", "protected", "throw", "byte",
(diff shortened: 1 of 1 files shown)


mrbkap:
It might be easier to test this by adding tests to the [bottom of tokenize.rs](https://github.com/bill-mccloskey/searchfox/blob/0a3572cbd1c70c4572d0ddbceac6e2256e39d59f/tools/src/tokenize.rs#L915). Difficulties in getting this set up locally was my motivation for writing those (also the fix/compile/test cycle is much easier with those tests!).

bill-mccloskey:
Thanks Thom. I tested your PR and it seems to work. It should go live tomorrow.

I also fixed up the Vagrantfile. Sorry about that.

PR #23, opened 2017-11-09
Bug 1397801 - In VisitCallExpr, wait until we get the real location o…

amccreight:
…f the callee to decide if it is interesting.

The start of a declaration of a thing being called can end up being
the return type, such as bool. When it is a primitive type like bool,
then the normalized location ends up being uninteresting, so nothing
ends up being reported for the call.

(billm, mrbkap and I worked together to figure this out.)

### DIFF
--- a/clang-plugin/Indexer.cpp
+++ b/clang-plugin/Indexer.cpp
@@ -1188,12 +1188,7 @@ public:
 
     const NamedDecl *namedCallee = dyn_cast<NamedDecl>(callee);
 
-    SourceLocation startLoc = callee->getLocStart();
-    SourceLocation loc = startLoc;
-    NormalizeLocation(&loc);
-    if (!IsInterestingLocation(loc)) {
-      return true;
-    }
+    SourceLocation loc;
 
     const FunctionDecl *f = dyn_cast<FunctionDecl>(namedCallee);
     if (f->isTemplateInstantiation()) {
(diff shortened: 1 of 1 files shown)


clang-plugin/Indexer.cpp
```diff
@@ -1208,20 +1203,25 @@ class IndexConsumer : public ASTConsumer,
...
     } else if (MemberExpr::classof(calleeExpr)) {
       MemberExpr* member = dyn_cast<MemberExpr>(calleeExpr);
       loc = member->getMemberLoc();
-      NormalizeLocation(&loc);
     } else if (DeclRefExpr::classof(calleeExpr)) {
       // We handle this in VisitDeclRefExpr.
       return true;
     } else {
-      if (callee->getLocEnd() != startLoc) {
+      if (callee->getLocEnd() != callee->getLocStart()) {
```
emilio: Shouldn't this be `==`? We're skipping if `end != start`, while I'd expect to skip when the start and end are the same.

bill-mccloskey: I decided to pull out this whole branch. I don't think any of it makes any sense anyway. It's testing |callee|, which is the callee declaration. It would only make sense to look at calleeExpr.

bill-mccloskey:
I merged this with the change mentioned above.

PR #24, opened 2017-11-09
Use HSTS in release

bgrins:
Unfortunately it seems we need to add this header manually for all locations, at least based on https://www.nginx.com/blog/http-strict-transport-security-hsts-and-nginx/.

We should test this on dev.


bill-mccloskey:
I pushed this to dev and it doesn't seem to work. I think the problem is this:
https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/

I'm not really sure how to fix it. We might need to have separate configurations for local testing and running on AWS.

bill-mccloskey:
This landed. I had to pass a special "hsts" param when we're running from AWS.

PR #27, opened 2017-11-14
Reduce memory usage in crossref.rs by sharing strings

amccreight:
(This includes my patch for bug 1416876 because it is based on top of that and I don't know how to do that the right way.)

I did some light local testing, all the way up to running a webserver for the results of a full Firefox run and the few things I checked seemed ok. The output from this file was also unchanged, when I checked it for the small set of test files.

### DIFF
--- a/tools/src/bin/crossref.rs
+++ b/tools/src/bin/crossref.rs
@@ -75,9 +75,9 @@ fn main() {
     let jump_file = format!("{}/jumps", tree_config.paths.index_path);
     let id_file = format!("{}/identifiers", tree_config.paths.index_path);
 
-    let mut table = HashMap::new();
+    let mut table = BTreeMap::new();
     let mut pretty_table = HashMap::new();
-    let mut id_table = HashMap::new();
+    let mut id_table = BTreeMap::new();
     let mut jumps = Vec::new();
 
     {
--- a/tools/src/bin/crossref.rs
+++ b/tools/src/bin/crossref.rs
@@ -82,70 +82,68 @@ fn main() {
     let mut id_table = BTreeMap::new();
     let mut jumps = Vec::new();
 
-    {
-        for path in &file_paths {
-            print!("File {}\n", path);
+    for path in &file_paths {
+        print!("File {}\n", path);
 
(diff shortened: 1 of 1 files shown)


tools/src/bin/crossref.rs
```diff
@@ -61,6 +64,28 @@ fn split_scopes(id: &str) -> Vec<String> {
     return result;
 }
 
+struct StringIntern {
+    set: HashMap<Rc<String>, ()>
```
emilio: nit: Isn't this basically a `HashSet<Rc<String>>`?

emilio:
(We could also consider using https://github.com/servo/string-cache, instead of a hand-rolled one)

tools/src/bin/crossref.rs
amccreight: As far as I can tell, HashSet<> doesn't support the Entry interface, or anything like that. I want to use the existing value if there is one, insert our new value if there isn't, and avoid repeated lookups. I guess that's overoptimizing. I could use a HashSet and just do two lookups.

amccreight:
I've looked at the Servo string cache before, but I think it is threadsafe and it was slower than the trivial thing when I tried it before. I also think just writing a tiny 6 line function is simpler than introducing a new dependency with its own API people would have to understand.

tools/src/bin/crossref.rs
emilio: Well, that's fair... Sounds ok :)

emilio:
> I've looked at the Servo string cache before, but I think it is threadsafe and it was slower than the trivial thing when I tried it before. I also think just writing a tiny 6 line function is simpler than introducing a new dependency with its own API people would have to understand.

I mean, if we're really going to uber-optimize we can also stop using siphash and all that stuff, but yeah, this is fine, sorry for the contextless review, and thanks for saving my laptop's memory ;)

amccreight:
I actually did write a patch to replace the hash function in the string interner, but there was no impact on performance. I vaguely recall reading that the default hash function is only slow on small keys, and we're using strings, so maybe it isn't an issue? Anyways, I should just profile this thing to see if there are some easy wins in terms of speed.

PR #29, opened 2017-11-14
Use hg.m.o's raw-file viewer instead of github, so that it renders th…

staktrace:
…ings like SVGs and HTML files

(This patch is untested, since the nav panel only seems to show up on m-c which I haven't indexed locally).

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1413035

### DIFF
--- a/tools/src/bin/output-file.rs
+++ b/tools/src/bin/output-file.rs
@@ -129,6 +129,10 @@ fn main() {
         let panel = if path.contains("__GENERATED__") {
             vec![]
         } else if let Some(oid) = head_oid {
+            let hg_rev : &str = tree_config.git.as_ref()
+                .and_then(|git| git.hg_map.get(&oid))
+                .and_then(|rev| Some(rev.as_ref())) // &String to &str conversion
+                .unwrap_or(&"tip");
             vec![PanelSection {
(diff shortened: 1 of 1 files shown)


bill-mccloskey:
Unfortunately, the OID we have is from git, so it won't work in hg. Let's talk today about how to do this.

staktrace:
I just pointed it to /tip/ rather than using the oid, since the "raw" link only appears when you're using the tip version anyway (when viewing a permalink there's no raw link). But yeah we should talk about this more, I'd like to make the navigation panel appear in the test repos as well so this kind of change can be exercised more easily.

staktrace:
Updated patch to use the hg_map to pull the hg revision from the git oid, and only fall back to tip if we couldn't find it. Note that I couldn't test this change properly because the test repos don't have corresponding blame repos. I tried generating one but ran into python problems...

bill-mccloskey:
I merged a version of this change as part of https://github.com/bill-mccloskey/mozsearch/commit/18355d19721f3959e4f42a59960a2bd41fe7c4c4. I realized that output-file.rs always generates output for tip anyway. It only makes sense to have the raw link for old revisions link back to old hg revisions, so I moved the hg_rev code there.

PR #30, opened 2017-11-14
Make the blame repo optional.

staktrace:
This makes it so that we can specify a git repo in the config.json
without a blame repo, and get the partial functionality that unlocks.
In particular this enables the navigation panel if the tree has a git
repo specified, even if it doesn't have the blame repo.
This also adds the searchfox git repo to the tests/config.json file, so
now using the test repos it's possible to exercise the navigation panel.

Note that the links in the navigation panel are still hard-coded to
mozilla-central/gecko-dev which is a separate pre-existing issue (happens
on e.g. the nss repo on searchfox.org as well).

### DIFF
--- a/tests/config.json
+++ b/tests/config.json
@@ -13,6 +13,7 @@
       "index_path": "$WORKING/searchfox",
       "files_path": "$WORKING/searchfox/files",
       "objdir_path": "$WORKING/searchfox/objdir",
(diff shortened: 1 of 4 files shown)


staktrace:
There might be more rust-idiomatic ways to express this; my rust-fu isn't that great. Feel free to be nitpicky on style and such.

PR #31, opened 2017-11-14
Allow the config.json to specify the hg root

staktrace:
This allows a repo to override the hg URL root so that the log/raw links point to the right repos. Combined with a patch to the mozsearch-mozilla repo (will make a PR on that as well) this should fix https://bugzilla.mozilla.org/show_bug.cgi?id=1325043


staktrace:
(Note that this applies on top of #29)

bill-mccloskey:
I merged this change as part of https://github.com/bill-mccloskey/mozsearch/commit/18355d19721f3959e4f42a59960a2bd41fe7c4c4.

PR #43, opened 2017-11-16
Bug 1417945 - Linkify URLs in comments

staktrace:


tools/src/format.rs
```diff
@@ -18,9 +18,28 @@ use git2;
 use chrono::naive::datetime::NaiveDateTime;
 use chrono::offset::fixed::FixedOffset;
 use chrono::datetime::DateTime;
+use linkify::{LinkFinder, LinkKind};
 
 use config;
 
+fn linkify(s: String) -> String {
```
bill-mccloskey: It might be nice to have this in its own module along with the linkify_commit_header function from blame.rs. Maybe we could combine those two together as well into a single linkification function.

bill-mccloskey:
That could be a separate PR though.

staktrace:
Makes sense, I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1418072 for it so I don't forget to do it.

PR #51, opened 2017-11-18
Remove setversion.js since the version function has been removed from…

staktrace:
… SpiderMonkey as of mozilla bug 1417844


staktrace:
I'm re-running the indexer now with this change, will merge it if all goes well.

staktrace:
The indexer run worked, so I'm merging this. I'd still like @bill-mccloskey or somebody else to review it to make sure I'm not missing something.

PR #53, opened 2017-11-19
Don't cross-reference local JS variables

bill-mccloskey:
This change removes the ability to search for local JS variables. This should reduce the load on the cross-referencer and make it more likely that we'll be able to index add-on JS. It will also bring the JS indexer in line with the C++ one. Note that globals variables are treated as properties, so they'll still be indexed.

### DIFF
--- a/js-analyze.js
+++ b/js-analyze.js
@@ -139,14 +139,18 @@ let Analyzer = {
     }
   },
 
-  source(loc, name, syntax, pretty, sym) {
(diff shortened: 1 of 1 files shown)


staktrace:
Do you have a reviewer in mind for this?

bill-mccloskey:
I was going to ask Kris, but he hasn't joined the GitHub team. Feel free to review.

PR #55, opened 2017-11-19
Add link to switch between .h and .cpp files

bill-mccloskey:
This is a nice feature that DXR has. It exposes an item in the navigation panel to quickly switch between .h and .cpp files.

### DIFF
--- a/tools/src/bin/output-file.rs
+++ b/tools/src/bin/output-file.rs
@@ -1,4 +1,6 @@
+use std::collections::HashMap;
 use std::env;
+use std::ffi::OsStr;
 use std::fs;
 use std::fs::File;
 use std::io;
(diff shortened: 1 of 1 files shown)


tools/src/bin/output-file.rs
```diff
@@ -126,7 +128,46 @@ fn main() {
             }
         }
 
-        let panel = if path.contains("__GENERATED__") {
+        let mut extension_mapping = HashMap::new();
+        extension_mapping.insert("cpp", ("header", vec!["h", "hh"]));
```
staktrace: Do we want to add .hpp here as well? We only have a couple in mozilla-central so maybe not worth it.

tools/src/bin/output-file.rs
```diff
@@ -126,7 +128,46 @@ fn main() {
...
+
+        let extension = path_wrapper.extension().unwrap_or(&OsStr::new("")).to_str().unwrap();
+        let show_header = match extension_mapping.get(extension) {
+            Some(&(ref description, ref try_extensions)) => {
+                let mut result = None;
+                for try_ext in try_extensions {
+                    let try_buf = path_wrapper.with_extension(try_ext);
+                    let try_path = try_buf.as_path();
+                    if try_path.exists() {
+                        let (path_sec, _) = path.split_at(path.len() - extension.len() - 1);
```
staktrace: Not clear to me what _sec stands for. I would have called it path_dir or path_base or something.

staktrace:
Couple of minor comments, feel free to merge with or without addressing them :)

PR #58, opened 2017-11-19
Add sticky highlight support

bill-mccloskey:
This patch allows you to click on an identifier and then select "Sticky highlight" from the context menu. The context menu disappears and the identifier stays highlighted until you click somewhere. This is nice to find all the uses of a local variable in a large function.

### DIFF
--- a/static/js/context-menu.js
+++ b/static/js/context-menu.js
@@ -29,8 +29,12 @@ function setContextMenu(menu, event)
   });
 }
 
+// When this is set to true, moving the mouse doesn't change what is highlighted.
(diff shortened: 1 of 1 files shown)


staktrace:
I updated this PR to account for the review comments by @asutherland - applying those comments exactly as-is didn't work since we need the original `elt` later to find the closest element with a `data-id` attribute, which may not be the same as the closest element with a `data-i` attribute. The changes were pretty minor though, and since the rest of the patch was reviewed as ok, I'll go ahead and merge it.

PR #60, opened 2017-11-19
Add support for "Go to definition of IDL ..."

bill-mccloskey:
One annoying weakness of Searchfox is as follows. Suppose a method foo() is defined in an IDL file nsIBar. If some C++ code calls bar->Foo() and you click on Foo, the context menu will have "Go to definition" and "Search" options. The "go to definition" item will take you to a C++ implementation of Foo, if there is one. The "search" option will show you the XPIDL line that defines foo(), as well as any C++ implementations and uses. To find JS implementations and uses, you have to click on the IDL search result and then search again.

This patch improves things somewhat. It adds an option to jump directly to the IDL file from the C++ method call. From there, you can find JS implementations and uses. To do this, I had to expand the notion of a definition so that there can be one regular definition and one IDL definition.

Eventually it would be nice to have some sort of "See also" functionality to expose the JS code directly.

Note that this same code will work equally well for IPDL and WebIDL (once we support WebIDL in bug 1416899).

### DIFF
--- a/tools/src/bin/crossref.rs
+++ b/tools/src/bin/crossref.rs
@@ -23,6 +23,7 @@ struct SearchResult {
     lineno: u32,
     bounds: (u32, u32),
     line: Rc<String>,
+    pretty: Rc<String>,
     context: Rc<String>,
     contextsym: Rc<String>,
     peek_lines: Rc<String>,
@@ -110,7 +111,6 @@ fn main() {
     let empty_string = strings.add("".to_string());
 
     let mut table = BTreeMap::new();
-    let mut pretty_table = HashMap::new();
     let mut id_table = BTreeMap::new();
     let mut jumps = Vec::new();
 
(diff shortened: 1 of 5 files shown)


tools/src/bin/crossref.rs
```diff
@@ -225,21 +225,28 @@ fn main() {
...
-                        v.push(id.to_json());
-                        v.push(path.to_json());
-                        v.push(results[0].lineno.to_json());
-                        let pretty = pretty_table.get(&id).unwrap();
-                        v.push(pretty.to_json());
-                        jumps.push(Json::Array(v));
+        {
+            let mut add_jumps = |kind: &AnalysisKind, kindstr: &str| {
+                if id_data.contains_key(kind) {
+                    let defs = id_data.get(kind).unwrap();
```
staktrace: Instead of doing the contains_key check followed by get(..).unwrap(), you could combine them using the map function on Option:

```rust
let mut add_jumps = |kind: &AnalysisKind, kindstr: &str| {
    id_data.get(kind).map(|defs| {
        if defs.len() ...
    });
};

File: tools/src/format.rs [hunk 2]
```diff
@@ -129,13 +128,16 @@ pub fn format_code(jumps: &HashMap<String, Jump>, format: FormatAs,
...
                     for sym in syms {
                         match jumps.get(sym) {
-                            Some(jump) => {
-                                if !(&jump.path == path && jump.lineno == cur_line) {
-                                    let key = format!("{}:{}", jump.path, jump.lineno);
-                                    let mut obj = json::Object::new();
-                                    obj.insert("sym".to_string(), Json::String(sym.to_string()));
-                                    obj.insert("pretty".to_string(), Json::String(jump.pretty.clone()));
-                                    menu_jumps.insert(key, Json::Object(obj));
+                            Some(jump_list) => {
```
staktrace: pre-existing, but instead of using a match on jumps.get(sym) you can use the `if let` construct to reduce one level of nesting:

`if let Some(jump_list) = jumps.get(sym) {`

asutherland:
I think this ended up effectively being duplicated by https://bugzilla.mozilla.org/show_bug.cgi?id=1758394 and https://bugzilla.mozilla.org/show_bug.cgi?id=1759064.  The mechanism for supporting multiple definitions would potentially still stand, but I am planning to move us from the per-location array of [jumps, searches] stored in ANALYSIS_DATA in each HTML file to a SYMBOL_MAP (name still TBD) that contains the structured data for each symbol observed in the file including the defs/decls/idl hit-lists (or similar) now that output-file can directly access the crossref database from rust.

PR #63, opened 2017-11-21
Don't allow VM to use more than half of system memory.

staktrace:
For those of us trying to run on 8GB macbook air devices giving the VM
10G locks up the machine and forces a reboot :(

This patch keeps the 10G desired memory amount, but clamps it to no more than half of system memory.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -10,7 +10,20 @@ Vagrant.configure("2") do |config|
   config.vm.network :forwarded_port, guest: 80, host: 8000
 
   config.vm.provider "virtualbox" do |v|
-    v.memory = 10000
+    host = RbConfig::CONFIG['host_os']
+    if host =~ /darwin/
+      # sysctl returns Bytes and we need to convert to MB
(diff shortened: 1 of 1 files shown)


amccreight:
Are you able to successfully run the analysis with this? When I had the VM memory set to 8GB, the JS analysis script kept randomly failing with invalid parse errors, which I attributed to it running out of memory.

staktrace:
I can run the analysis on the test repo, I haven't tried running it on m-c. But without this I couldn't even get to that point, `vagrant up` would lock up my machine completely.

amccreight:
Could you change it to something like 62% of system memory rather than 50%? That'll keep it at 10GB on 16GB machines. With your patch as it is here, I'll have to revert it locally to be able to handle an analysis of m-c.

staktrace:
I tried at 75% and it still locked up my system. I don't really want to try 62% as well, I'd rather just leave it as-is and set it up on a machine with more memory instead. I'll close this issue.

PR #65, opened 2017-11-29
Remove symlinks as they don't work well on a Windows host

staktrace:
Developing mozsearch locally on a Windows host doesn't work well because
the repository contains symlinks and the Windows checkout of the symlinks
may not work properly (at least by default). Since all the symlinks
point to the same script, we can just put that script in the scripts/ folder
and use it from there. The script is also updated to incorporate changes
from the mozsearch-mozilla repo, as this script will now be used to process
all trees.

### DIFF
--- a/tests/searchfox/find-repo-files
+++ b/tests/searchfox/find-repo-files
@@ -1,1 +1,3 @@
-../find-repo-files.py
+#!/usr/bin/env bash
+
+exec $(dirname $0)/../find-repo-files.py $*
--- a/tests/tests/find-repo-files
+++ b/tests/tests/find-repo-files
@@ -1,1 +1,3 @@
-../find-repo-files.py
+#!/usr/bin/env bash
+
+exec $(dirname $0)/../find-repo-files.py $*


staktrace:
An alternative approach would be to turn the symlinks into real files that exec the .py script. But I figured this is probably simpler for now, as we don't seem to be making use of the have-a-different-script-per-repo capability anyway.

I'll submit a patch to the mozsearch-mozilla config repo as well to remove the symlinks there, if we merge this one. (Will wait a day to make sure this is deployed before I do that)

staktrace:
@bill-mccloskey Any objections to removing these symlinks from the repo?

bill-mccloskey:
Sorry, I lost track of this. I guess this is okay. The exec approach definitely seems better though.

staktrace:
Ok, I updated it to use `exec` instead.

staktrace:
Hmm the patch doesn't look right. Fixing...

staktrace:
Ok, should be good now. I also opened https://github.com/mozsearch/mozsearch-mozilla/pull/2 to do the equivalent in the mozsearch-mozilla repo.

PR #66, opened 2017-11-29
Fix visiting of constructor expressions

staktrace:
This is either the dumbest API, or the worst documented API. Or both.

### DIFF
--- a/clang-plugin/MozsearchIndexer.cpp
+++ b/clang-plugin/MozsearchIndexer.cpp
@@ -500,7 +500,7 @@ public:
     AutoSetContext Asc(this, D);
     return Super::TraverseRecordDecl(D);
   }
-  bool TraverseCxxRecordDecl(CXXRecordDecl *D) {
+  bool TraverseCXXRecordDecl(CXXRecordDecl *D) {
     AutoSetContext Asc(this, D);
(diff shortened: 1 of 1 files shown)


staktrace:
There's a few other places in the file we use Cxx where we probably should be using CXX but I'm not sure what impact those actually have so I didn't change them yet. It's not clear to me what the TraverseCxx* functions actually do.

emilio:
(Feel free to wait for billm if you want feedback on the rest of TraverseCxx and VisitCxx stuff... It's certainly worrying.)

At least the ones that call to `Super::` should probably be audited, and see if it works, and why.

bill-mccloskey:
Ugh. This got broken when I used clang-tidy to rewrite all the variable names to match LLVM style. All of the Cxx occurrences should be replaced with CXX.

This is the best description of how the Traverse/Visit stuff works:
https://clang.llvm.org/doxygen/classclang_1_1RecursiveASTVisitor.html#details

The purpose of the Traverse methods is to allow us to have an AutoSetContext on the stack while the Visit functions run. The Visit functions do most of the work.

staktrace:
Ok, I updated the PR to use CXX everywhere that Cxx was used before.

PR #68, opened 2017-12-01
Bug 1417946 - Linkify bug numbers in comments

staktrace:


tools/src/links.rs
```diff
@@ -12,26 +12,33 @@ pub fn linkify(s: String) -> String {
...
     }
     if last == 0 {
-        s
+        result = s;
     } else {
         result.push_str(&s[last ..]);
-        result
     }
+
+    linkify_bug_numbers(&result)
```
emilio: Won't this double-linkify bug numbers?

That is, if I have a commit message like:

```
Foo bar https://bugzilla.mozilla.org/show_bug.cgi?id=1413619
```

Wouldn't this linkify first the bug, and _then_ the bug number again, creating nested links and what not?

Maybe this is good enough though...

staktrace: @emilio The commit message linkification should be unchanged, I just extracted the bug number regex into a separate function for reuse. What you said does apply to `linkify_comment` though; I'll update the regex to also look for the word "bug" which should take care of this.

emilio: Right, sorry for the confusion, I meant `linkify_comment` ofc. Yeah, probably at least linkify_comment should look for `bug XXX`. For commit messages it may be tricky because people use `b=XXX`, `Bug: XXX` and different patterns.

staktrace:
Updated the commit to not mess with the commit headers, and use a new regex "bug <optional whitespace> <number>" (case insensitive) to linkify bug numbers.

tools/src/links.rs
```diff
@@ -12,11 +12,21 @@ pub fn linkify(s: String) -> String {
...
+}
+
+pub fn linkify_bug_numbers(s: &str) -> String {
+    lazy_static! {
+        static ref BUG_NUMBER_REGEX: Regex = {
+            Regex::new(r"\b(?i)bug\s*(?P<bugno>[1-9][0-9]{2,6})\b").unwrap()
```
emilio: A comment here about why the regex is not just the number would be nice I think. We can always make the regex smarter in the future if we find uses with different patterns.

emilio: Or alternatively, we could just call `linkify_bug_numbers` during the same linkify phase, like:

```diff
diff --git a/tools/src/links.rs b/tools/src/links.rs
index 3063cef..7d66410 100644
--- a/tools/src/links.rs
+++ b/tools/src/links.rs
@@ -7,14 +7,14 @@ pub fn linkify(s: String) -> String {
     let mut last = 0;
     let mut result = String::new();
     for link in finder.links(&s) {
-        result.push_str(&s[last .. link.start()]);
+        result.push_str(&linkify_bug_numbers(&s[last .. link.start()]));
         result.push_str(&format!("<a href=\"{}\">{}</a>", link.as_str(), link.as_str()));
         last = link.end();
     }
     if last == 0 {
-        s
+        linkify_bug_number(&s)
     } else {
-        result.push_str(&s[last ..]);
+        result.push_str(&linkify_bug_numbers(&s[last .. link.start()]));
         result
     }
 }
```

And keep the number replacement as-is.

staktrace: I tried something similar to that but it turns out a lot of comments have numbers that are not bug number references, so just looking for a number gets a lot of false positives. I'll add the comment like you suggested and we can look into adjusting the regex more in the future if we need to.

PR #69, opened 2017-12-04
Bug 1422729 - Produce an error when trying to diff a documentation file, instead of…

staktrace:
… panicking

### DIFF
--- a/tools/src/format.rs
+++ b/tools/src/format.rs
@@ -589,6 +589,9 @@ pub fn format_diff(cfg: &config::Config,
         FormatAs::Binary => {
             return Err("Cannot diff binary file");
         },
+        FormatAs::FormatDoc(_) => {
+            return Err("Cannot diff documentation file");
+        },
         _ => {},
     };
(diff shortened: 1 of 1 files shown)


staktrace:
See comment 1 and 2 in https://bugzilla.mozilla.org/show_bug.cgi?id=1422729 for details on what happened

asutherland:
I'm not sure how drive-bys work here, but r=asuth :)

staktrace:
drive-bys are welcome and gladly accepted :)

staktrace:
Also for the record I verified I could reproduce the problem by loading a /diff/<rev>/path/to/something.md URL in the searchfox repo itself (after I built a corresponding blame repo for it), and verified that the patch prevented the server from crashing.

I think in the normal course of using searchfox you can't actually get to that sort of a URL because .md and .rst files get rendered rather than showing blame, and so there's no way to get a /diff/ link to one of those files. That's probably why this hasn't gotten hit before, even though the bug has been in the code for a while.

PR #71, opened 2017-12-05
band-aid: return Err instead of panicking until we can investigate an…

staktrace:
…d fix the root cause


staktrace:
We seem to be hitting this panic a lot and the web-server keeps going down. I've had to restart it three times in the last 24 hours already. Changing it to a return Err so that it's just the single request that fails instead of the entire server going down. We can then investigate and fix it at our leisure.

staktrace:
Merging without review for now so that this patch catches the next indexing run and I managed to exercise the code path in my local setup to make sure it doesn't take down the server.

PR #74, opened 2017-12-23
Improve ergonomics around notifying when the indexing is done

staktrace:
When indexing fails this might help diagnose the failure faster.

r? @amccreight or anybody else


staktrace:
I added some more patches here to improve the flow when indexing is done. The failure email is now sent immediately, instead of waiting for the 6 hours to expire. Also when doing dev-channel indexing, the email is only sent to the author of the head commit in the repo. And for dev-channel indexing an email is sent on successful completion as well.

Note this will require updating the lambda.zip but I can do that after these patches have been reviewed and merged. Anybody is welcome to review, the patches should be easy to follow.

staktrace:
Thanks! I uploaded the new lambda.zip and am going to merge this PR.

staktrace:
Deployment seems to have worked fine, yay

PR #76, opened 2018-01-04
rust-indexer: Handle better trait function impls. 

emilio:
This allows finding all the trait implementations from a trait function with a
single click.

Before this, in the test code, for example, for the "fn do_bar" line in
"impl MyTrait for Loader" we get the following symbol:

```
  simple<Loader as test_rust_dependency::MyTrait>::do_bar
```

This is fine, but we also want to get:

```
  test_rust_dependency::MyTrait::do_bar
```

So we can relate all the impls of the same function for the trait.


emilio:
Thanks for the review!

PR #77, opened 2018-01-08
 Make rust and C agree on symbols for extern functions

emilio:
(By returning the unmangled name in both cases)

This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1428841.


emilio:
Sure thing :)

PR #80, opened 2018-01-19
blame: Work around broken submodule references.

emilio:


blame/transform-repo.py
```diff
@@ -84,11 +84,15 @@ def sanitize(s):
...
 def get_tree_data(repo, tree, path):
     for elt in path:
-        if elt in tree:
-            item = tree[elt]
-            tree = repo.get(item.id)
-        else:
+        # Tree should never be None, but Servo has a broken submodule entry in
+        # commit b2a5225831a8eee3ff596dce2be8dc08df4300a0 pointing to a wpt
+        # submodule, but without information about it in .gitmodules.
+        if not tree or elt not in tree:
```
staktrace: Might be a little clearer to put the "if not tree" part of the check at the start of the function, outside the "for elt in path" loop. But it's not a big deal, this is fine too.

emilio: It needs to be there. `tree` at the beginning of the function is never `None`, but we re-assign to it during the loop.

staktrace: Oh, of course. Thanks 👍

PR #81, opened 2018-01-20
rust-analyze: only stick with the deps/save-analysis file, instead of all of them.

emilio:
When building with crates.io, all the crate dependencies also get save-analysis
dirs.

We probably want to use them, but not quite yet.

### DIFF
--- a/scripts/mkindex.sh
+++ b/scripts/mkindex.sh
@@ -26,12 +26,12 @@ $MOZSEARCH_PATH/scripts/build.sh $CONFIG_REPO $CONFIG_FILE $TREE_NAME
 
 date
 
-$MOZSEARCH_PATH/scripts/find-objdir-files.py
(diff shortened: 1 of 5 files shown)


emilio:
This unblocks servo.

asutherland:
@emilio Is there anything from this patch we still want noting that we may want to switch to using LSIF for rust (https://bugzilla.mozilla.org/show_bug.cgi?id=1761287) and that I think I addressed the generated files issues in other patches (https://github.com/mozsearch/mozsearch/pull/299 and https://github.com/mozsearch/mozsearch/pull/339 maybe based on blame)?

Note that the rust stuff is not super high on my personal priority list right now, but the JS LSIF stuff is (https://bugzilla.mozilla.org/show_bug.cgi?id=1740290), and there's a good chance of there being synergies there.

PR #83, opened 2018-01-27
Bug 1432300 - Improve the heuristic to pick the symbol that will be a…

staktrace:
…ssociated with a token's highlight.

Instead of just picking the first symbol in the list, check to see
if any of the symbols for that token contain the token's user-visible
string representation. If so, use that token. Otherwise fall back to
the first one in the list.

### DIFF
--- a/tools/src/format.rs
+++ b/tools/src/format.rs
@@ -102,7 +102,17 @@ pub fn format_code(jumps: &HashMap<String, Jump>, format: FormatAs,
 
         let data = match (&token.kind, datum) {
             (&tokenize::TokenKind::Identifier(None), Some(d)) => {
-                let ref id = d[0].sym;
+                // In case the token has multiple symbols associated with it,
+                // try to pick the "best" one for determining which other tokens
+                // this gets highlighted with. For now we just take the first
+                // symbol that contains the token, or the first one in the list
+                // if none contain the token.
+                let id = if d.len() == 1 {
+                    &d[0].sym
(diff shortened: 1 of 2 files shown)


tools/src/format.rs
```diff
@@ -102,7 +102,12 @@ pub fn format_code(jumps: &HashMap<String, Jump>, format: FormatAs,
 
         let data = match (&token.kind, datum) {
             (&tokenize::TokenKind::Identifier(None), Some(d)) => {
-                let ref id = d[0].sym;
+                let id = if d.len() == 1 {
```
emilio: Maybe add a comment of what you're trying to do? It's not _that_ obvious without reading the commit message. But you can figure out looking a bit more carefully at the code, so whatever you prefer really.

tools/src/format.rs
```diff
@@ -102,7 +102,12 @@ pub fn format_code(jumps: &HashMap<String, Jump>, format: FormatAs,
...
+                    &d[0].sym
+                } else {
+                    let displayed = &input[token.start .. token.end];
+                    &d.iter().find(|ref item| { item.sym.contains(displayed) }).unwrap_or(&d[0]).sym
```
emilio: nit: no need for `ref` (`find` already gives you a reference I'd think), nor for braces in single line closures (though I notice they're used below too). So also your call :)

staktrace:
Thanks, updated patch per comments.

PR #84, opened 2018-01-27
Bug 1322740 - Don't let the height of the code pane grow more than gu…

staktrace:
…tter

@mikeconley or @asutherland - do you have any suggestions on better ways to do this? CSS is not really my forte

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -139,6 +139,7 @@ td pre code {
     margin: 0;
     padding: 0;
     display: block;
+    max-height: calc(12px * 1.3); /* don't let bolded code increase the line-height - see bug 1322740 comment 9 */
 }
 code a {
     cursor: pointer;


mstange:
The `<code>` element's height is determined by the height of its line box, and the line box is influenced by the `line-height` and `vertical-align` properties of its children. I learned this from https://iamvdo.me/en/blog/css-font-metrics-line-height-and-vertical-align . So even though the line height is already explicitly set to 1.3 for all children (and all children have the same computed font size), since there are two different fonts in the same line, with different baselines, the line box ends up slightly larger than that.

It doesn't seem like there's a way to set the height of the line box to a fixed value. But I think adding a rule like `td pre code span { vertical-align: middle; }` might also do the trick, though that might make the two fonts slightly misaligned with each other because they'll no longer be aligned to have the same baseline.

staktrace:
@theres-waldo - can you try markus' suggestion in the previous comment and check if it works for you? If so that would be a cleaner fix.

theres-waldo:
Adding `td pre code span { vertical-align: middle; }` actually exacerbates the problem - the code lines now take up even more vertical space than the line numbers, and the line numbers are therefore even more misaligned.

staktrace:
It's not worth spending more time on this, I'm just going to merge.

asutherland:
I think the max-height fix is a reasonable and good solution.  HTML/JS-based editors like Atom are configured by specifying a Font Size defined as "Height in pixels of editor text." and then further down on the same Editor prefs screen, a Line Height defined as "Height of editor lines, as a multiplier of font size."

Obviously, when we start allowing automated inlining of cat gifvs and rick-rolls in the source output we may need to overhaul things then, but this should work for now.

PR #88, opened 2018-02-05
Allow catch statements with no pattern

staktrace:


staktrace:
For reference this occurs "in the wild" at https://hg.mozilla.org/mozilla-central/file/95fd9deac911/testing/web-platform/tests/resources/chromium/web-bluetooth-test.js#l384 for example.

PR #98, opened 2018-03-21
Make ssh.py compatible with python3

staktrace:
r? @emilio

### DIFF
--- a/infrastructure/aws/ssh.py
+++ b/infrastructure/aws/ssh.py
@@ -25,7 +25,7 @@ def print_instances():
             for tag in instance.tags:
                 tags[tag['Key']] = tag['Value']
 
-        print instance.id, group, tags
+        print(instance.id, group, tags)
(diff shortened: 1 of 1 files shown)


infrastructure/aws/ssh.py
```diff
@@ -66,9 +66,9 @@ def log_into(instance):
     sys.exit(p.returncode)
 
 if len(sys.argv) == 1:
-    print 'usage: %s <instance-id>' % sys.argv[0]
+    print('usage: %s <instance-id>' % sys.argv[0])
```
emilio: I thought this would need to use `.format`, but I guess you've tested it?

staktrace:
Yeah I think this file is simple enough that it just works. I tested it using a python 2.7 virtualenv as well as a python 3.6 virtualenv and in both cases the modified file ran without issues.

PR #99, opened 2018-03-21
Encoding

staktrace:
r? @emilio

### DIFF
--- a/router/router.py
+++ b/router/router.py
@@ -48,6 +48,12 @@ def parse_path_filter(filter):
 
     return filter
 
+def escape_regex(searchString):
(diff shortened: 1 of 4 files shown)


router/router.py
```diff
@@ -48,6 +48,11 @@ def repl(m):
 
     return filter
 
+def escape_regex(searchString):
+    # a version of re.escape that doesn't escape every non-ASCII character,
+    # and therefore doesn't mangle utf-8 encoded characters.
```
emilio: Maybe point to the bug?

staktrace: Done, thanks!

PR #100, opened 2018-03-27
rust-indexer: Handle properly generated files.

emilio:
This will allow the indexer on Servo to actually be useful.

### DIFF
--- a/scripts/rust-analyze.sh
+++ b/scripts/rust-analyze.sh
@@ -12,10 +12,11 @@ set -x # Show commands
 CONFIG_FILE=$(realpath $1)
 TREE_NAME=$2
 
-find $OBJDIR -type d -name save-analysis |
-while read ANALYSIS_DIR; do
+if [ -d "$OBJDIR" ]; then
+  ANALYSIS_DIRS="$(find $OBJDIR -type d -name save-analysis)"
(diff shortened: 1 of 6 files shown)


emilio:
cc @jdm

scripts/rust-analyze.sh
```diff
@@ -12,10 +12,11 @@ set -x # Show commands
 CONFIG_FILE=$(realpath $1)
 TREE_NAME=$2
 
-find $OBJDIR -type d -name save-analysis |
-while read ANALYSIS_DIR; do
+if [ -e "$OBJDIR" ]; then
```
staktrace: This should probably be -d instead of -e to ensure it's a directory and not just a file

tests/tests/files/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 name = "files"
 version = "0.1.0"
 authors = ["Emilio Cobos Álvarez <emilio@crisal.io>"]
+build = "build.rs"
```
staktrace: Forgot to add this file?

emilio: Whoops, fixed

scripts/rust-analyze.sh
emilio: Yup, agreed. Done

PR #102, opened 2018-03-28
Bug 1449606 - Linkify bug numbers and links in string literals

staktrace:


staktrace:
Will do. I also want to add some way of actually checking for regressions easily, but I'll do that separately.

PR #106, opened 2018-04-08
Move files around

staktrace:
This moves some files from the top level folder into appropriate subfolders. Just feels weird to have them in the top level.


staktrace:
(No functional changes)

PR #108, opened 2018-04-27
Bug 1417222 - Stop rendering markdown/RST files by default

staktrace:


staktrace:
Agreed, I'm happy to add a link to view the formatted version but I think the default should be the blame view.

PR #115, opened 2018-07-11
Bug 1475041 - Handle SpreadExpression properties inside objects

staktrace:


scripts/js-analyze.js
```diff
@@ -727,7 +727,13 @@ let Analyzer = {
 
     case "ObjectPattern":
       for (let prop of pat.properties) {
-        this.pattern(prop.value);
+        if ("value" in prop) {
```
amccreight: Could this instead be a check if (prop.type == "Property")? Looking at ASTSerializer::objectPattern() (I couldn't find any better documentation of this...), that seems to be the only other case. That feels a little easier to understand than checking for some property.

staktrace: Yup, makes sense. I updated the PR to do that. Also added a test.

amccreight:
Thanks for adding the test! I didn't realize we had them but it makes sense.

PR #118, opened 2018-07-27
Bug 1478097 - Visit implicit code inside constructors

staktrace:


staktrace:
Carrying r+ from the bug

PR #120, opened 2018-07-28
Allow using an envvar to make the output files diff easily

staktrace:
If you generate two index dirs with different versions of the mozsearch code,
doing a recursive diff on the two file/ directories turns up a lot of spew
because of the changed timestamp. With this patch you can set the
MOZSEARCH_DIFFABLE=1 envvar while running indexer-run.sh to omit the timestamp
and get a cleaner diff. This makes it easier to test changes.

### DIFF
--- a/tools/src/format.rs
+++ b/tools/src/format.rs
@@ -1,3 +1,4 @@
+use std::env;
 use std::io::Write;
 use std::collections::HashMap;
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


tools/src/format.rs
```diff
@@ -296,12 +297,15 @@ pub fn format_file_data(cfg: &config::Config,
     let filename = path_wrapper.file_name().unwrap().to_str().unwrap();
 
     let title = format!("{} - mozsearch", filename);
-    let opt = Options {
+    let mut opt = Options {
         title: &title,
         tree_name: tree_name,
         include_date: true,
```
emilio: Just `include_data: env::var(..).is_err()`.

PR #123, opened 2018-07-29
rust-indexer: Index rust standard library source.

emilio:
This reuses the generated files infrastructure just for convenience.

We pull the source and the analysis data from the `rustup` component, so no need to build anything, which means it's quite fast.

### DIFF
--- a/scripts/export-rust-source.sh
+++ b/scripts/export-rust-source.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -x # Show commands
+set -eu # Errors/undefined vars are fatal
+set -o pipefail # Check all commands in a pipeline
+
+OBJDIR="$1"
+
+mkdir -p "$OBJDIR/__RUST__"
+
+cp -r "$(rustc --print sysroot)/lib/rustlib/src/rust/src" "$OBJDIR/__RUST__"
--- a/scripts/rust-analyze.sh
+++ b/scripts/rust-analyze.sh
@@ -15,9 +15,15 @@ TREE_NAME=$2
 
 if [ -d "$OBJDIR" ]; then
   ANALYSIS_DIRS="$(find $OBJDIR -type d -name save-analysis)"
+  if [ "x$ANALYSIS_DIRS" = "x" ]; then
+    exit 0 # Nothing to analyze really.
+  fi
+  ANALYSIS_DIRS="$ANALYSIS_DIRS $(find `rustc --print sysroot` -type d -name analysis)"
+  RUST_SRC="$(rustc --print sysroot)/lib/rustlib/src/rust/src"
   $MOZSEARCH_PATH/tools/target/release/rust-indexer \
     "$FILES_ROOT" \
     "$INDEX_ROOT/analysis" \
     "$OBJDIR" \
+    "$RUST_SRC" \
     $ANALYSIS_DIRS
 fi
--- a/infrastructure/indexer-provision.sh
+++ b/infrastructure/indexer-provision.sh
@@ -44,6 +44,7 @@ sudo update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-4.0
 
 # Install Rust. We need rust nightly to use the save-analysis
 curl -sSf https://static.rust-lang.org/rustup.sh | sh -s -- --channel=nightly
+rustup component add rust-src rust-analysis
 
 # Install codesearch.
 rm -rf livegrep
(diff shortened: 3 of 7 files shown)


staktrace:
This looks like it's missing the export-rust-source.sh script?

emilio:
Ugh, indeed :)

Updated.

scripts/export-rust-source.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
staktrace: Please update the decorations to match the changes from #116

scripts/rust-analyze.sh
```diff
@@ -14,10 +14,12 @@ CONFIG_FILE=$(realpath $1)
 TREE_NAME=$2
 
 if [ -d "$OBJDIR" ]; then
-  ANALYSIS_DIRS="$(find $OBJDIR -type d -name save-analysis)"
+  ANALYSIS_DIRS="$(find $OBJDIR -type d -name save-analysis) $(find `rustc --print sysroot` -type d -name analysis)"
```
staktrace: We should probably do a thing here where we first check for the save-analysis files in $OBJDIR, and if there aren't any, we bail out. Otherwise we'll run the rust-indexer even on repos with no rust code (e.g. for nss) and just "analyze" the stdlib analysis files.

scripts/export-rust-source.sh
```diff
@@ -0,0 +1,10 @@
...
+
+set -e # Errors are fatal
+set -x # Show commands
+
+OBJDIR="$1"
+
+mkdir -p "$OBJDIR/__RUST__"
+
+cp -r "$(rustc --print sysroot)/lib/rustlib/src/rust/src" "$OBJDIR/__RUST__"
```
staktrace: Shouldn't the destination have a `__GENERATED__` in it?

asutherland: Maybe we want to create a fourth category beyond normal/test/generated of "stdlib"?

asutherland: > Sure. Though that can be compatible to still have the same setup right? Do you know where we generate the data?

https://github.com/mozsearch/mozsearch/blob/master/router/router.py#L121 near the bottom checks for "__GENERATED__" to return generated.

staktrace:
Is this blocked on anything? If not I can rebase and merge it.

emilio:
Need to add the extra category, though that can be a followup, and probably ensure it still works with your recent changes. If you could do that it'd be awesome, otherwise let me know and I can do it.

staktrace:
Yeah I'll take a look and let you know if I get stuck anywhere.

staktrace:
I rebased the patches and ran into a problem. Now that we are getting m-c's rust analysis files from taskcluster, they are based on whatever version of rustc is being used in m-c (currently 1.29.0). Specifically the crate disambiguator values are also from that version. So if we install the nightly `rust-analysis` component on the indexer the disambiguators don't match up and the `rust-indexer.rs` program can't link them together. A couple of options:

1) Install the stable `rust-analysis` component on the indexer, rather than the nightly one. This would work, but only if m-c is also using whatever the current stable version is. It seems kinda brittle. We could make it less brittle by somehow scraping the actual version of rustc being used to build m-c (from the taskcluster configs or something).
2) Have the taskcluster searchfox job also produce the analysis files for the version of rustc being used. I'm not sure how hard this would be; the `repack_rust.py` script could probably be modified to pull the analysis files and then we could grab them somehow.

I'm kinda leaning towards (2) as being less brittle but it also seems somewhat complex. Any thoughts?

emilio:
> I'm kinda leaning towards (2) as being less brittle but it also seems somewhat complex. Any thoughts?

I think the pulling-the analysis part should actually be trivial. I had to do something very similar for rustfmt already, which I think only needs an extra call:

  https://searchfox.org/mozilla-central/rev/0b8ed772d24605d7cb44c1af6d59e4ca023bd5f5/taskcluster/scripts/misc/repack_rust.py#285

That was https://bugzilla.mozilla.org/show_bug.cgi?id=1432563.

I think once that's done we could pull the artifacts using something like:

```
./mach artifact toolchain --from-build linux64-rust
```

And the analysis files would be somewhere in `rustc/lib/$target/`.

emilio:
Well I guess given we download multiple standard libraries, the call should be somewhere around here, maybe:

https://searchfox.org/mozilla-central/rev/0b8ed772d24605d7cb44c1af6d59e4ca023bd5f5/taskcluster/scripts/misc/repack_rust.py#229

staktrace:
I finally got around to rebasing this again and hooking it up to the taskcluster stuff. Updated version of  this patch is at https://github.com/mozsearch/mozsearch/pull/209 so I'll close this.

PR #125, opened 2018-07-30
Tag indexers with some useful info

staktrace:
This makes it easier to see which stopped instances were triggered by who/where


staktrace:
Yeah, it will show in the Amazon EC2 console if you click on an instance. It also shows up if you run `infrastructure/aws/ssh.py` as part of the listing output.

PR #129, opened 2018-08-14
Bug 1466402 - Bump livegrep version to fork from latest master

staktrace:
This will require rebuilding the indexer and web-server AMIs.

Current dev instance is running with this.

### DIFF
--- a/scripts/build-codesearch.py
+++ b/scripts/build-codesearch.py
@@ -77,7 +77,7 @@ livegrep_config['fs_paths'].append({
 
 json.dump(livegrep_config, open('/tmp/livegrep.json', 'w'))
 
-run(['codesearch', '/tmp/livegrep.json', '-dump_index', tree['codesearch_path'],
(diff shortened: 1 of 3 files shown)


scripts/build-codesearch.py
```diff
@@ -77,7 +77,7 @@ def copy_objdir_files(dest_dir, config):
 
 json.dump(livegrep_config, open('/tmp/livegrep.json', 'w'))
 
-run(['codesearch', '/tmp/livegrep.json', '-dump_index', tree['codesearch_path'],
+run(['codesearch', '/tmp/livegrep.json', '-dump_index', tree['codesearch_path'], '-index_only',
```
amccreight: Is this just something that is needed for the new version?

staktrace: Yeah, otherwise it starts the server along with dumping the index. Previously I think the `-dump_index` automatically suppressed starting the server.

PR #131, opened 2018-08-16
Install nodejs, now needed by the mozilla-central build

staktrace:


staktrace:
For the record this goes with https://bugzilla.mozilla.org/show_bug.cgi?id=1484018

PR #133, opened 2018-08-17
Bug 1483933 - Don't start a new instance of livegrep so eagerly

staktrace:
Based on server logs, we can get exceptions thrown while searching for a number of
different reasons, including e.g. bad regex queries. In all of these cases we
were attempting to start a new instance of the server, which is wrong - we only
want to do it if we got a connection error indicating the server died. In the other
cases we just return an empty result for now, but in the future we should do a
better job of reporting some error message details back to the user.


staktrace:
Gonna merge this now because without it the web server eventually creates too many instances of livegrep and OOMs.

PR #134, opened 2018-08-18
Install cbindgen for m-c

emilio:
I added this build dependency in https://bugzilla.mozilla.org/show_bug.cgi?id=1478813.

Perhaps we should run `./mach bootstrap` somehow...


staktrace:
This installs `cbindgen` but `~/.cargo/bin` isn't on the path. We'll have to put that in the mozconfig generator.

emilio:
@staktrace the mozilla-central build does look at `~/.cargo/bin`, fwiw :)

PR #137, opened 2018-08-31
rust indexing tweaks

staktrace:
Some changes that I want for bug 1487583, but that also fix bug 1429877 along the way. Plus some development ergonomic improvement in the third patch.


tools/src/bin/rust-indexer.rs
```diff
@@ -226,6 +227,21 @@ fn find_generated_or_src_file(
     file_name.strip_prefix(tree_info.src_dir).ok().map(From::from)
 }
 
+fn read_existing_contents(
+    map: &mut BTreeSet<String>,
+    file: &PathBuf,
```
emilio: this can just be `&Path` I think.

tools/src/bin/rust-indexer.rs
```diff
@@ -226,6 +227,21 @@ fn find_generated_or_src_file(
...
+) {
+    match File::open(file) {
```
emilio: nit: use `if let Ok(f) = ...`?

tools/src/bin/rust-indexer.rs
```diff
@@ -331,7 +333,7 @@ fn analyze_file(
         let def = match defs.get(file_analysis, ref_.ref_id) {
             Some(d) => d,
             None => {
-                eprintln!("Dropping ref {:?}, kind {:?}, no def", ref_.ref_id, ref_.kind);
+                debug!("Dropping ref {:?}, kind {:?}, no def", ref_.ref_id, ref_.kind);
```
emilio: I'd prefer if a couple of these would be `warn!` ings instead, but if they clobber the log a lot `debug!` is fine.

PR #142, opened 2018-09-11
Warnings

staktrace:
/cc @emilio - turns out the "dropping ref" warnings were actually getting hit a lot so I'm downgrading it to debug here


amccreight:
I haven't actually looked at the second diff yet. I didn't realize the review was global for a pull request.

amccreight:
Ok, now I have looked at it all.

PR #145, opened 2018-09-17
Bug 1490521 - Add script to better merge analysis files.

staktrace:
This script can be used to merge analysis files from different
platforms; it will union the symbols for source lines.

### DIFF
--- a/scripts/merge-analyses.py
+++ b/scripts/merge-analyses.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python
+
+# This script merges the analysis data in the files provided as
+# arguments, and prints the merged analysis data to stdout.
+# The "target" data lines from the input files are copied to
+# the output as-is. The "source" data lines are merged such
+# that the syntax and symbol properties are unioned across
+# all input lines that have a matching (loc, pretty) tuple.
+# This ensures that for a given identifier, only a single context
+# menu item will be displayed for a given "pretty" representation,
+# and that context menu will link to all the symbols from all
+# the input files that match that.
+
+# Note that this code intentionally preserves the order of stuff
+# within each line, as much as possible. However, the lines
+# themselves get reordered, but the output can be piped through
+# `sort` as needed. Maintaining the order of stuff within the
+# line is important as reordering things can result in identical
+# lines getting sorted differently. e.g.
+#   { loc: foo, pretty: bar }
+# and
+#   { pretty: bar, loc: foo }
+# are semantically identical but would not get deduplicated by
+# piping through `sort` and `uniq`.
+
+from collections import OrderedDict
+import json
+import sys
+
+if len(sys.argv) == 1:
+    sys.stderr.write("Usage: merge-analyses.py <filename> [<filename> ...]\n")
+    sys.stderr.write("  This script will merge the analysis data from the given files\n")
+    sys.stderr.write("  and print it to stdout\n")
+    sys.exit(1)
+
+sourcemap = {}
+
+def SplitToOrderedSet(comma_sep_str):
+    return OrderedDict([(item, None) for item in comma_sep_str.split(',')])
+
+for filename in sys.argv[1:]:
+    with open(filename) as filehandle:
+        for line in filehandle.readlines():
+            # parse the JSON into an OrderedDict to maintain order of properties
+            entry = json.loads(line, object_pairs_hook=OrderedDict)
+
+            if "source" not in entry:
+                print(line.strip())
+                continue
+
+            for prop in entry:
+                if prop not in ["source", "loc", "pretty", "syntax", "sym", "no_crossref"]:
+                    sys.stderr.write("WARNING: Unexpected property %s found in source line in analysis file %s\n" % (prop, filename))
+
+            key = (entry["loc"], entry["pretty"])
+            if key in sourcemap:
+                # We already encountered a line with this (loc,pretty) tuple,
+                # so merge the syntax and sym properties. Again use an
+                # OrderedDict (with None values, so it is effectively an
+                # ordered set) to avoid reordering the existing tokens.
+                oldEntry = sourcemap[key]
+                syntaxes = SplitToOrderedSet(oldEntry["syntax"])
+                syntaxes.update(SplitToOrderedSet(entry["syntax"]))
+                oldEntry["syntax"] = ','.join(i for i in syntaxes)
+                symbols = SplitToOrderedSet(oldEntry["sym"])
+                symbols.update(SplitToOrderedSet(entry["sym"]))
+                oldEntry["sym"] = ','.join(i for i in symbols)
+                sourcemap[key] = oldEntry
+            else:
+                # Haven't encountered this before, so add it to the map
+                sourcemap[key] = entry
+
+for key in sourcemap:
+    print(json.dumps(sourcemap[key], separators=(',', ':')))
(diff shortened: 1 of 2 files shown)


scripts/merge-analyses.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python
```
amccreight: Please add a mention and description of this script to analysis.md (to the "Analysis records are currently generated from" section).

scripts/merge-analyses.py
```diff
@@ -0,0 +1,57 @@
...
+# the output as-is. The "source" data lines are merged such
+# that the syntax and symbol properties are unioned across
+# all input lines that have a matching (loc, pretty) tuple.
+# This ensures that for a given identifier, only a single context
+# menu item will be displayed for a given "pretty" representation,
+# and that context menu will link to all the symbols from all
+# the input files that match that.
+
+# Note that this code intentionally preserves the order of stuff
+# within each line, as much as possible. However, the lines
```
amccreight: Why does it preserve the order?

scripts/merge-analyses.py
```diff
@@ -0,0 +1,57 @@
...
+import json
+import sys
+
+sourcemap = {}
+
+def SplitToOrderedSet(comma_sep_str):
+    return OrderedDict([(item, None) for item in comma_sep_str.split(',')])
+
+for filename in sys.argv[1:]:
+    filehandle = open(filename)
```
amccreight: I guess there's no real reason to explicitly close these file handles when you are done?

scripts/merge-analyses.py
```diff
@@ -0,0 +1,57 @@
...
+from collections import OrderedDict
+import json
+import sys
+
+sourcemap = {}
+
+def SplitToOrderedSet(comma_sep_str):
+    return OrderedDict([(item, None) for item in comma_sep_str.split(',')])
+
+for filename in sys.argv[1:]:
```
amccreight: Maybe print out a help message if there are no arguments provided? (Or maybe even if there's only one provided.)

scripts/merge-analyses.py
```diff
@@ -0,0 +1,57 @@
...
+    filehandle = open(filename)
+    for line in filehandle.readlines():
+        # parse the JSON into an OrderedDict to maintain order of properties
+        obj = json.loads(line, object_pairs_hook=OrderedDict)
+
+        if "source" not in obj:
+            print(line.strip())
+            continue
+
+        key = obj["loc"] + obj["pretty"]
```
amccreight: I think it would be better if you used (obj["loc"], obj["pretty"]) as your key, to avoid concerns about collisions between the ambiguity of the end of loc and the start of pretty (eg loc = FooBar and pretty = Baz would collide with loc = Foo and pretty = BarBaz.) Technically it might be a little more efficient, too, because you aren't copying the strings. IIRC tuple types in Python are immutable and can be used as keys in dicts.

scripts/merge-analyses.py
```diff
@@ -0,0 +1,57 @@
...
+            # OrderedDict (with None values, so it is effectively an
+            # ordered set) to avoid reordering the existing tokens.
+            entry = sourcemap[key]
+            syntaxes = SplitToOrderedSet(entry["syntax"])
+            syntaxes.update(SplitToOrderedSet(obj["syntax"]))
+            entry["syntax"] = ','.join(i for i in syntaxes)
+            symbols = SplitToOrderedSet(entry["sym"])
+            symbols.update(SplitToOrderedSet(obj["sym"]))
+            entry["sym"] = ','.join(i for i in symbols)
+            sourcemap[key] = entry
```
amccreight: Maybe rename |obj| to |entry| and |entry| to |oldEntry|? I think it would make it easier to tell that they are the same "type" of thing.

scripts/merge-analyses.py
```diff
@@ -0,0 +1,57 @@
...
+        key = obj["loc"] + obj["pretty"]
+        if key in sourcemap:
+            # We already encountered a line with this (loc,pretty) tuple,
+            # so merge the syntax and sym properties. Again use an
+            # OrderedDict (with None values, so it is effectively an
+            # ordered set) to avoid reordering the existing tokens.
+            entry = sourcemap[key]
+            syntaxes = SplitToOrderedSet(entry["syntax"])
+            syntaxes.update(SplitToOrderedSet(obj["syntax"]))
+            entry["syntax"] = ','.join(i for i in syntaxes)
```
amccreight: Is the "syntax" field ever different between Linux and OSX? I assume this is more for completeness than something that actually affects the output in this particular case. It might be interesting to have a hot path that skips this conversion if the fields in |entry| and |obj| are identical, but I suppose that's premature optimization.

scripts/merge-analyses.py
```diff
@@ -0,0 +1,57 @@
...
+            print(line.strip())
+            continue
+
+        key = obj["loc"] + obj["pretty"]
+        if key in sourcemap:
+            # We already encountered a line with this (loc,pretty) tuple,
+            # so merge the syntax and sym properties. Again use an
+            # OrderedDict (with None values, so it is effectively an
+            # ordered set) to avoid reordering the existing tokens.
+            entry = sourcemap[key]
```
amccreight: Maybe assert that there are no other unexpected fields in |entry|?

scripts/merge-analyses.py
```diff
@@ -0,0 +1,57 @@
...
+# Note that this code intentionally preserves the order of stuff
+# within each line, as much as possible. However, the lines
+# themselves get reordered, but the output can be piped through
+# `sort` as needed.
+
+from collections import OrderedDict
+import json
+import sys
+
+sourcemap = {}
```
amccreight: This should also be an OrderedDict if you want the output to be deterministic.

amccreight: I guess the sort and uniq you run on the output means this doesn't matter.

scripts/merge-analyses.py
staktrace: Done. I also filed https://bugzilla.mozilla.org/show_bug.cgi?id=1492322 to update the docs more generally, that's something I've been meaning to do.

scripts/merge-analyses.py
staktrace: I updated the comment with this:

```
# Maintaining the order of stuff within the
# line is important as reordering things can result in identical
# lines getting sorted differently. e.g.
#   { loc: foo, pretty: bar }
# and
#   { pretty: bar, loc: foo }
# are semantically identical but would not get deduplicated by
# piping through `sort` and `uniq`.
```

scripts/merge-analyses.py
staktrace: Makes sense, will do.

scripts/merge-analyses.py
staktrace: Good idea, done.

scripts/merge-analyses.py
staktrace: I didn't really check to see if it differs or not. As you said, this was for completeness mostly. I'll leave it as-is for now.

scripts/merge-analyses.py
staktrace: Will do

scripts/merge-analyses.py
staktrace: Yeah I figured that I need to run sort and uniq on it anyway because the source lines from one platform will end up getting added to the sourcemap after all the source lines for the other platform (assuming the lines were distinct and therefore didn't get merged).

PR #146, opened 2018-09-18
Ensure we provision rustup.

staktrace:
Previously we were installing rust using static.rust-lang.org which
installs rustc into /usr/local/bin and doesn't install rustup. However,
in order to get the rust stdlib source, it's better to use the more
traditional rustup installation mechanism, which installs it into the
user home folder and also installs rustup. This patch does that and
then uses `rustup update` to update rust as needed, instead of rerunning
the installer in update.sh

This also means we need to run the update.sh script as the ubuntu user,
instead of root, so that rustup is on the $PATH. This is desirable anyway
because running things as root sucks, and the update.sh scripts already
use sudo for individual commands that need root, so in this case running
the script as root is completely avoidable.

### DIFF
--- a/infrastructure/aws/trigger-web-server.py
+++ b/infrastructure/aws/trigger-web-server.py
@@ -40,7 +40,7 @@ userData = '''#!/usr/bin/env bash
 
 cd ~ubuntu
 touch web_server_started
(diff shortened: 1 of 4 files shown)


staktrace:
#123 requires this

staktrace:
Shoot, I forgot this needs to update the lambda task. I need to do that.

PR #148, opened 2018-09-20
Bug 1474434 - Fix 'pretty' property for rust target records.

staktrace:
The pretty name in target records is used to build the identifiers table,
and should not contain spaces. The source record, however, can contain
spaces and be more human-readable. This patch corrects the code from #78
which made both source and target records human-readable instead of just
the source records.

### DIFF
--- a/tools/src/bin/rust-indexer.rs
+++ b/tools/src/bin/rust-indexer.rs
@@ -191,14 +191,7 @@ fn visit(
     out.insert("loc".into(), Value::String(span_to_string(location)));
     out.insert("target".into(), json!(1));
     out.insert("kind".into(), Value::String(kind.into()));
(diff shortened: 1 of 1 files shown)


tools/src/bin/rust-indexer.rs
```diff
@@ -187,18 +187,19 @@ fn visit(
     use serde_json::map::Map;
     use serde_json::value::Value;
 
-    let mut out = Map::new();
-    out.insert("loc".into(), Value::String(span_to_string(location)));
-    out.insert("target".into(), json!(1));
-    out.insert("kind".into(), Value::String(kind.into()));
     let pretty = {
```
emilio: nit: Maybe move this below?

PR #149, opened 2018-09-22
Bug 1474380 - Add accelerator key for permalink, log, and raw navlinks

staktrace:


staktrace:
(This is currently deployed to dev.searchfox.org if you want to test it out)

static/js/panel.js
```diff
@@ -19,4 +19,44 @@ $(function() {
...
+        }
+        switch (event.key) {
+            case 'y':
+            case 'Y':
+                if (history.state && history.state.permalink) {
+                    break;
+                }
+                var linkNode = $('.panel a[title="Permalink"]');
+                if (linkNode.length) {
+                    history.pushState(
```
asutherland: Using pushState here is an awesome UX improvement since we don't actually navigate (which means no load time and we still have all the smarts of the page), but is there a reason to only do this for the accelerator and not a left-click?  (Other than it being annoying since I think we need to check that button===0 and !{ctrlKey, shiftKey, altKey, metaKey} before we preventDefault() and use pushState instead>).

If we stick with diverging behavior, please add a comment here that makes it more obvious what's going on here.

staktrace: I was feeling a bit weird about this change in that you end up with different content depending on whether you use the accelerator or click on a permalink (with the accelerator, as you said, you get the "smarts" of the page, whereas loading a permalink will give you the "dumb" rendering of the page). That feels like a HTTP violation of some sort because the same URL should give you the same content, and so I didn't want to spread that violation any further than needed. But you're right that clicking on the permalink text and using the accelerator should produce consistent results, and making that change is kind of orthogonal to the fact that we would have this "HTTP violation" anyway, so I might as well do it. Will update patch shortly.

PR #152, opened 2018-09-26
Bug 1494348 - Add a toggle to disable keyboard shortcuts.

staktrace:
It doesn't play well with some features like typeahead find, so
people may want to disable it. This also adds a small cookie
library to save the user's choice.

### DIFF
--- a/static/css/icons.css
+++ b/static/css/icons.css
@@ -68,6 +68,10 @@ a.type {
 }
 /* End of Context-menu only icons */
 
+.panel label.panel-accel {
+    padding: 0.5rem
+}
+
 .panel .accel {
     /* simulate a keyboard key effect */
     font-family: monospace;
--- a/static/js/panel.js
+++ b/static/js/panel.js
@@ -49,7 +49,7 @@ $(function() {
         });
     }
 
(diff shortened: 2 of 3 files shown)


staktrace:
This is currently deployed to dev.searchfox.org

staktrace:
Oh wait, need to fix the cookie setting to apply to all paths

staktrace:
Patch updated. I kicked off a new dev run since I snuck in a few more tweaks.

asutherland:
Er, note that my bit about performance is moot since you already had structured the code to only do things on deltas.  (You definitely don't want to be doing localStorage reads like they were local reads; so caching the state or impacts of the state is the right thing to do, and you were already doing that.)

staktrace:
>    $('#panel-accel-enable')[0].checked = newState;
>    updateAccelerators();

I don't think this part quite works because then updateAccelerators() will early-exit since newstate == oldState. But I'll make changes in line with what you suggested.

staktrace:
^ Patch updated and deployed to dev.searchfox.org.

asutherland:
Changes look great!

PR #154, opened 2018-10-01
Bug 1493527 - Optimize analysis merging

staktrace:


staktrace:
Thanks for the quick review! Addressed the nits, except I decided to leave the contains() check as-is since I feel it's slightly more readable.

PR #155, opened 2018-10-02
Some small tweaks to make mozsearch a bit less Gecko-specific.

emilio:
I convinced some Igalia folks to try maintaining their own searchfox instance with an up-to-date WebKit index (which would be very useful for us as well, since grepping WebKit is a bit of a pain).

This were some fixes that I needed to set something very simple up.


staktrace:
LGTM, thanks!

PR #156, opened 2018-10-05
Downgrade a panic to a warning so we handle it more gracefully

staktrace:
We'll still get notified via a warning email but at least we'll finish indexing, minus the analysis data that couldn't be parsed.

### DIFF
--- a/tools/src/file_format/analysis.rs
+++ b/tools/src/file_format/analysis.rs
@@ -180,15 +180,18 @@ pub fn read_analyses<T>(filenames: &[&str], filter: &mut FnMut(&Object) -> Optio
             Err(_) => continue,
         };
         let reader = BufReader::new(&file);
-        let mut lineno = 1;
+        let mut lineno = 0;
         for line in reader.lines() {
             let line = line.unwrap();
+            lineno += 1;
             let data = Json::from_str(&line);
             let data = match data {
                 Ok(data) => data,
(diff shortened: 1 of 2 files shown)


tools/src/file_format/analysis.rs
```diff
@@ -180,15 +180,18 @@ pub fn read_analyses<T>(filenames: &[&str], filter: &mut FnMut(&Object) -> Optio
...
+        let mut lineno = 0;
         for line in reader.lines() {
             let line = line.unwrap();
+            lineno += 1;
             let data = Json::from_str(&line);
             let data = match data {
                 Ok(data) => data,
-                Err(e) => panic!("error {} on file {} line {}: {}", e, filename, lineno, &line),
+                Err(e) => {
+                    warn!("Error [{}] trying to read analysis from file [{}] line [{}]: [{}]", e, filename, lineno, &line);
```
emilio: maybe `error!` is more appropriate?

staktrace: With warn! we will get email notification, per https://github.com/mozsearch/mozsearch/blob/c7862c843d09467d1f24cd5abeee938650057762/infrastructure/aws/send-warning-email.py#L14

emilio: Oh right, we should maybe handle `error` as well... Anyway, sg :)

PR #157, opened 2018-10-09
Lone surrogates and warning emails

staktrace:


scripts/js-analyze.js
```diff
@@ -53,13 +53,36 @@ function locstr2(loc, str)
...
+      // c is a high surrogate, check to make sure next char is a low surrogate
+      var d = name.charCodeAt(i + 1);
+      if (d >= 0xDC00 && d <= 0xDFFF) {
+        // valid; skip over the pair and continue
+        i++;
+        continue;
+      }
+    }
+    // fail on any surrogate characters that weren't part of a pair
+    if (c >= 0xD800 && c <= 0xDFFF) {
```
emilio: Looks like you could avoid having the `if (c >= 0xD800)` duplicated by early `continue`ing when `c` is less than 0xD800 and such... But maybe the code would be less obvious, so it's not a big deal.

staktrace: Ah good point, I made the change as it's probably worth doing the early-continue for the common case.

PR #162, opened 2018-10-31
Windows

staktrace:
Stuff I needed to fix for windows compat. Note that the patch to rust-indexer.rs doesn't properly deal with some cases that happen in practice (the mixed-separators case at https://github.com/nrc/rls-data/issues/20) but there's relatively few cases where that actually happens. I'm hoping it can be fixed in rls-data rather than having to make this hack even uglier.

### DIFF
--- a/tools/src/bin/rust-indexer.rs
+++ b/tools/src/bin/rust-indexer.rs
@@ -349,6 +349,41 @@ fn analyze_file(
     }
 }
 
+/// Uses heuristics to determine if the path is a Windows path,
+/// and if so, returns a Linux "equivalent". Not really equivalent,
+/// because there is no such thing, but equivalent enough for our
+/// purposes - that is, is_relative() returns a correct thing, and
+/// strip_prefix will work on relative paths.
+/// Note that this is a "linuxzed" path and not a "unixed" path
+/// this code should ever only really run inside the linux vagrant
+/// machine.
+/// There are real-life cases in which this code is too simple, see
+/// https://github.com/nrc/rls-data/issues/20 for an example.
+fn linuxized_path(path: &PathBuf) -> PathBuf {
+    // Windows paths have backslashes and so on a Linux host will
+    // just have a single long component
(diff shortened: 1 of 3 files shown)


tools/src/bin/rust-indexer.rs
```diff
@@ -349,6 +349,40 @@ fn analyze_file(
...
+            if pathstr.find('\\').is_some() {
+                // Has a backslash, so probably a Windows path. Let's lean
+                // those slashes forward
+                let converted = pathstr.replace('\\', "/");
+                if converted.find(":/") == Some(1) {
+                    // Starts with a drive letter, so let's turn this into
+                    // an absolute path
+                    let abs = "/".to_string() + &converted;
+                    return PathBuf::from(abs);
+                } else {
```
emilio: nit: No `else` after return? In this case the symmetry doesn't look that important, but if you think it is you can just do `return if ... { } else { };`

tools/src/bin/rust-indexer.rs
```diff
@@ -349,6 +349,40 @@ fn analyze_file(
...
 }
 
+/// Uses heuristics to determine if the path is a Windows path,
+/// and if so, returns a Linux "equivalent". Not really equivalent,
+/// because there is no such thing, but equivalent enough for our
+/// purposes - that is, is_relative() returns a correct thing, and
+/// strip_prefix will work on relative paths.
+/// Note that this is a "linuxzed" path and not a "unixed" path
+/// this code should ever only really run inside the linux vagrant
+/// machine.
```
emilio: Please link to the rls issue so that we can notice if / when it gets fixed :)

tools/src/bin/rust-indexer.rs
staktrace: Fixed, thanks

tools/src/bin/rust-indexer.rs
staktrace: Done

PR #165, opened 2018-11-06
Add mailmap support

mystor:
I haven't yet had the time to actually build this locally, so it's very very likely that there are build errors in it. I don't currently have docker set up & my system openssl is 0.0.1 versions too new for the openssl-sys crate.

The main place here where I'm expecting a compilation error needs to be fixed are in `lookup`, where I call `realname.as_ref().unwrap_or(name)`. The type of `realname` is `&Option<String>` while the type of `name` is `&str`, so a `.map(|&x| &x[..])` may need to be added before the `unwrap_or` to make the deref explicit. 

I whipped this together because it seems like `libgit2` is pretty slow to ship new releases, so while @emilio and landed mailmap support upstream a while ago, it's not available in any release yet. For mozsearch we don't need quite as robust support, so this simplified approach should probably work fine.

### DIFF
--- a/tools/Cargo.lock
+++ b/tools/Cargo.lock
@@ -33,7 +33,7 @@ version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "kernel32-sys 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.30 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.43 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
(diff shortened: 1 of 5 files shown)


staktrace:
Thanks! I can get it building and do a dev push for you to verify the results.

staktrace:
Fixups needed: https://github.com/staktrace/mozsearch/commit/b73ff1d85361bb1f1c41eae393aaf7775c300c46

staktrace:
@mystor Also if the dev run looks good, feel free to steal my fixups patch and massage it to your liking (make it more idiomatic etc.) and squash it into this PR. If it looks fine I can do the squash.

staktrace:
I found an old commit of yours and checked the differences. So e.g. https://dev.searchfox.org/mozilla-central/source/extensions/spellcheck/tests/mochitest/test_bug1272623.html has the updated info in the blame although the same page on searchfox.org doesn't. So that's good. But a page like https://dev.searchfox.org/mozilla-central/commit/b86c1ff10f7b2de0b36f7ad71c38dd52508e4d2e isn't updated. That's fixable in format.rs, I'll try that.

staktrace:
Updated and redeployed, seems good

mystor:
@staktrace Thanks for doing those cleanups & fixing up the full-commit page! They look awesome!

Feel free to perform the squash and land the changes :-)

staktrace:
Squashed and landed. Should be in tomorrow's indexing run.

amccreight:
I'm not familiar with the blame code. It sounds like this works without needing to do the huge rebuild of the blame data when the mailmap file changes. Is that right?

staktrace:
That's correct, yeah. It applies the translation when using the blame data from the blame repo. That's all done on-the-fly in response to /commit-info and /commit URL requests.

PR #167, opened 2018-11-27
highlighter: Add support for multiple symbols per identifier.

emilio:
This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1510318 and
https://bugzilla.mozilla.org/show_bug.cgi?id=1430112.

There's a chance that this highlights too much for C++, but from my testing it
looks ok, and we can always filter symbols / add more heuristics if needed.

I removed some unused [data-kind] rules while I was at it.

### DIFF
--- a/tools/src/format.rs
+++ b/tools/src/format.rs
@@ -117,21 +117,21 @@ pub fn format_code(jumps: &HashMap<String, Jump>, format: FormatAs,
                 let d = d.iter().filter(|item| !item.no_crossref).collect::<Vec<_>>();
 
                 let mut menu_jumps : HashMap<String, Json> = HashMap::new();
-                for item in d.iter() {
-                    for sym in &item.sym {
-                        match jumps.get(sym) {
-                            Some(jump) => {
-                                if !(&jump.path == path && jump.lineno == cur_line) {
-                                    let key = format!("{}:{}", jump.path, jump.lineno);
-                                    let mut obj = json::Object::new();
-                                    obj.insert("sym".to_string(), Json::String(sym.to_string()));
-                                    obj.insert("pretty".to_string(), Json::String(jump.pretty.clone()));
-                                    menu_jumps.insert(key, Json::Object(obj));
-                                }
-                            },
-                            None => {}
-                        }
+                for sym in d.iter().flat_map(|item| item.sym.iter()) {
+                    let jump = match jumps.get(sym) {
+                        Some(jump) => jump,
+                        None => continue,
+                    };
+
+                    if jump.path == *path && jump.lineno == cur_line {
+                        continue;
                     }
+
+                    let key = format!("{}:{}", jump.path, jump.lineno);
+                    let mut obj = json::Object::new();
+                    obj.insert("sym".to_string(), Json::String(sym.to_string()));
+                    obj.insert("pretty".to_string(), Json::String(jump.pretty.clone()));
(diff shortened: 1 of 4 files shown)


tools/src/format.rs
```diff
@@ -140,15 +131,27 @@ pub fn format_code(jumps: &HashMap<String, Jump>, format: FormatAs,
 
                 let menu_jumps = menu_jumps.into_iter().map(|(_, v)| v).collect::<Vec<_>>();
 
+                // Build the list of symbols for the highlighter.
                 let index = generated_json.len();
+                let syms = {
+                    let mut syms = String::new();
+                    for (i, sym) in d.iter().flat_map(|item| item.sym.iter()).enumerate() {
```
staktrace: Can't you do `d.iter().flat_map(|item| item.sym.iter()).join(",")` ?

emilio: Not without cloning all the symbols, but maybe that's ok.

staktrace:
So the dev instance is up. I'm actually seeing missing highlights for local variables, see around https://dev.searchfox.org/mozilla-central/source/gfx/layers/apz/src/APZCTreeManager.cpp#1044 for example - `clipRegion`, `aLayer`, `aParent` etc don't seem to highlight at all...

tools/src/format.rs
```diff
@@ -102,15 +102,6 @@ pub fn format_code(jumps: &HashMap<String, Jump>, format: FormatAs,
...
-                // In case the token has multiple symbols associated with it,
-                // try to pick the "best" one for determining which other tokens
-                // this gets highlighted with. For now we just take the first
-                // symbol that contains the token, or the first one in the list
-                // if none contain the token.
-                let displayed = &input[token.start .. token.end];
-                let mut syms = d.iter().flat_map(|item| item.sym.iter());
-                let id = syms.find(|sym| sym.contains(displayed)).unwrap_or(&d[0].sym[0]);
-
                 let d = d.iter().filter(|item| { !item.no_crossref }).collect::<Vec<_>>();
```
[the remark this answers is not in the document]
emilio: Thanks, nice catch :)

staktrace:
Maybe we can further filter the results of `findReferences` and only select the `span`s where the textContent matches the textContent of the element the user is mousing over? That should then restrict it to tokens that appear the same to the user and hovering over an `EventStateManager` should only highlight other `EventStateManager`s.

emilio:
> Maybe we can further filter the results of findReferences and only select the spans where the textContent matches the textContent of the element the user is mousing over?

That works for types, but not for locals. For example, for the `aEvent` variable in that function you mentioned above, the symbols are: `V_0bd2a3ae_885764fe,V_c95394516f7eb61f_885764fe256`. So we can't just do that. We could try to strip out symbols for types (which in C++ start by `T_`) that don't match the `textContent` or something, but I'd rather merge this as-is.

emilio:
I fixed up the patch to build the symbol list before the `no_crossref` filtering, and also cleaned up a bit the surrounding code in a separate commit.

staktrace:
> For example, for the aEvent variable in that function you mentioned above, the symbols are: V_0bd2a3ae_885764fe,V_c95394516f7eb61f_885764fe256

I was thinking this filtering would use the user visible token, not the symbol. But anyway I'll experiment with that separately.

emilio:
Ah! Yeah, that makes sense to me.

PR #169, opened 2018-11-28
Bug 1498024 - Further filter the matching tokens by textContent.

staktrace:
The findReferences code returns any tokens anywhere on the patch that
share at least one symbol with the target token. This can sometimes
result in a superset of what we actually want, because "hidden" symbols
(ones that the user doesn't necessarily care about but that get
attributed to tokens) can result in spurious matches.

This patch augments the filtering to ensure only tokens with the same
user-visible string get matched. This is not a perfect fix since it's
still possible to have tokens with the same user-visible strings and
symbols in separate scopes (e.g. different functions) that will get
co-highlighted. But it helps with some cases.

### DIFF
--- a/static/js/context-menu.js
+++ b/static/js/context-menu.js
@@ -54,14 +54,15 @@ function symbolsFromString(symbols) {
   return symbols.split(",");
 }
 
-function findReferences(symbols) {
+function findReferences(symbols, visibleToken) {
   symbols = symbolsFromString(symbols);
   if (!symbols.length)
     return $();
   symbols = new Set(symbols);
   return $([...document.querySelectorAll("span[data-symbols]")].filter(span => {
-    return symbolsFromString(span.getAttribute("data-symbols"))
-      .some(symbol => symbols.has(symbol))
+    return span.textContent == visibleToken &&
+           symbolsFromString(span.getAttribute("data-symbols"))
+                .some(symbol => symbols.has(symbol));
   }));
 }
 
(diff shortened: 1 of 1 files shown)


staktrace:
I made this change manually (since it's JS-only) in the current dev instance and played around with it. It seems to help on bug 1498024 and fixes the other `Maybe<>` case I described in #167.

staktrace:
Oh I broke sticky highlighting with this. Will fix

staktrace:
Ok updated to fix that.

static/js/context-menu.js
```diff
@@ -184,10 +188,12 @@ $("#file").on("click", "code", function(event) {
                     icon: "search"});
   }
 
-  var symbols = elt.closest("[data-symbols]").attr("data-symbols");
+  var token = elt.closest("[data-symbols]");
+  var symbols = token.attr("data-symbols");
+  var visibleToken = token[0].textContent;
```
emilio: Hmm, doesn't this throw when you're not hovering over something with symbols? `$().attr("foo")` returns undefined, but `$()[0].foo` throws. You should probably move this line under the `if`.

static/js/context-menu.js
```diff
@@ -54,14 +54,18 @@ function symbolsFromString(symbols) {
...
 }
 
-function findReferences(symbols) {
+function findReferences(symbols, visibleToken) {
   symbols = symbolsFromString(symbols);
   if (!symbols.length)
     return $();
   symbols = new Set(symbols);
   return $([...document.querySelectorAll("span[data-symbols]")].filter(span => {
+    if (span.textContent != visibleToken) {
```
emilio: Maybe just `return span.textContent == visibleToken && ...`, though no big deal in any case.

static/js/context-menu.js
staktrace: Yeah, I just assumed that if we're in this part of the code we'd find something. But your suggestion is better, I moved it inside the if.

static/js/context-menu.js
staktrace: Updated.

PR #171, opened 2018-12-14
Bug 1514170 - dont addEventlistener to undefined

mozfreddyb:
This patch binds the result of `$('#panel-accel-enable')[0];` to a variable panelAccelEnable
and only adds an event listener if the variable is truthy.
I'm also changing further uses of the jQuery function to the newly bound variable.

https://bugzilla.mozilla.org/show_bug.cgi?id=1514170

### DIFF
--- a/static/js/panel.js
+++ b/static/js/panel.js
@@ -95,7 +95,10 @@ $(function() {
     } else {
         // Keyboard accelerators disabled, so reflect that state in the checkbox and
         // hide the accelerators. Also don't register the keyboard listeners.
-        $('#panel-accel-enable')[0].checked = false;
+        let panelAccelEnable = $('#panel-accel-enable')[0];
+        if (panelAccelEnable) {
+            panelAccelEnable.checked = false;
+        }
         $('.panel span.accel').hide();
     }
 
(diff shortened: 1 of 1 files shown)


static/js/panel.js
```diff
@@ -118,8 +118,10 @@ $(function() {
     }
 
     // If the user toggles the checkbox let's update the state accordingly.
-    $('#panel-accel-enable')[0].addEventListener('change', () => {
-        var newState = ($('#panel-accel-enable')[0].checked);
+    let panelAccelEnable = $('#panel-accel-enable')[0];
+    if (panelAccelEnable) {
```
emilio: This brace needs to be closed :)

static/js/panel.js
```diff
@@ -118,8 +118,10 @@ $(function() {
...
+        panelAccelEnable.addEventListener('change', () => {
```
emilio: nit: Indentation is broken here.

static/js/panel.js
```diff
@@ -118,8 +118,10 @@ $(function() {
     }
 
     // If the user toggles the checkbox let's update the state accordingly.
-    $('#panel-accel-enable')[0].addEventListener('change', () => {
-        var newState = ($('#panel-accel-enable')[0].checked);
+    let panelAccelEnable = $('#panel-accel-enable')[0];
```
emilio: Given how we're using it, we may as well just use `document.getElementById('panel-accel-enable')`, if you want.

mozfreddyb:
Great lesson in how *not* to create pull requests from GitHub's edit interface. :confounded: 
Thank you for reviewing, @emilio! :slightly_smiling_face:

staktrace:
Patch looks fine, but could you squash the two commits into a single one and force-update the PR? With that I'm happy to merge.

staktrace:
Actually looks like there are other places we use `$('#panel-accel-enable')[0]` that should be updated as well. With this patch you still get the JS error if you load e.g. https://kats.searchfox.org/mozilla-central/source/ with accelerators disabled in local storage, because of the instance on line 98. So if you're going to fix this you might as well fix those instances as well. Thanks!

mozfreddyb:
Thanks for the pointer @staktrace. Please take a look again.

staktrace:
This works, thanks. Would have been better to be consistent with the jquery vs getElementById usage, but whatever.

PR #176, opened 2018-12-18
describe: Teach the descriptions code about Rust module-level comments.

emilio:


tools/src/describe.rs
```diff
@@ -66,6 +76,26 @@ fn describe_readme(contents: &str) -> Option<String> {
         .map(str::to_string)
 }
 
+/// Returns the first Rust-style module doc-comment (`//!`).
+fn describe_from_rust_comment(contents: &str) -> Option<String> {
+    let mut description = String::new();
+    let mut in_comment = false;
+    for line in contents.lines() {
+        if !line.starts_with("//!") {
```
staktrace: Can there be whitespace to the left of the `//!` ? If so you might want to trim left or something here. Otherwise looks good.

emilio: Yeah, this was intentional. Otherwise we'd catch stuff like doc comments for nested modules and such which may not be great.

PR #177, opened 2018-12-19
Add links to file bugs

staktrace:
Depends on https://github.com/mozsearch/mozsearch-mozilla/pull/27

### DIFF
--- a/tools/src/bin/output-file.rs
+++ b/tools/src/bin/output-file.rs
@@ -153,52 +153,53 @@ fn main() {
             }
         };
 
-        let mut panel = if let Some((other_desc, other_path)) = show_header {
-            vec![PanelSection {
+        let mut panel = vec![];
+
+        let mut source_panel_items = vec![];
+        if let Some((other_desc, other_path)) = show_header {
+            source_panel_items.push(PanelItem {
+                title: format!("Go to {} file", other_desc),
+                link: other_path,
+                update_link_lineno: false,
+                accel_key: None,
+            });
+        };
+
+        if !source_panel_items.is_empty() {
+            panel.push(PanelSection {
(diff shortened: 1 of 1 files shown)


staktrace:
Currently deployed to https://kats.searchfox.org/mozilla-central/source

tools/src/bin/output-file.rs
```diff
@@ -164,6 +197,18 @@ fn main() {
                 accel_key: None,
             });
         };
+        if let Some(ref bugzilla) = &bugzilla_data {
```
emilio: nit: no need for the ampersand.

tools/src/bin/output-file.rs
```diff
@@ -24,6 +24,31 @@ use languages::FormatAs;
...
+    json::Json::from_reader(&mut reader).ok()?.into_object()
+}
+
+/// For a given path, looks up the bugzilla product and component and
+/// returns it in a tuple if it could be found. The JSON data format
+/// comes from https://searchfox.org/mozilla-central/rev/47edbd91c43db6229cf32d1fc4bae9b325b9e2d0/python/mozbuild/mozbuild/frontend/mach_commands.py#209-223,243
+/// and is fairly straightforward.
+fn get_bugzilla_component<'a>(bugzilla_data: &'a json::Object, path: &str) -> Option<(&'a str, &'a str)> {
+    let mut path_obj = bugzilla_data.get("paths")?;
+    for path_component in path.split('/') {
```
emilio: Does this need to be different for windows? I guess since we're skipping generated files anyway it may not matter.

staktrace: No, these paths are all from the VM instance, so they will always have forward slashes (assuming that's what you meant by windows differences)

PR #181, opened 2019-01-05
Add some CI

staktrace:
It's 2019, we should have some automated testing


tools/src/tokenize.rs
```diff
@@ -218,7 +216,7 @@ pub fn tokenize_c_like(string: &str, spec: &LanguageSpec) -> Vec<Token> {
             let mut start = start;
             'raw_string: loop {
                 if peek_pos() == string.len() {
-                    writeln!(&mut std::io::stderr(), "Unterminated raw string").unwrap();
```
emilio: If you want to preserve the behavior while removing the ugliness you can use `eprintln!`, but `debug!` works as well.

PR #184, opened 2019-01-08
Bug 1515107 - Advance the cur_datum across columns as well

staktrace:
This seems to be a silly error where we don't advance cur_datum properly.
We advance it to the correct line but fail to advance it over to the
correct column. So in some cases we don't find the analysis data for the
token, even though it is present in the dataset.

We have scenarios in simple.rs that are fixed by this patch, and I've pushed a test run to the `kats` channel to verify it fixes bug 1515107.

### DIFF
--- a/tools/src/format.rs
+++ b/tools/src/format.rs
@@ -81,10 +81,18 @@ pub fn format_code(jumps: &HashMap<String, Jump>, format: FormatAs,
 
         let column = (token.start - line_start) as u32;
 
-        // This should never happen, but sometimes the analysis
-        // has bugs in it. This works around them.
+        // Advance cur_datum as long as analysis[cur_datum] is pointing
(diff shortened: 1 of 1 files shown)


tools/src/format.rs
```diff
@@ -81,13 +81,21 @@ pub fn format_code(jumps: &HashMap<String, Jump>, format: FormatAs,
 
         let column = (token.start - line_start) as u32;
 
-        // This should never happen, but sometimes the analysis
-        // has bugs in it. This works around them.
```
emilio: Hmm, should this still not happen? If so it'd be good to know why. In any case this looks good to me.

staktrace: Honestly I don't know why that comment is there. It was added by billm 3 years ago but I think that scenario does legitimately happen and that code is needed to advance `cur_datum` properly.

PR #187, opened 2019-02-26
Remove hard-coded URL stuffs

staktrace:


tools/src/format.rs
```diff
@@ -844,19 +863,18 @@ fn generate_commit_info(tree_name: &str,
     let git = try!(config::get_git(tree_config));
     let hg = match git.hg_map.get(&commit.id()) {
         Some(hg_id) => {
-            let hg_link = format!("<a href=\"{}/rev/{}\">{}</a>", config::get_hg_root(tree_config), hg_id, hg_id);
+            assert!(tree_config.paths.hg_root.is_some());
```
emilio: I don't think this assert is really helpful given unwrap kinda does the same, but no big deal.

tools/src/format.rs
```diff
@@ -872,7 +890,7 @@ fn generate_commit_info(tree_name: &str,
             F::T(format!("<tr><td>commit</td><td>{}</td></tr>", format_rev(tree_name, commit.id()))),
             F::Seq(parents),
             F::Seq(hg),
-            F::T(format!("<tr><td>git</td><td>{}</td></tr>", gitstr)),
+            F::T(gitstr.map_or(String::new(), |g| format!("<tr><td>git</td><td>{}</td></tr>", g))),
```
emilio: I guess ideally there shouldn't be an empty entry, but shrug.

tools/src/format.rs
```diff
@@ -291,8 +291,10 @@ pub fn format_file_data(cfg: &config::Config,
 
     if let Some(ext) = path_wrapper.extension() {
         if ext.to_str().unwrap() == "svg" {
-            let url = format!("{}/raw-file/tip/{}", config::get_hg_root(tree_config), path);
-            try!(output::generate_svg_preview(writer, &url))
+            if let Some(ref hg_root) = tree_config.paths.hg_root {
+                let url = format!("{}/raw-file/tip/{}", hg_root, path);
+                try!(output::generate_svg_preview(writer, &url))
```
emilio: If while you're here you want to untry it (using `?` instead) that'd be great :)

tools/src/format.rs
staktrace: Removed

tools/src/format.rs
staktrace: Done

tools/src/format.rs
staktrace: I left it this way for aesthetic reasons... didn't want to insert an if condition to drop the `F::T` entirely.

PR #189, opened 2019-03-10
Support the mozilla-releases channel with more repos

staktrace:
This depends on mozsearch/mozsearch-mozilla#31. Also when merging this PR we'll need to rebuild/redeploy the AWS lambda task (now we'll have two, one for the `release` channel using `config.json` and one for the `mozilla-releases` channel using `mozilla-releases.json`).


tools/src/bin/web-server.rs
```diff
@@ -247,6 +247,15 @@ fn main() {
         res.send(&output).unwrap();
     };
 
+    {
+        let mut status_out = OpenOptions::new()
+            .append(true)
```
emilio: Should this really append? I don't think it matters given the usage we're giving it but...

staktrace: Yeah the append is actually important because router.py writes to the same file and we don't want to clobber that one if it happens first. I'll add comments at both write locations.

staktrace:
I'll merge this and deploy the new lambda things once today's indexing run is done, or it might break today's run.

PR #192, opened 2019-03-20
Remove redundant checks

marco-c:
I suspect the redundant checks are optimizations, but I'm submitting the PR just in case.

### DIFF
--- a/router/router.py
+++ b/router/router.py
@@ -125,11 +125,13 @@ class SearchResults(object):
 
     def categorize_path(self, path):
         def is_test(p):
+            # Except /unit/, all other paths contain the substring 'test', so we can exit early
+            # in case it is not present.
             if '/unit/' in p:
                 return True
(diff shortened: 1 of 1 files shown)


staktrace:
Yeah I imagine this was done for performance reasons. The checks were added in https://github.com/mozsearch/mozsearch/commit/7db0a83b7364e22e2a90f35e7c2cba72b386d696 when the list of strings was expanded. Presumably the goal is to early-exit when `test` isn't in `p` to avoid doing all the string compares, but the `unit` check needs to happen before that. Unless there's a strong argument to remove this I'd rather leave it as-is. Although removing just the `/unit/` check from the big return statement should be strictly a win, since if that matches we will already have taken the early-exit codepath.

marco-c:
I agree with keeping it, I've added a comment to explain that it is an optimization.

staktrace:
Thanks!

PR #193, opened 2019-03-22
Add Kotlin support

makotokato:
Some GV's tests and mobile projects use Kotlin, so we should format Kotlin's source files.

Kotlin lang's keyword list is the following URL.
- http://kotlinlang.org/docs/reference/keyword-reference.html


staktrace:
Thanks!

PR #195, opened 2019-04-10
Bug 1543473 - Allow the rust web-server to find files inside subrepos…

staktrace:
… as well


tools/src/format.rs
```diff
@@ -475,7 +488,41 @@ pub fn format_path(cfg: &config::Config,
     let commit_obj = try!(git.repo.revparse_single(rev).map_err(|_| "Bad revision"));
     let commit = try!(commit_obj.into_commit().map_err(|_| "Bad revision"));
     let commit_tree = try!(commit.tree().map_err(|_| "Bad revision"));
-    let entry = try!(commit_tree.get_path(Path::new(path)).map_err(|_| "File not found"));
+    let path_obj = Path::new(path);
+    let data = match commit_tree.get_path(path_obj) {
+        Ok(entry) => try!(entry_to_blob(&git.repo, &entry)),
```
emilio: You can use `?` if you want now, though we can probably migrate it altogether if you want :)

tools/src/format.rs
```diff
@@ -475,7 +488,41 @@ pub fn format_path(cfg: &config::Config,
...
+        Err(_) => {
+            // Check to see if this path is inside a submodule
+            let mut test_path = path_obj.parent();
+            let mut submod_file_data = Err("File not found");
+            while let Some(subrepo_path) = test_path {
```
emilio: would this be cleaner written like:

```rust
let mut test_path = path_obj.parent();
loop {
    let subrepo_path = match test_path {
        Some(path) => path,
        None => return Err("File not found"),
    };

    let object = match commit_tree.get_path(subrepo_path) {
        Ok(obj) => obj,
        Err(..) => {
            test_path = subrepo_path.parent();
            continue;
        }
    };

    if e.kind() != Some(git2::ObjectType::Commit) {
        return Err("File not found");
    }

    let subrepo_path = subrepo_path.to_str().ok_or("UTF-8 error")?;
    let subrepo = git.repo.find_submodule(subrepo_path).map_err(|_| "Can't find submodule")?;
    let subrepo = subrepo.open().map_err(|_| "Can't open submodule")?;
    let path_in_subrepo = path_obj.strip_prefix(subrepo_path).map_err(...)?;
    let entry = subrepo.find_commit(e.id()).and_then(..)?;
    break entry_to_blob(&subrepo, &entry)?;
}
```

?

I find that slightly easier to follow, but not a big deal in any case.

staktrace:
Thanks, updated patch with your suggestions. Indeed it is cleaner this way.

PR #197, opened 2019-04-12
Bug 1544046 - Prevent tokenizer crash.

staktrace:
If an .rs file had a `r#` sequence with no `"` characters following it,
the tokenizer would run off the end of the file and panic. This can
happen in valid rust files thanks to the magic of macros.


staktrace:
Will merge this for now since this is blocking searchfox from updating the index, but happy to make changes in a followup.

PR #202, opened 2019-04-18
Update livegrep for native submodule support

staktrace:
This is currently deployed to kats.searchfox.org. Note that this requires reprovisioning the AMIs.


staktrace:
> Should we make the comm-central mozilla/ subdir an actual git submodule?

Maybe, yeah. I thought there was some reason it was hard but I don't remember what it was now. Anyway I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1545891 to do this (or at least try).

> Thanks for doing this, you are amazing :)

You're welcome, thanks for all the reviews! :)

PR #209, opened 2019-04-21
Index rust stdlib source

staktrace:
Rebased version of #123. Depends on https://github.com/mozsearch/mozsearch-mozilla/pull/40 which downloads the src and analysis data from taskcluster.

### DIFF
--- a/scripts/rust-analyze.sh
+++ b/scripts/rust-analyze.sh
@@ -15,11 +15,17 @@ TREE_NAME=$2
 
 if [ -d "$OBJDIR" ]; then
   ANALYSIS_DIRS="$(find $OBJDIR -type d -name save-analysis)"
-  if [ "x$ANALYSIS_DIRS" != "x" ]; then
-    $MOZSEARCH_PATH/tools/target/release/rust-indexer \
-      "$FILES_ROOT" \
-      "$INDEX_ROOT/analysis" \
-      "$OBJDIR" \
-      $ANALYSIS_DIRS
+  if [ "x$ANALYSIS_DIRS" = "x" ]; then
+    exit 0 # Nothing to analyze really.
   fi
+  # If we have rust stdlib sources and analysis data, pick that up too
+  if [ -d "$INDEX_ROOT/rustlib" ]; then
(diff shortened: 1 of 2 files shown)


scripts/rust-analyze.sh
```diff
@@ -15,11 +15,14 @@ TREE_NAME=$2
...
-  if [ "x$ANALYSIS_DIRS" != "x" ]; then
-    $MOZSEARCH_PATH/tools/target/release/rust-indexer \
-      "$FILES_ROOT" \
-      "$INDEX_ROOT/analysis" \
-      "$OBJDIR" \
-      $ANALYSIS_DIRS
+  if [ "x$ANALYSIS_DIRS" = "x" ]; then
+    exit 0 # Nothing to analyze really.
   fi
+  ANALYSIS_DIRS="$ANALYSIS_DIRS $(find $INDEX_ROOT/rustlib -type d -name analysis)"
```
emilio: So this will probably fail if there's no `rustlib`, right? Probably shouldn't... But then we'd need to make the stdlib argument optional to the indexer. May be worth filing a bug for this?

staktrace: I wasn't planning on making this optional - if we don't get a rustlib then yeah, it would just fail. Do you have a particular scenario you're thinking of that might require us to make it optional?

emilio: I was thinking on getting Servo to be indexed and such, or maybe the rust compiler itself, some rustc folks were interested in setting up a searchfox instance a while ago.

But none of them are problems that we have now, so it can just be fixed-up if / when that happens if you prefer.

staktrace: Oh I misunderstood your original point - that this will fail if we are indexing a repo which has rust analysis data, but no rustlib data. Which would be true of Servo/rustc, but is also true of mozilla-beta at the moment. I only tested this on the `config.json` repos which doesn't include beta, so I didn't catch it in testing.

I'll update the patch to make this more robust against that scenario as well.

emilio: Right, thanks!

PR #212, opened 2019-04-26
Bug 1533802 - Sticky nesting for C++, Rust, JS r=kats

asutherland:
This implements position: sticky display of lines for "source" analysis lines
for which we have a "nestingRange" available.  Support for emitting
"nestingRange" is added to the C++ (clang), Rust, and JS analyzer logic.

However, the Rust analyzer changes are very limited because the "save-analysis"
data which forms the basis of the Rust analysis is extremely limited.  It
effectively already is being provided with the result of an analysis pass which
has had all AST information stripped away.  There did not seem to be any
readily available mechanism in the "RLS" to attempt to reconstitute an AST and
relate it to the analysis data.  The best hack I was able to come up with was
to attempt to recursively build up a span covering all of the 'def' children of
a 'def' or an 'impl'.  This means that nesting for rust is largely limited to
"struct" definitions and for inherent "impl" blocks, with the nestingRange only
covering up until the start of the last "fn" in the "impl".


asutherland:
So this seems to be sufficiently working for C++/Rust/JS.  As noted in the commit message (and which I've updated into the root comment), the rust stuff is pretty hacky but I didn't see any other immediately available options.  I'll try and spin a new dev run now so I know how to do that.

asutherland:
And I'll stop force-pushing from here on out.

tools/src/bin/rust-indexer.rs
```diff
@@ -200,13 +200,142 @@ fn def_kind_to_human(kind: DefKind) -> &'static str {
...
+}
+
+/// Given two spans, create a new super-span that encloses them both if the files match.  If the
+/// files don't match, just return the first span as-is.
+fn union_spans(a: &data::SpanData, b: &data::SpanData) -> data::SpanData {
+    if a.file_name != b.file_name {
+        return a.clone();
+    }
+
+    let (start_byte, start_line, start_col) = if a.byte_start < b.byte_start {
```
emilio: You can call them `byte_start`, `line_start`, `column_start`, ...

tools/src/bin/rust-indexer.rs
```diff
@@ -200,13 +200,142 @@ fn def_kind_to_human(kind: DefKind) -> &'static str {
...
+    };
+
+    let (end_byte, end_line, end_col) = if a.byte_end > b.byte_end {
+        (a.byte_end, a.line_end, a.column_end)
+    } else {
+        (b.byte_end, b.line_end, b.column_end)
+    };
+
+
+    data::SpanData {
```
emilio: Then use:

```rust
data::SpanData {
    byte_start,
    line_start,
    column_start,
    ...
}
```

tools/src/bin/rust-indexer.rs
```diff
@@ -200,13 +200,142 @@ fn def_kind_to_human(kind: DefKind) -> &'static str {
...
+    for id in &def.children {
+        // It should already be the case that the children are in the same krate, but better safe
+        // than sorry.
+        if id.krate != def.id.krate {
+            continue;
+        }
+        let kid = defs.get(file_analysis, *id);
+
+        if let Some(ref kid) = kid {
+            if kid.children.len() > 0 {
```
emilio: You don't need this check right? If the children is empty then it'd just return `kid.span`.

tools/src/bin/rust-indexer.rs
```diff
@@ -200,13 +200,142 @@ fn def_kind_to_human(kind: DefKind) -> &'static str {
...
+            }
+        }
+    }
+
+    span
+}
+
+/// Given a list of ids of defs, run recursive_union_spans_of_def on all of them and union up the
+/// result.  Necessary for when dealing with impls.
+fn union_spans_of_defs(initial_span: &data::SpanData, ids: &Vec<data::Id>, file_analysis: &data::Analysis,
```
emilio: nit: Please run rustfmt on this.

Also `&[data::Id]`, `&Vec` never makes sense.

tools/src/bin/rust-indexer.rs
```diff
@@ -200,13 +200,142 @@ fn def_kind_to_human(kind: DefKind) -> &'static str {
...
+                       defs: &Defs) -> data::SpanData {
+    let mut span = initial_span.clone();
+    for id in ids {
+        let kid = defs.get(file_analysis, *id);
+
+        if let Some(ref kid) = kid {
+            if kid.children.len() > 0 {
```
emilio: Same comment about the len() > 0 thing not being needed.

tools/src/bin/rust-indexer.rs
```diff
@@ -297,6 +450,14 @@ fn analyze_file(
         }
     };
 
+    let mut source_file = match File::open(file_name.clone()) {
+        Ok(f) => f,
+        Err(_) => {
+            error!("Unable to open source file for snippet purposes: {}", file_name.display());
```
emilio: When does this happen? And should it really avoid the rest of the analysis from running?

tools/src/bin/rust-indexer.rs
```diff
@@ -281,6 +416,24 @@ fn read_existing_contents(
...
+    reader: &mut File,
+    span: &data::SpanData,
+) -> String {
+    reader.seek(std::io::SeekFrom::Start(span.byte_start.into())).unwrap();
+    let len = (span.byte_end - span.byte_start) as usize;
+    let mut buffer: Box<[u8]> = vec![0; len].into_boxed_slice();
+    match reader.read_exact(&mut buffer) {
+        Ok(_) => String::from_utf8(buffer.to_vec()).unwrap(),
+        Err(e) => {
+            println!("ERROR: Unable to read file: {:?}", e);
```
emilio: This should probably `error!(`.

tools/src/bin/rust-indexer.rs
```diff
@@ -281,6 +416,24 @@ fn read_existing_contents(
...
+            String::from("")
```
emilio: String::new()

tools/src/bin/rust-indexer.rs
```diff
@@ -281,6 +416,24 @@ fn read_existing_contents(
...
+/// covers multiple lines, then whatever newline delimiters the file has will be included.
+fn extract_span_from_source(
+    reader: &mut File,
+    span: &data::SpanData,
+) -> String {
+    reader.seek(std::io::SeekFrom::Start(span.byte_start.into())).unwrap();
+    let len = (span.byte_end - span.byte_start) as usize;
+    let mut buffer: Box<[u8]> = vec![0; len].into_boxed_slice();
+    match reader.read_exact(&mut buffer) {
+        Ok(_) => String::from_utf8(buffer.to_vec()).unwrap(),
```
emilio: What guarantees that this doesn't fail? This should probably just return a `Result<>` and skip the impl if we fail to read it, or something.

emilio: Also, `buffer.into_vec()` to avoid copying.

tools/src/bin/rust-indexer.rs
```diff
@@ -297,6 +450,14 @@ fn analyze_file(
         }
     };
 
+    let mut source_file = match File::open(file_name.clone()) {
```
emilio: I'm ~sure this can just be `match File::open(&file_name) {`...

tools/src/bin/rust-indexer.rs
asutherland: As you can tell by the println below, I guess I'm not sure exactly how we want to handle errors.  I was viewing the file not existing as indicative of a major configuration error that should fail fast.  Like how can we be indexing a file if we don't have the source?  (But maybe the fact that all of this analysis is being reverse-engineered from crate analysis stuff means it's possible?)

asutherland:
Thanks for all the rust feedback!  I'm still very much at the "cargo cult what I see in the existing code and randomly google for what I don't see", so I deeply appreciate the feedback.  (Also, apologies for all the beginner mistakes.  I have the Blandy/Orendorff book, but have clearly not gotten very far into it.  Arguably it would have been wiser to take some time at the start of the rust indexing delve in order to level up my understanding and decrease my stumbling.)

tools/src/bin/rust-indexer.rs
asutherland: This actually is somewhat interesting because a problem I had last time I was trying to do some searchfox hacking was that vagrant was constantly giving me short reads.  But agreed on consistency and that silent errors on short reads aren't helpful either.

tools/src/file_format/analysis.rs
```diff
@@ -103,6 +111,7 @@ pub struct AnalysisSource {
     pub pretty: String,
     pub sym: Vec<String>,
     pub no_crossref: bool,
+    pub nesting_range: SourceRange,
 }
 
 impl AnalysisSource {
```
staktrace: The `AnalysisSource::merge` function I think also needs to be updated so that it unions the nesting_range of `other` into `self`. Consider for example c++ code like so:
```c++
void func()
#ifdef LINUX
{ linux_body; }
#else 
{ other_body; }
#endif 
```

In this case, we'll get the analysis data output by clang for two different platforms, which will have different nesting ranges for the symbol, and then we'll run through this merge code to merge those source entries. Really we want to treat the union of the different function bodies as being the impl of `func`.

tools/src/format.rs
```diff
@@ -111,6 +145,25 @@ pub fn format_code(jumps: &HashMap<String, Jump>, format: FormatAs,
...
             (&tokenize::TokenKind::Identifier(None), Some(d)) => {
+                // If this symbol starts a relevant nesting range and we haven't already pushed a
+                // symbol for this line, push it onto our stack.  A range is "relevant" if it
+                // its end line is not on the current line or the next line and therefore will
+                // actually trigger the "position:sticky" display scenario.
+                for a in d.iter() {
+                    let nests = match (a.nesting_range.start_lineno, nesting_stack.last()) {
+                        (0, _) => false,
+                        (_, None) => true,
+                        (a_start, Some(top)) if a_start == cur_line &&
```
staktrace: This might be a bit clean to write as:

```rust
(a_start, Some(top)) => {
  a_start == cur_line &&
  a_start != top.nesting_range.start_lineno &&
  a.nesting_range.end_lineno > cur_line + 1
}

File: tools/src/bin/rust-indexer.rs (hunk 6 above)
Comment (staktrace) [reply tools/src/bin/rust-indexer.rs]: Often times we get analysis data that has "weird" filenames (see e.g. https://github.com/rust-lang/rls/issues/1339) so I think we probably do want to handle this gracefully

File: tools/src/bin/rust-indexer.rs [hunk 13]
```diff
@@ -349,20 +510,43 @@ fn analyze_file(
...
-        visit(&mut dataset, AnalysisKind::Def, &def.span, &qualname, &def, parent.as_ref().map(|p| &*p.qualname))
+        let nested_span = recursive_union_spans_of_def(def, &file_analysis, &defs);
+        let maybe_nested = ignore_boring_spans(&nested_span);
+        visit_def(&mut dataset, AnalysisKind::Def, &def.span, &qualname, &def,
+                  parent.as_ref().map(|p| &*p.qualname), maybe_nested)
+    }
+
+    // We want to expose impls as "def,namespace" with an inferred nesting_range for their
+    // contents.  I don't know if it's a bug or just a dubious design decision, but the impls all
+    // have empty values and no names, so to get a useful string out of them, we need to extract
```
staktrace: Would be good to file an issue against https://github.com/rust-lang/rls for this

tools/src/bin/rust-indexer.rs
```diff
@@ -373,13 +557,14 @@ fn analyze_file(
...
         };
-        visit(
+        visit_def(
             &mut dataset,
             AnalysisKind::Use,
             &ref_.span,
             &def.qualname,
             &def,
             /* context = */ None, // TODO
+            /* nesting = */ None, // Only used for defs.
```
staktrace: Comment "Only used for defs" here doesn't make sense to me

tools/src/bin/rust-indexer.rs
asutherland: I'm making this into an Option<> and having us skip the `impls` traversal that depends on this to retrieve the impl name.

tools/src/bin/rust-indexer.rs
asutherland: I split this fn in two because the Result error types for std::io::Error and std::string::FromUtf8Error didn't line up which made it hard to use `?`.  So the lower-level one retuns io::Result and the outer returns Option<String>.

tools/src/bin/rust-indexer.rs
asutherland: rustfmt wanted to make some unrelated changes so my plan is to rebase this patch stack as:
1. rustfmt run on tree without any changes.
2. squashed changes with rustfmt run again.

I'll also push all of my changes before undertaking that change so that you can see the deltas made without any rustfmt going on.

tools/src/format.rs
asutherland: Nice cleanup, thanks!  I've spent too long away from my scheme/lisp roots! ;)

tools/src/bin/rust-indexer.rs
asutherland: So the information is actually missing at the save-analysis generation in the compiler, so it's a rust-lang/rust issue.  I filed https://github.com/rust-lang/rust/issues/60872 to track this.

tools/src/bin/rust-indexer.rs
asutherland: Okay, I think I've done this now...

asutherland:
Okay, so now there are 2 parts: part 1 is running rustfmt on only the rust code I end up touching and part 2 is just my changes (with rustfmt run again afterwards).  I'm going to try and do the dev.searchfox.org thing now (or get it started)>

staktrace:
>  I'm going to try and do the dev.searchfox.org thing now (or get it started)

So sadly this didn't work because the update script that checks out the repo assumes your github clone is called "mozsearch" and not "searchfox"

```
$ tail update-log
+ sudo install js /usr/local/bin
+ sudo install libnspr4.so libplc4.so libplds4.so /usr/local/lib
+ sudo ldconfig
+ popd
~
+ rm -rf mozsearch
+ git clone -b scope-nesting https://github.com/asutherland/searchfox
Cloning into 'searchfox'...
+ pushd mozsearch
./update.sh: line 44: pushd: mozsearch: No such file or directory
```

I'll shut down the indexer and file a bug to fix this.

asutherland:
Thanks for investigating this and fixing the underlying issue!  Your strong ownership of and professionalism on searchfox (/ mozsearch ;) are inspiring.

asutherland:
@staktrace @emilio So there's a clean run up at https://dev.searchfox.org/ now where the C++ looks very good, the JS looks good[1], and the rust is better than nothing.  It looks like my "impl" hack isn't working, so I'll iterate on that, potentially adding another commit and triggering another "dev" channel re-indexing when I do.  I do think we'll need to pursue explicitly enhancing the save-analysis stuff to include significantly more AST info before the rust stuff is actually useful.

1: https://dev.searchfox.org/mozilla-central/source/browser/modules/Sanitizer.jsm#46 should have Sanitizer be sticky.

staktrace:
I just tried it out on some C++ code and it does indeed work pretty well. Then I went to nsTArray.h and found some weirdness... if you to e.g. https://dev.searchfox.org/mozilla-central/rev/94c6b5f06d2464f6780a52f32e917d25ddc30d6b/xpcom/ds/nsTArray.h#1692 there's a whole stack of declarations that are still stuck even though they shouldn't be.

staktrace:
(Maybe because the function at https://dev.searchfox.org/mozilla-central/rev/94c6b5f06d2464f6780a52f32e917d25ddc30d6b/xpcom/ds/nsTArray.h#2391 is declared and implemented separately in the same file and we're unioning the ranges?)

emilio:
I see the sticky line half-under the search bar. The search bar is a bit higher than the CSS variable.

The hard-coded pixel value will also likely not play well with any kind of zoom.

asutherland:
tl;dr: The nestingRange is being emitted for declarations when it should only be emitted for definitions and the emission isn't sanity checking that the range is in the same file.  I think the union logic is probably right in general, but because we use the "loc" as the start of the region for nesting and only require that the nestingRange.start is >= loc, when we erroneously tie the nestingRange to a declaration even in the same file, we effectively end up unioning the ranges.  (This was a change I made so that the nestingRange could point at the braces for cool feature creep purposes, which changed our constraints in format.rs.)

I will make sure we only emit nestingRange if the "source" we're emitting is the definition instance.  I'll also add defensive sanity-checking that the loc and nestingRange are in the same file.

### Investigation ###

Aside: Holy cow, so "wc -l nsTArray.h" for the analysis says there are 1,936,965 lines in there.  Which makes sense because we have target entries for every permutation of T for nsTArray<T>.

Grepping that down to just "source" lines shows that we indeed then union all of those permutations into a single "source" which is very hard to scroll through.  Installing "jq" and then doing `grep "source" nsTArray.h | jq -c 'select(.nestingRange) | {loc, pretty, nestingRange}' | less` provides a more sane experience.

Looking at `{"loc":"00316:16-36","pretty":"function Gecko_ClearPODTArray","nestingRange":"1280:48-1287:0"}` for example, it appears that the problem is that at line 316 we have a declaration in nsTArray.h of `extern "C" void Gecko_ClearPODTArray(void* aArray, size_t aElementSize, size_t aElementAlign);` but the definition is in GeckoBindings on lines 1279-1287, with the opening brace on 1280.

asutherland:
So the other variation I'm trying to deal with now is that [nsTArray.h:1337](https://searchfox.org/mozilla-central/rev/f8b11433159cbc9cc80500b3e579d767473fa539/xpcom/ds/nsTArray.h#1336)
```cpp
  template <class Item, typename ActualAlloc = Alloc>
  elem_type* ReplaceElementsAt(index_type aStart, size_type aCount,
                               const Item* aArray, size_type aArrayLen);
```

is turning into `"syntax":"decl,def,function"` from a simplified line of
```json
{"loc":"01337:13-30","pretty":"function nsTArray_Impl::ReplaceElementsAt","syntax":"decl,def,function","nestingRange":"2207:18-2222:0"}
```

as produced via `grep "source" nsTArray.h | jq -c 'select(.nestingRange) | {loc, pretty, syntax, nestingRange}' | less`.

So presumably we're merging "source" entries that have syntax values of "decl,function" and "def,function".  The clang indexer can't actually output a syntax of "decl,def,function" on its own that I'm aware of.  I've been trying to re-create the problem in big_header.h after seeing the import list at the top of nsTArray.h and deciding adding nsTArray.h to the test repo was not going to be sufficiently straightforward.

Oh, and here's the direct clang++ ast-dump of the header (aka not through a build system):
```
| |-ClassTemplateDecl 0x24ab0f0 <line:1336:3, line:1339:7> line:1337:12 is_base_of
| | |-TemplateTypeParmDecl 0x24aaf68 <line:1336:12, col:21> col:21 referenced typename depth 0 index 0 _Base
| | |-TemplateTypeParmDecl 0x24aafd8 <col:28, col:37> col:37 referenced typename depth 0 index 1 _Derived
| | `-CXXRecordDecl 0x24ab068 <line:1337:5, line:1339:7> line:1337:12 struct is_base_of definition
| |   |-DefinitionData empty aggregate standard_layout trivially_copyable pod trivial literal has_constexpr_non_copy_move_ctor can_const_default_init
| |   | |-DefaultConstructor exists trivial constexpr needs_implicit defaulted_is_constexpr
| |   | |-CopyConstructor simple trivial has_const_param needs_implicit implicit_has_const_param
| |   | |-MoveConstructor exists simple trivial needs_implicit
| |   | |-CopyAssignment trivial has_const_param needs_implicit implicit_has_const_param
| |   | |-MoveAssignment exists simple trivial needs_implicit
| |   | `-Destructor simple irrelevant trivial needs_implicit
| |   `-CXXRecordDecl 0x24ab3d8 <col:5, col:12> col:12 implicit struct is_base_of
```

asutherland:
Okay, managed to reproduce the following non-merged analysis output:
```json
{"loc":"00064:12-52","source":1,"nestingRange":"81:35-89:0","syntax":"def,function","pretty":"function WhatsYourVector_impl::forwardDeclaredTemplateThingInlinedBelow","sym":"_ZN20WhatsYourVector_impl40forwardDeclaredTemplateThingInlinedBelowEPKT_"}
{"loc":"00064:12-52","source":1,"syntax":"decl,function","pretty":"function WhatsYourVector_impl::forwardDeclaredTemplateThingInlinedBelow","sym":"_ZN20WhatsYourVector_impl40forwardDeclaredTemplateThingInlinedBelowEPKT_"}
{"loc":"00064:12-52","source":1,"syntax":"decl,function","pretty":"function WhatsYourVector_impl::forwardDeclaredTemplateThingInlinedBelow","sym":"_ZN20WhatsYourVector_implIN7outerNS5HumanE9CoolAllocE40forwardDeclaredTemplateThingInlinedBelowEPKT_"}
{"loc":"00064:12-52","source":1,"syntax":"decl,function","pretty":"function WhatsYourVector_impl::forwardDeclaredTemplateThingInlinedBelow","sym":"_ZN20WhatsYourVector_implIN7outerNS9SuperheroE9CoolAllocE40forwardDeclaredTemplateThingInlinedBelowEPKT_"}
```

From the following mess.  Interestingly, the decl's are somewhat superfluous, although obviously the def is the unexpected, no good, very bad type definition:
```cpp
class CoolAlloc {
public:
  static void ChewGumAndAllocateMemory() {
    // nom nom nom
  }
};

template <class T>
class JokeBase {
  int walkIntoBar() {
    int i = 0;
    return i--;
  }
};

template <class T, class Alloc>
class WhatsYourVector_impl
  : public JokeBase<T> {
public:
  typedef T* elem_type;

  elem_type mStorage;

  WhatsYourVector_impl(elem_type thing)
  : mStorage(thing) {
  }

  int forwardDeclaredThingInlinedBelow();

  template <class Item, typename ActualAlloc = Alloc>
  elem_type forwardDeclaredTemplateThingInlinedBelow(
    const Item* aThing);

};

template <class T, class Alloc>
int WhatsYourVector_impl<T, Alloc>::forwardDeclaredThingInlinedBelow() {
  int i = 0;
  i++;
  i--;
  return i;
}


template <typename T, class Alloc>
template <class Item, typename ActualAlloc>
auto WhatsYourVector_impl<T, Alloc>::forwardDeclaredTemplateThingInlinedBelow(
  const Item* aThing) -> elem_type {
  int i = 0;
  if (aThing)  {
    i++;
  }
  i--;
  ActualAlloc::ChewGumAndAllocateMemory();
  return mStorage;
}

template <class T>
class WhatsYourVector: public WhatsYourVector_impl<T, CoolAlloc> {
public:
  typedef T* elem_type;

  WhatsYourVector(elem_type thing)
  : WhatsYourVector_impl<T, CoolAlloc>(thing) {
  }

};
```

asutherland:
And here's what the AST excerpt for that looks like:
```
| | |-FunctionTemplateDecl 0x3a10240 <line:63:3, line:65:23> line:64:13 forwardDeclaredTemplateThingInlinedBelow
| | | |-TemplateTypeParmDecl 0x3a0ff10 <line:63:13, col:19> col:19 class depth 0 index 0 Item
| | | |-TemplateTypeParmDecl 0x3a0ff78 <col:25, col:48> col:34 typename depth 0 index 1 ActualAlloc
| | | | `-TemplateArgument type 'CoolAlloc':'CoolAlloc'
| | | |-CXXMethodDecl 0x3a101a8 <line:64:3, line:65:23> line:64:13 forwardDeclaredTemplateThingInlinedBelow 'WhatsYourVector_impl<outerNS::Superhero, CoolAlloc>::elem_type (const Item *)'
| | | | `-ParmVarDecl 0x3a100b8 <line:65:5, col:17> col:17 aThing 'const Item *'
| | | `-CXXMethodDecl 0x3a155b8 <line:80:1, line:89:1> line:64:13 used forwardDeclaredTemplateThingInlinedBelow 'WhatsYourVector_impl<outerNS::Superhero, CoolAlloc>::elem_type (const outerNS::Couch *)'
| | |   |-TemplateArgument type 'outerNS::Couch'
| | |   |-TemplateArgument type 'CoolAlloc'
| | |   |-ParmVarDecl 0x3a154c8 <line:65:5, col:17> col:17 used aThing 'const outerNS::Couch *'
| | |   `-CompoundStmt 0x3a1d980 <line:81:36, line:89:1>
| | |     |-DeclStmt 0x3a1d720 <line:82:3, col:12>
| | |     | `-VarDecl 0x3a1d6c0 <col:3, col:11> col:7 used i 'int' cinit
| | |     |   `-IntegerLiteral 0x3a04d60 <col:11> 'int' 0
| | |     |-IfStmt 0x3a1d7d8 <line:83:3, line:85:3>
| | |     | |-ImplicitCastExpr 0x3a1d770 <line:83:7> 'bool' <PointerToBoolean>
| | |     | | `-ImplicitCastExpr 0x3a1d758 <col:7> 'const outerNS::Couch *' <LValueToRValue>
| | |     | |   `-DeclRefExpr 0x3a1d738 <col:7> 'const outerNS::Couch *' lvalue ParmVar 0x3a154c8 'aThing' 'const outerNS::Couch *'
| | |     | `-CompoundStmt 0x3a1d7c0 <col:16, line:85:3>
| | |     |   `-UnaryOperator 0x3a1d7a8 <line:84:5, col:6> 'int' postfix '++'
| | |     |     `-DeclRefExpr 0x3a1d788 <col:5> 'int' lvalue Var 0x3a1d6c0 'i' 'int'
| | |     |-UnaryOperator 0x3a1d810 <line:86:3, col:4> 'int' postfix '--'
| | |     | `-DeclRefExpr 0x3a1d7f0 <col:3> 'int' lvalue Var 0x3a1d6c0 'i' 'int'
| | |     |-CallExpr 0x3a1d8f8 <line:87:3, col:41> 'void'
| | |     | `-ImplicitCastExpr 0x3a1d8e0 <col:3, col:16> 'void (*)()' <FunctionToPointerDecay>
| | |     |   `-DeclRefExpr 0x3a1d8b0 <col:3, col:16> 'void ()' lvalue CXXMethod 0x3a015f8 'ChewGumAndAllocateMemory' 'void ()'
| | |     `-ReturnStmt 0x3a1d970 <line:88:3, col:10>
| | |       `-ImplicitCastExpr 0x3a1d958 <col:10> 'WhatsYourVector_impl<outerNS::Superhero, CoolAlloc>::elem_type':'outerNS::Superhero *' <LValueToRValue>
| | |         `-MemberExpr 0x3a1d928 <col:10> 'WhatsYourVector_impl<outerNS::Superhero, CoolAlloc>::elem_type':'outerNS::Superhero *' lvalue ->mStorage 0x3a0fc30
| | |           `-CXXThisExpr 0x3a1d918 <col:10> 'WhatsYourVector_impl<outerNS::Superhero, CoolAlloc> *' this
```

compare with the AST for its non-template friend just above it:
```
| | |-CXXMethodDecl 0x3a0fe68 <line:61:3, col:40> col:7 forwardDeclaredThingInlinedBelow 'int ()'
```

asutherland:
That seems to have done it locally, I'm starting the try server dance.

asutherland:
@staktrace @emilio Okay, the try server dance was successful.  nsTArray.h now looks fine, in particular the example line :kats gave of https://dev.searchfox.org/mozilla-central/source/xpcom/ds/nsTArray.h#1692

I addressed the searchbar positioning issues by changing from using position: fixed to using flexbox and have tested it in Firefox on linux/os x/windows and it seems fine with that and zoom used.

static/css/mozsearch.css
```diff
@@ -167,6 +167,42 @@ mark {
...
+ * New fixed header mechanism!
+ *
+ * Previously, the search box stayed stuck at the top of the screen via
+ * use of "position: fixed".  This was problematic with browser default
+ * anchor positioning, but became untenable when we wanted to start using
+ * "position: sticky" to show nesting scopes because we needed to specify "top"
+ * positions for each nested sticky level, and it was hard to know for sure how
+ * big the search box would actually be without involving JS.
+ *
+ * So our solition is to use flexbox to create a non-scrolling fixed-header,
```
staktrace: typo: solition

clang-plugin/MozsearchIndexer.cpp
```diff
@@ -1202,7 +1202,21 @@ class IndexConsumer : public ASTConsumer,
       PrettyKind = "function";
       PeekRange = getFunctionPeekRange(D2);
 
-      if (D2->isThisDeclarationADefinition()) {
+      // Only emit the nesting range if:
+      // - This is a definition AND
+      // - This isn't a template instantiation.  Function templates'
+      //   instantiations can end up as a definition with a Loc at their point
+      //   of declaration but with the CompountStmt of the template's
```
staktrace: typo: CompountStmt

asutherland:
In preparation for landing, I've prepared https://phabricator.services.mozilla.com/D33701 to include the clang indexer changes.

While it would potentially be cool to uplift it to branches other than nightly, in practice, we don't have to because the only schema change was to add the nestingRange.  No nestingRange means no nesting.  We can uplift as appropriate, although if older branches don't use clang 8 maybe they will get sad.

PR #217, opened 2019-05-17
Bug 1552381 - Explicitly specify the folder to clone into

staktrace:


staktrace:
I'll have to rebuild the AMIs with this change since the update.sh script gets baked in as part of provisioning.

PR #221, opened 2019-06-10
Bug 1511275 - Unify line numbers and source code listings in DOM. r=kats

asutherland:
This also includes elements to fix bug 1557139.

This patch encompasses these changes (which can be split out):

- Have the test 'searchfox' repo generate and use blame information.  It's a
  bit difficult to iterate on blame and diff display without blame and the
  ability to thereby derive diff links.  I've modified the test searchfox
  setup script to run blame generation for the self-same searchfox repo when
  indexer setup is run.  The major downside to this is that if you have any
  local changes to the searchfox codebase that haven't been integrated into
  the blame data, then output-file may panic when the line counts aren't the
  same.  This means an iteration loop looks like:
  - outside the vagrant VM
    - `git commit -a --amend --reset-author`
  - inside the vagrant VM
    - `/vagrant/infrastructure/indexer-setup.sh /vagrant/tests config.json ~/index`
    - `rm -rf ~/index/tests/analysis`
    - `/vagrant/infrastructure/indexer-run.sh /vagrant/tests ~/index`
    - `/vagrant/infrastructure/web-server-run.sh /vagrant/tests ~/index ~`
  I think there's probably a better status quo here that's possible.  Maybe the
  `test` test repo could use `git filter-branch` so only changes to the test
  files generate the need for a blame rebuild.

- Change the HTML structure as it relates to line number display.
  - Pre-patch, the source code display is a 2-column 1-row table where the left
    column contains the line numbers and the right column contains the source
    code lines.  Additionally, the blame info is a child of the line number span
    which absolutely positions itself to the left of the line number.
  - Post-patch the blame, line number, and source line contents live together in
    a flexbox div.  The blame and line-number have a fixed width whereas the
    source line grows to encompass the entire line.
  - The primary motivating factor is to make the position:sticky display both
    more understandable and more usable by having the line numbers (and blame)
    stick with the source line.
  - A secondary but still important concern is improving the accessibility of
    searchfox.
  - The line numbers live as generated content to avoid having them included in
    copied and pasted selections of code.

- Fix bugs in the code-highlighter.js logic responsible for highlighting lines
  of code based on selection triggered by clicking on source code line numbers
  and optionally holding control and/or shift.  This broke when the nesting
  divs were created for the position: sticky logic.

- Implement a "scroll" event handler that marks the position:sticky stuck nodes
  as ".stuck" and ".last-stuck" so that:
  - Visual presentation can be improved.  Previously position:sticky-able
    source lines were given a light background color so that they would
    hopefully distinguish themselves when stuck.  This didn't work and added
    visual noise to normal source code display.  By dynamically marking stuck
    lines of code with a class, we can choose a much more obvious styling that
    is only applied when the position:sticky lines become stuck.  The downside
    is there may be a performance impact and the application of the style may
    not be immediate.
  - The behavior of clicking on a line number when the line is stuck can be
    changed from the existing selection logic to instead scroll up to the line
    number that was clicked on.  For this reason, the stuck logic is (perhaps)
    confusingly include in code-highlighter.js.

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -119,37 +119,48 @@ td.code {
 td#line-numbers {
     padding: 0;
 }
+
+.source-line-with-number {
+  padding: 0rem 1rem;
+  line-height: 1.3;
+  display: flex;
+}
+
+/* In order to display the line numbers and have them visible to screen readers
+   but not appear in any copy-and-pasted selection, we create a before
+   pseudo-element with generated content. */
+.line-number::before {
+    content: attr(data-line-number);
+}
 .line-number {
-    display: block;
+    display: inline-block;
     cursor: pointer;
-    -moz-user-select: none;
-    -o-user-select: none;
-    -khtml-user-select: none;
-    -webkit-user-select: none;
-    -ms-user-select: none;
     user-select: none;
     text-align: right;
     padding: 0 0.5rem;
-    position: relative;
+    /* this was originally set on the ".file td:first-child" */
+    color: #aaa;
+    width: 8ex;
 }
 .highlighted,
 .multihighlight {
     background: none repeat scroll 0 0 rgb(255, 255, 204) !important;
 }
-td.code {
-    display: block;
+.source-line {
     margin: 0;
     padding: 0;
-    position: relative;
-    z-index: 1;
+    display: inline-block;
+    white-space: pre;
+    flex-grow: 1;
 }
-td pre code {
-    padding: 0 0 0 1em;
-    margin: 0;
-    padding: 0;
-    display: block;
+.source-line-with-number.stuck {
+  background: #ddf;
 }
-code a {
+.source-line-with-number.last-stuck {
+  background: linear-gradient(#ddf, #ddf 90%, blue);
+}
+
+.source-line a {
     cursor: pointer;
 }
 .code code {
@@ -205,18 +216,10 @@ body {
 
 /**
  * Nesting!
- *
- * Because the search bar thing is position: fixed we need to add an offset to
- * all of our position:sticky "top" directives.  This needs to be the height of
- * the search box.  CSS variables provide us a way to adjust this based on other
- * factors.  Currently the main influencing factor is whether there is a
- * "Showing REVHASH: Commit message" line or if it's empty.  The presence of the
- * line matches up with the "old-rev" class being set on the body, so we use
- * this to decide which hard-coded value to use below.  JS could also update
- * the variable, but we currently don't need to do that.
  */
 
 .nesting-container {
+  display: block;
   position: relative;
 }
 .nesting-depth-1,
@@ -230,47 +233,50 @@ body {
 .nesting-depth-9,
 .nesting-depth-10 {
   position: sticky;
-  background-color: #f8f8f8;
+  /* even with our "stuck" and "last-stuck" classes, it's important to have an
+     opaque background color for these because those classes don't take effect
+     immediately. */
+  background-color: white;
 }
 .nesting-depth-1 {
   top: calc(0 * 12px * 1.3);
-  z-index: 900;
+  z-index: 100;
 }
 .nesting-depth-2 {
   top: calc(1 * 12px * 1.3);
-  z-index: 899;
+  z-index: 99;
 }
 .nesting-depth-3 {
   top: calc(2 * 12px * 1.3);
-  z-index: 898;
+  z-index: 98;
 }
 .nesting-depth-4 {
   top: calc(3 * 12px * 1.3);
-  z-index: 897;
+  z-index: 97;
 }
 .nesting-depth-5 {
   top: calc(4 * 12px * 1.3);
-  z-index: 896;
+  z-index: 96;
 }
 .nesting-depth-6 {
   top: calc(5 * 12px * 1.3);
-  z-index: 895;
+  z-index: 95;
 }
 .nesting-depth-7 {
   top: calc(6 * 12px * 1.3);
-  z-index: 894;
+  z-index: 94;
 }
 .nesting-depth-8 {
   top: calc(7 * 12px * 1.3);
-  z-index: 893;
+  z-index: 93;
 }
 .nesting-depth-9 {
   top: calc(8 * 12px * 1.3);
-  z-index: 892;
+  z-index: 92;
 }
 .nesting-depth-10 {
   top: calc(9 * 12px * 1.3);
-  z-index: 891;
+  z-index: 91;
 }
 
 /* Search results */
@@ -373,7 +379,7 @@ body {
     border-top-left-radius: 0;
     width: auto;
     list-style: none;
-    z-index: 2;
+    z-index: 102;
 }
 .context-menu a {
     display: block;
@@ -388,7 +394,7 @@ body {
 }
 
 .content {
-    padding: 1rem 2rem;
+    padding: 0.5rem 0rem;
 }
 
 /* Breadcrumbs */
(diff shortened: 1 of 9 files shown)


asutherland:
I'm requesting review for a first-pass here.  There are still a few things to be done here (at minimum):
1. Make sure the a11y goals are met.  :Jamie just gave me feedback in https://bugzilla.mozilla.org/show_bug.cgi?id=1511275#c7 so I'll apply that now.
2. I think I may have regressed some padding a little bit.  In order to get the "stuck" display to go edge-to-edge I did some refactoring that may not have been totally applied to the non source-listing views.  This should be corrected.
3. My "stuck" styling is almost certainly not perfect and could use some feedback from peoples.

tools/src/format.rs
```diff
@@ -498,58 +497,55 @@ pub fn format_file_data(
                                class, revs, filespecs, blame_linenos);
             data
         } else {
-            "".to_owned()
+            // If we have no blame data, we want the div here taking up space, but we don't want it
+            // to both screen readers.
```
staktrace: comment here seems to be missing a word or something

static/css/mozsearch.css
```diff
@@ -388,7 +394,7 @@ body {
 }
 
 .content {
-    padding: 1rem 2rem;
+    padding: 0.5rem 0rem;
```
staktrace: This removes the padding on the dir listing pages too (e.g. https://dev.searchfox.org/mozilla-central/source) which seems slightly undesirable. At least I prefer those listing pages with the padding and it looks weird if the the entries are shoved up against the left edge of the browser. But I don't feel strongly about it, I'm sure I'll get used to it.

static/js/code-highlighter.js
```diff
@@ -9,9 +9,146 @@
...
+  }
+
+  let newlyStuckElements = walkLinesAndGetLines();
+
+  let noLongerStuck = new Set(previouslyStuckElements);
+  let lastElem = null;
+  for (let elem of newlyStuckElements) {
+    elem.classList.add('stuck');
+    noLongerStuck.delete(elem);
+    lastElem = elem;
```
staktrace: I'd prefer to just assign `lastElem` outside the loop using a direct index the same way you do `prevLastElem`. Usually doing it inside the loop means it's there for a particular reason, like if you early-exit the loop, but that's not the case here and it confused me.

static/css/mozsearch.css
```diff
@@ -119,37 +119,48 @@ td.code {
...
-td pre code {
-    padding: 0 0 0 1em;
-    margin: 0;
-    padding: 0;
-    display: block;
+.source-line-with-number.stuck {
+  background: #ddf;
+}
+.source-line-with-number.last-stuck {
+  background: linear-gradient(#ddf, #ddf 90%, blue);
```
staktrace: In the case where a line is both selected and the last-stuck, the blue line (which I believe comes from this background) seems to be hidden by the yellow of the selection. Is there a way around that? I played around with the interaction between selection and sticky and other than this minor issue it behaves perfectly. Nice work!

tools/src/format.rs
asutherland: Yes, I meant s/both/bother/ here.

static/css/mozsearch.css
asutherland: I think it makes sense to put the padding back by default.  I'll augment the content node (`<div id="content class="content" ...>` so that it also provides a class indicating what the type of the content is.  So the `.content` can go back to the way it was and `.content.source-listing` can opt for decreased padding so we can get the "full bleed" edge-to-edge look for the sticky headers.

Another possibility would be to just use negative margins for the source listing case, but that seems like adding extra conceptually complexity to what's otherwise a reasonably straightforward table-ish thing.

static/js/code-highlighter.js
asutherland: (I was being a lazy typer.  Oh, my kingdom for a language with [-1] slicing without object creation (a la .slice(-1)[0]! ;)) Fixed.

static/css/mozsearch.css
asutherland: Yeah, if we add a selector like:
```css
.last-stuck .multihighlight {
	background: linear-gradient(rgb(255, 255, 204), rgb(255, 255, 204) 90%, blue) !important;
}
```

Then we get the yellow with a blue line at the bottom of the .last-stuck thing.  I also need to do something similar for the blame blocks which I think look a little weird without the blue line.  My plan for now is to introduce CSS variables for the highlight color, the sticky background color, and the last-stuck line at the bottom.

I'll also add comments to the CSS explaining the use of the gradient as a way to display a border line without affecting the layout box of the source lines, with it also being something we can potentially cross-animate without any layout impact.

emilio: Totally drive-by, but outline or inset-but-solid box shadows are usually a better fit for that (depending on whether you want the fake border outside or inline the box).

asutherland: Great suggestion, thank you!  I'll convert to that.

emilio: Actually, outline can also be inside the box with negative outline offset, like: `data:text/html,<div style="width: 100px; height: 100px; outline: 10px solid red; outline-offset: -10px"></div>`, and it is probably more efficient and easier to reason about.

asutherland: @emilio Hm, so if I use:
```css
.source-line-with-number.last-stuck,
.last-stuck .highlighted,
.last-stuck .multihighlight,
.last-stuck div .blame-strip {
  box-shadow: inset 0 -1px 0 0 var(--sticky-header-bottom-border-color);
}
```

Then I get the eternal anti-aliased line drawing problem that sometimes the line is 1 device pixel thick and sometimes it's 2 pixels thick.  This was with my "layout.css.devPixelsPerPx" at 1.25 and after resetting it back to "-1.0" (and squinting a little :)

asutherland: I had originally considered outline but was concerned that it's on all sides and general concern about accessibility/focus rings.  I did just give it a quick try and was surprised to realize that using a normal 1px outline somewhat does the right thing simply because the top side gets clipped at the top of the first .stuck line and on all the sides, and then the background colors of the neighboring .stuck elements hide things.

I'm not quite sure what you're proposing with your example where the outline is 10px wide though.

asutherland: Sorry for all the churn here.  I hadn't restarted after resetting my devPixelsPerPx and it seems that I properly get a single pixel line from an inset box-shadow once I do.  However, I either need to go back to 1.25 or I need to change my default font or zoom size or something.

asutherland: I've reverted to using the gradient approach.  It works with apparent visual consistency across all of my machines I tested on:
- Linux with gnome-shell under X11 with fonts scaling factor of 1.25
  - Nightly with layout.css.devPixelsPerPx=1.25
  - Nightly with layout.css.devPixelsPerPx=-1.0
- OS X 15" 2018 Macbook with default settings:
  - Nightly with layout.css.devPixelsPerPx=-1.0
- Windows 10 with 100% scaling:
  - Release with layout.css.devPixelsPerPx=-1.0

I did try using a value like -1.6px instead of 1.6px since that would equate to the 90% gradient transition point I chose, but that still resulted in oscillating line widths.

In any event, I do think we can do better than the gradient hack, I just think we should maybe defer that to a follow-up.  With any luck someone with better design sensibility and/or better understanding of Firefox graphics rendering and how devPixelsPerPx will step in :wink:

asutherland:
Since the review changes were relatively small in scope and my local testing of the changes has gone well, I'm going to land this now with the expectation that there will probably be a follow-up on the styling mechanism.  Also, there are definitely some further a11y improvements to be made as it relates to the blame popup.

PR #225, opened 2019-06-12
Bug 1558942 - Stop using generated content for line numbers in favor of user-select: none/-moz-user-select:none. r=kats

asutherland:
I've tested with NVDA on Windows Firefox release and copy-and-paste worked as expected as did NVDA.  (Although Firefox locked up for a long time producing the accessibility data on just MozsearchIndexer.cpp... :frowning_face: )

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -153,15 +153,14 @@ td#line-numbers {
   display: flex;
 }
 
-/* In order to display the line numbers and have them visible to screen readers
-   but not appear in any copy-and-pasted selection, we create a before
-   pseudo-element with generated content. */
-.line-number::before {
-    content: attr(data-line-number);
-}
 .line-number {
     display: inline-block;
     cursor: pointer;
+    -moz-user-select: none;
+    -o-user-select: none;
+    -khtml-user-select: none;
+    -webkit-user-select: none;
+    -ms-user-select: none;
     user-select: none;
     text-align: right;
     padding: 0 0.5rem;
--- a/static/js/code-highlighter.js
+++ b/static/js/code-highlighter.js
@@ -59,7 +59,7 @@ $("#scrolling").on('scroll', function() {
       return null;
     }
 
-    let num = parseInt(elem.dataset.lineNumber, 10);
+    let num = parseInt(elem.id.slice(1), 10);
     if (isNaN(num) || num < 0) {
       return null;
     }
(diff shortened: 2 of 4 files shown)


static/css/mozsearch.css
```diff
@@ -153,15 +153,14 @@ td#line-numbers {
...
-    content: attr(data-line-number);
-}
 .line-number {
     display: inline-block;
     cursor: pointer;
+    -moz-user-select: none;
+    -o-user-select: none;
+    -khtml-user-select: none;
+    -webkit-user-select: none;
+    -ms-user-select: none;
```
emilio: Probably not worth adding the -o- and khtml prefixed but fine... Also maybe add the unprefixed version?

static/js/code-highlighter.js
```diff
@@ -59,7 +59,7 @@ $("#scrolling").on('scroll', function() {
       return null;
     }
 
-    let num = parseInt(elem.dataset.lineNumber, 10);
+    let num = parseInt(elem.id.slice(1), 10);
```
staktrace: Wood it be better to use `elem.textContent` here? In case we end up changing the id later to just be the line number. (Ditto next hunk)

static/css/mozsearch.css
staktrace: +1 to adding the unprefixed version

asutherland: The unprefixed version is already there.  (All of these were actually here before.  I just experienced a bout of extreme optimism and removed all the prefixed versions.)

asutherland: (And my original a11y prototyping didn't actually use our CSS, I was working with github before...)

static/js/code-highlighter.js
asutherland: Hm, yeah, that's a good idea.  I initially avoided doing that because the `scrollIntoView` hack creates a div as a child of this element and although it doesn't perturb textContent (it's an empty div), it would be bad if it did.  But in point-of-fact, we would want textContent to change and break because it's the textContent that gets surfaced to screen readers.

asutherland: Note that there are a ton of locations in the code right now that use the id, but that happens via the `lineFromId` helper in the first ready block at the bottom of the file.  So those will need to be changed, but the helper makes it obvious what needs to change.

asutherland:
I had an hg/git muscle memory mixup and so I hadn't landed the textContent changes, those just landed in https://github.com/mozsearch/mozsearch/pull/227.  (I've gotten used to the output-file explosions due to blame line count mismatches when I forget to commit my changes protecting me from such mix-ups, but in this case no line counts were changed.)

PR #226, opened 2019-06-12
Bug 1558970 - Ensure that the number before a long line is not shrunk

Rob--W:
Test case: https://searchfox.org/mozilla-central/rev/227f5329f75bd8b16c6b146a7414598a420260cb/toolkit/locales/en-US/toolkit/about/aboutAddons.ftl#141

Before patch: Line is shifted to the left.
After patch: Line is at its normal place. The line number is happily reunited with the column of other line numbers.

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -168,6 +168,7 @@ td#line-numbers {
     /* this was originally set on the ".file td:first-child" */
     color: #aaa;
     width: 8ex;
+    flex-shrink: 0;
 }
 .highlighted,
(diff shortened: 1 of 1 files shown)


Rob--W:
Is there anything that automatically closes the BMO issue? Or should I wait for this to be in production and then manually close the bug?

asutherland:
We do it manually right now when it reaches production (so as to avoid dupes).  I know we had a web hook or something at some point for Boot2Gecko... if you know of any way to get such automation, please do file a BMO bug!

PR #228, opened 2019-06-13
Bug 1558616 - Fix scrolling logic. r=kats

asutherland:
- Restores the .line-number containing block necessary for scrollIntoView to
  work.
- Documents the position: relative/absolute instances in the CSS that I
  currently understand enough to document, possibly because I recently
  regressed them.
- Improve the documentation on scrollIntoView.
- Update the .goto padding to be based on our current position: sticky sizing.
  The math works out to -156 so this is actually a reduction right now.
- Remove the redundant manual JS scrolling that happens on startup.

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -14,6 +14,10 @@
   --highlighted-bgcolor: rgb(255, 255, 204);
 
   /* ### Sticky Headers ### */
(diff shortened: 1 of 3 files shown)


asutherland:
I have triggered a "dev" channel build of this so ideally you can sanity check that this works in the morning before the normal searchfox runs happen.

PR #230, opened 2019-07-02
Bug 1562879 - Sanitize away newlines from rust symbols

staktrace:


staktrace:
Looks like @asuth has a theory at https://bugzilla.mozilla.org/show_bug.cgi?id=1562879#c10 - I'm not sure how easy (or useful) it would be add a testcase for this given his explanation. For now I'll merge the fix but we can keep the bug open for further investigation.

asutherland:
I'll try and add a test case for at least what I suspect the problem is.  Many thanks for your excellent detective work in tracking down what file had gone awry.

That's currently probably a few days out though; I took some PTO and did some work travel after the all-hands and I'm very badly backlogged.

PR #232, opened 2019-08-05
Bug 1571295 - update getops to fix mutable borrow breakage. r=emilio

asutherland:
The fix actually showed up in getopts 0.2.17 but it seems equally as reasonable to take 0.2.19.  Feel free to push a change that just does 0.2.17 if needed, etc.


asutherland:
Please hit merge, etc. if it looks good, I'm not entirely sure I'll be able to hit the button in time.

PR #233, opened 2019-08-16
Bug 1574482 - Bad tokenization for preproc directives containing tabs

calixteman:


tools/src/tokenize.rs
```diff
@@ -291,7 +291,7 @@ pub fn tokenize_c_like(string: &str, spec: &LanguageSpec) -> Vec<Token> {
             tokens.push(Token {start: start, end: nl, kind: TokenKind::Comment});
             tokens.push(Token {start: nl, end: peek_pos(), kind: TokenKind::Newline});
         } else if ch == '#' && spec.c_preprocessor {
-            while peek_char() == ' ' {
+            while peek_char() == ' ' || peek_char() == '\t' {
```
emilio: Should these also handle other kinds of whitespace?

calixteman: Good question.
According to this line: https://github.com/mozsearch/mozsearch/blob/master/tools/src/tokenize.rs#L27
whitespace could be a "\r" but could we really have such a char after a "#" ?

emilio: Probably. Anyhow this can probably be a follow-up. Let's get this merged.

emilio:
Thanks for the patch!

PR #234, opened 2019-08-28
Bug 1577060 - Support ClassField nodes. r=kats,emilio

asutherland:
Bug 1499448 added support for the new ES class field syntax.  It also included
some provisional support for private fields which were largely disabled in
bug 1559269.

See https://github.com/tc39/proposal-class-fields for info on the spec.

This patch:
- Corrects ClassMethod processing to also generate "class#method" in addition to
  just "#method".  While figuring out how to implement ClassField I realized
  that we already did this for ObjectExpression/ObjectPattern, but not
  ClassMethod.  Our context would be right, but we'd be missing cross-reference
  symbols for the qualified name.
- Makes us process ClassField nodes similarly to how we process ObjectPattern.
  We emit an unqualified and qualified name (similar to the corrected
  ClassMethod), and we process the "init" (compare with prop's "value") with the
  name of the field pushed onto the contextStack.

### DIFF
--- a/scripts/js-analyze.js
+++ b/scripts/js-analyze.js
@@ -182,6 +182,8 @@ let Analyzer = {
    * `symbolTableStack`, `nameForThis`, or `className` are modified, the name
    * (possibly falsey) that is being used for the the thing is pushed.  When
    * traversing an ObjectExpression or ObjectPattern, the key is also pushed.
+   * (Object "dictionaries" like `{ a: { b: 1 } }` create a name hierarchy for
+   * "a.b" but do not create lexical scopes on their own.)
    */
   contextStack: [],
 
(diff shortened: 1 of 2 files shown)


asutherland:
I just triggered a "dev" indexer run with just this patch in effect against the normal full config.json, so that should turn up on "dev" soon.  (I'll move my weird UI stuff to "asuth.searchfox.org"; I was trying to juggle try review runs previously.)

If no changes are needed, it's probably most expedient for the reviewer to land.  I'll re-trigger if we're still a-ways off from the next day's run and if the reviewer doesn't say they re-triggered themselves.

asutherland:
Thanks!  The dev push ran to completion and a brief smoketest suggested all is well so I've triggered both indexing jobs.

PR #235, opened 2019-09-04
Bug 1567724 - Make indexing faster through use of scratch dir and upgrading to C5 instances.

asutherland:


infrastructure/aws/index.sh
```diff
@@ -24,16 +24,36 @@ CONFIG_INPUT="$6"
...
 export AWS_ROOT=$(realpath $MOZSEARCH_PATH/infrastructure/aws)
 VOLUME_ID=$(python $AWS_ROOT/attach-index-volume.py $CHANNEL $EC2_INSTANCE_ID)
 
+# The EBS volume will no longer be mounted at /dev/xvdf but instead at an
+# arbitrarily assigned nvme id.  However, since we only have a single EBS volume
+# and we dynamically attach it, we're pretty certain what the ID will be:
+EBS_NVME_DEV=nvme2n1
+
 set +o pipefail   # The grep command below can return nonzero, so temporarily allow pipefail
 while true
```
staktrace: We should change this `while true` to a bounded loop, `for ((i = 0; i < 3600; i++))` (or similar) so that if the device is called something else we won't spin forever.

infrastructure/aws/web-serve.sh
```diff
@@ -18,10 +18,15 @@ MOZSEARCH_PATH=$(dirname "$SCRIPT_PATH")/..
 CONFIG_REPO=$(readlink -f $1)
 CONFIG_INPUT="$2"
 
+# The EBS volume will no longer be mounted at /dev/xvdf but instead at an
+# arbitrarily assigned nvme id.  However, since we only have a single EBS volume
+# and we dynamically attach it, we're pretty certain what the ID will be:
+EBS_NVME_DEV=nvme1n1
+
 set +o pipefail   # The grep command below can return nonzero, so temporarily allow pipefail
 while true
```
staktrace: Ditto infinite loop

infrastructure/aws/web-serve.sh
```diff
@@ -18,10 +18,15 @@ MOZSEARCH_PATH=$(dirname "$SCRIPT_PATH")/..
 CONFIG_REPO=$(readlink -f $1)
 CONFIG_INPUT="$2"
 
+# The EBS volume will no longer be mounted at /dev/xvdf but instead at an
+# arbitrarily assigned nvme id.  However, since we only have a single EBS volume
+# and we dynamically attach it, we're pretty certain what the ID will be:
+EBS_NVME_DEV=nvme1n1
```
staktrace: Not clear to me why this isn't `nvme2n1` - won't 1n1 be the local storage on the web server instance? Or does the web instance not get that?

asutherland: t3.large doesn't have instance storage.  The new naming scheme appears to be that machines with instance storage have a "d" suffix.  So "m5.large" has no disk, "m5d.large" does have a disk, etc.  Only our "c5d.2xlarge" indexer and "c5d.2xlarge" AMI provisioners have disks.

staktrace: Ok, makes sense then.

infrastructure/aws/index.sh
asutherland: I've made this change and pushed to "dev" (plus the speculative change at https://bugzilla.mozilla.org/show_bug.cgi?id=1567724#c16).

infrastructure/aws/web-serve.sh
asutherland: ditto fixed

asutherland:
As discussed on https://bugzilla.mozilla.org/show_bug.cgi?id=1567724#c16 I've made the changes requested and pushed to "dev" again plus the read-json change for ordering purposes to see if the indexer is just getting tired as the indexing process goes on.  I'll post there when that run completes and with the results.

PR #239, opened 2019-09-09
Bug 1579882 - Add a very basic scp command to copy from the server to the local machine. r=kats

asutherland:
I now see I also left some comments out-of-date.  I'm fine to iterate on this, I just want to avoid letting this bit-rot or me duplicating this command from my laptop.


asutherland:
I added a bunch of comments and realized I had left in a call to list the instances that was moot, so I removed that and added some comments about that at the top.

PR #241, opened 2019-09-10
Bug 1558589 - better nginx caching directives. r=kats

asutherland:
nginx is able to provide us with etags and last-modified times for our static
source files which allows for efficient revalidation, but everything else we
serve is dynamic and currently isn't etag aware.  Because of this, we need
some time-based expiration for at least dynamic stuff.  I've arbitrarily
chosen 2 minutes, but this is motivated primarily by the development cycle.

That said, it's unlikely this will pose a major problem.

### DIFF
--- a/scripts/nginx-setup.py
+++ b/scripts/nginx-setup.py
@@ -55,8 +55,8 @@ def location(route, directives):
 
 print '''# we are in the "http" context here.
 map $status $expires {
-  default 1d;
-  "301" 10m;
+  default 2m;
+  "301" 1m;
 }
 
 server {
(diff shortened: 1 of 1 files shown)


asutherland:
So, uh, we appear to have something like that already magically: https://clicky.visophyte.org/files/screenshots/20190911-132030.png

I suspect there's more things we could do integration-wise.  CloudWatch seems to have a super neat log parsing/visualization tool under "Logs > Insights"

asutherland:
Which is to say, I think we can get historical data for any instances that are still currently alive, and since we keep the web-servers around for an extra successful indexing job, we should be able to compare A/B-wise between today's run and tomorrow's run without doing any extra changes to our setup like archiving the web-server logs, etc.

PR #244, opened 2019-09-27
Bug 1584119 - Add link to code coverage frontend in 'Other Tools' panel

marco-c:


tools/src/bin/output-file.rs
```diff
@@ -266,6 +266,14 @@ fn main() {
                 accel_key: None,
             });
         }
+        if let Some(ref ccov_root) = tree_config.paths.ccov_root {
+            tools_items.push(PanelItem {
+                title: "Code Coverage".to_owned(),
+                link: format!("{}#revision=latest&path={}", ccov_root, path),
```
emilio: Should this be revision-aware rather than assuming `latest`?

marco-c: We could do that, but right now the Mercurial revision is not available here (we only have the git one).
We could use the mapper, but I think we can do that in a follow-up (adding support for HG Web and DXR too, which are both pointing to "tip").

emilio: Sure, sounds good.

marco-c: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1585266.

PR #248, opened 2019-10-04
[wip] Move to Ubuntu 18.04 LTS.

emilio:
And also allow libvirt as a back end, so I don't have to deal with incompatible
VirtualBox drivers :(

I _think_ the main issue here is that the disk space goes down to the regular
one (40GB), which may be not enough for mozilla-index...

Testing that with the indexing atm. But test repo and provisioning works fine.

Also I guess we'd need to update the AMIs and such...

I got into this because I wanted to hack on Searchfox and Virtualbox is 
unusable in my current kernel...

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,15 +1,22 @@
 Vagrant.configure("2") do |config|
-  # We use this image to get a large (160GB) disk rather than the normal 40GB.
-  config.vm.box = "cbednarski/ubuntu-1604-large"
-  config.vm.box_version = "0.1.0"
+  config.vm.box = "generic/ubuntu1804"
+  config.vm.box_version = "1.9.34"
 
   config.vm.provision :shell, privileged: false, path: "infrastructure/indexer-provision.sh"
   config.vm.provision :shell, privileged: false, path: "infrastructure/vagrant/indexer-provision.sh"
   config.vm.provision :shell, privileged: false, path: "infrastructure/web-server-provision.sh"
 
-  config.vm.network :forwarded_port, guest: 80, host: 8000
+  config.vm.network :forwarded_port, guest: 80, host: 8001
 
   config.vm.provider "virtualbox" do |v|
+    v.memory = 10000
+    v.cpus = 4
+  end
+
+  config.vm.provider "libvirt" do |v|
+    # Need to do this manually for libvirt...
+    config.vm.synced_folder './', '/vagrant', type: 'nfs', nfs_udp: false, accessmode: "squash"
+
     v.memory = 10000
     v.cpus = 4
   end
--- a/infrastructure/indexer-provision.sh
+++ b/infrastructure/indexer-provision.sh
@@ -37,15 +37,15 @@ popd
 
 # Clang
 wget -O - http://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
-sudo apt-add-repository "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-${CLANG_VERSION} main"
+sudo apt-add-repository "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-${CLANG_VERSION} main"
 sudo apt-get update
 sudo apt-get install -y clang-${CLANG_VERSION} libclang-${CLANG_VERSION}-dev
 
-# Firefox: https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Build_Instructions/Linux_Prerequisites
-sudo apt-get install -y zip unzip mercurial g++ make autoconf2.13 yasm libgtk2.0-dev libgtk-3-dev libglib2.0-dev libdbus-1-dev libdbus-glib-1-dev libasound2-dev libcurl4-openssl-dev libiw-dev libxt-dev mesa-common-dev libgstreamer0.10-dev libgstreamer-plugins-base0.10-dev libpulse-dev m4 flex libx11-xcb-dev ccache libgconf2-dev
-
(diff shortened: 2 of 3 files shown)


emilio:
@staktrace / @asutherland is this something desirable?

I guess we still have a year plus of support for 16.04 per https://ubuntu.com/about/release-cycle... So maybe not urgent. But this allowed me to work on searchfox locally, and get rid of VirtualBox :)

Vagrantfile
```diff
@@ -1,16 +1,23 @@
...
-  config.vm.box_version = "0.1.0"
+  config.vm.box = "generic/ubuntu1804"
+  config.vm.box_version = "1.9.34"
 
   config.vm.provision :shell, privileged: false, path: "infrastructure/indexer-provision.sh"
   config.vm.provision :shell, privileged: false, path: "infrastructure/vagrant/indexer-provision.sh"
   config.vm.provision :shell, privileged: false, path: "infrastructure/web-server-provision.sh"
 
-  config.vm.network :forwarded_port, guest: 80, host: 8000
+  config.vm.network :forwarded_port, guest: 80, host: 8001
```
emilio: The port was because it conflicts with WPT and was doing something else, can revert or update the docs.

jens1o:
Does that mean it will no longer complain about non-matching guest driver extension-thing versions? That'd be great!<

emilio:
> Does that mean it will no longer complain about non-matching guest driver extension-thing versions? That'd be great!

Yes, you need to run `vagrant up --provider libvirt`, but yeah, no more virtualbox stuff :)

infrastructure/indexer-provision.sh
```diff
@@ -37,15 +37,15 @@ popd
...
 
-# Firefox: https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Build_Instructions/Linux_Prerequisites
-sudo apt-get install -y zip unzip mercurial g++ make autoconf2.13 yasm libgtk2.0-dev libgtk-3-dev libglib2.0-dev libdbus-1-dev libdbus-glib-1-dev libasound2-dev libcurl4-openssl-dev libiw-dev libxt-dev mesa-common-dev libgstreamer0.10-dev libgstreamer-plugins-base0.10-dev libpulse-dev m4 flex libx11-xcb-dev ccache libgconf2-dev
-
 # Other
-sudo apt-get install -y parallel realpath python-virtualenv python-pip
+sudo apt-get install -y parallel python-virtualenv python-pip
+
+# Firefox: https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Build_Instructions/Linux_Prerequisites
+wget -O bootstrap.py https://hg.mozilla.org/mozilla-central/raw-file/default/python/mozboot/bin/bootstrap.py && python bootstrap.py --application-choice=browser --no-interactive && rm bootstrap.py
```
staktrace: Does this actually work? Last time I tried it, `--no-interactive` didn't actually install all the things needed, or maybe it was still interactive, I don't remember which.

Vagrantfile
asutherland: +1 to this.  I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1491956 on this previously (but didn't push a fix).

asutherland:
> So in general I'm fine with updating to Ubuntu 18.04, but we'll need to make sure the changes to provisioning can still build AMIs properly and everything works in production. And of course a bunch of documentation will need to be updated as well for the VirtualBox/libvirt change.

To elaborate on this:
- `trigger_provision.py` wants to have the base AMI updated to be the appropriate us-west-2 bionic amd64 "hvms:ebs-ssd" AMI.  Browsing to https://cloud-images.ubuntu.com/locator/ec2/ and searching gets us the 20191002 image of https://cloud-images.ubuntu.com/locator/ec2/.  So the hunk that is currently `trigger_provision.py`
```shell
# ubuntu/images/hvm-ssd/ubuntu-xenial-16.04-amd64-server-20190816 - ami-0135f076a31aebe66
image_id = 'ami-0135f076a31aebe66'
```
  instead wants to be
```shell
# ubuntu/images/hvm-ssd/ubuntu-bionic-18.04-amd64-server-20191002 - ami-06d51e91cea0dac8d
image_id = 'ami-06d51e91cea0dac8d'
```
- The instructions at https://github.com/mozsearch/mozsearch/blob/master/docs/aws.md#provisioning-and-cloud-init are very good but note that I moved us to use "indexer-16.04-ena" (not "indexer-16.04") and "web-server-16.04-ena" (not "web-server-16.04") when I updated our AMIs to get us fancy new EC2 features.  Presumably the new images would want to just be named "indexer-18.04" and "web-server-18.04" and then we can just delete the "16.04-ena" ones once we have switched over.  (The new 18.04 images will automatically have the "ena" stuff we need but we don't need it in the name.)
  - `trigger-web-server.py` and `trigger_indexer.py` will need to be updated with the new AMI names.  (And I guess the docs too.)
  - You should be able to generate the AMIs from your branch since trigger_provision.py just uploads the scripts locally from your branch.

infrastructure/indexer-provision.sh
emilio: It seems to work now.

asutherland:
I've filed a bug for this at https://bugzilla.mozilla.org/show_bug.cgi?id=1601451 and am planning to steal the patch and run with it.  (I upgraded to ubuntu to 19.10 some time ago and would rather spend the time on this than trying to get virtualbox happy.)

asutherland:
Closing this since the changes were subsumed into #258.

PR #249, opened 2019-10-05
Make some scripts work when ran from the VM in a git worktree.

emilio:
When you're in a worktree, the `.git` thing that gets to `/vagrant` is not the
repo itself but a reference like:

> gitdir: /home/emilio/src/moz/mozsearch/.git/worktrees/mozsearch-ubuntu-up

Which is obviously _not_ in the /vagrant directory, and thus git operations fail
from inside the virtual machine.

### DIFF
--- a/infrastructure/aws/build-lambda-indexer-start.sh
+++ b/infrastructure/aws/build-lambda-indexer-start.sh
@@ -18,7 +18,7 @@ CONFIG_INPUT=$3
 BRANCH=$4
 CHANNEL=$5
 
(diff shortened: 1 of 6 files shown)


emilio:
Right, you cannot access it, and this is the only thing that actually accesses it, so by changing it to not do it... :)

emilio:
(To clarify, the files in the worktree are accessible to the VM, it's the git metadata what is not)

staktrace:
Oh I see, you don't need git commands to work, just need the dir. Makes sense.

PR #251, opened 2019-10-16
Bug 1583031 - Have rust analysis run (optionally) per-platform

asutherland:
This patch combined with the mozsearch-mozilla patch causes us to process the rust save-analysis data into the per-platform analysis directories, and then merge the results.

With these changes, I got a config.json run to seem to complete happily where:
- https://asuth.searchfox.org/mozilla-central/source/servo/components/style/values/specified/text.rs is no longer broken.  (Note that I've just triggered a mozilla-releases.json indexer run, so while the webserver for mozilla-central will stay up, the load balancer config will end up shifting away and the URL will become a 503 or whatever.  Feel free to inspect via ssh or reconfigure the load balancers if desired.)
- We end up with an analysis file in the expected place that looks right.
- Because our scripts still delete the per-platform analysis directories, it's a little harder to verify that those look right.

Note that I am still planning to improve the formatting logic in a separate patch so that the HTML can never get broken again, but I was trying hard to make sure I wasn't ignoring other correctness issues, which indeed is what I found with us processing all platforms together.  (This patch should also speed us up because we will run all 4 platforms in parallel.)

### DIFF
--- a/scripts/rust-analyze.sh
+++ b/scripts/rust-analyze.sh
@@ -4,27 +4,54 @@ set -x # Show commands
 set -eu # Errors/undefined vars are fatal
 set -o pipefail # Check all commands in a pipeline
 
-if [ $# -ne 2 ]
+if [ $# -lt 2 ]
 then
-    echo "Usage: rust-analyze.sh config-file.json tree_name"
+    echo "Usage: rust-analyze.sh config-file.json tree_name [platform-filter]"
     exit 1
 fi
 
 CONFIG_FILE=$(realpath $1)
 TREE_NAME=$2
 
+# Our current rust-indexer implementation can get confused if it's given the
+# same crate but for different platforms at the same time.  So we provide an
+# optional mechanism that takes a platform name per mozsearch-mozilla
+# convention and:
+# - filters the `find` invocations to only pick up output for that platform.
+# - puts the output in `analysis-linux64` for example instead of just
+#   `analysis`.
+PLATFORM=${3:-}
+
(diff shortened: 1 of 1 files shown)


scripts/rust-analyze.sh
```diff
@@ -4,27 +4,54 @@ set -x # Show commands
...
+# Hacky mechanism where we take a rust objdir value like x86_64-unknown-linux-gnu
+# to filter the rust analysis to just that.
+PLATFORM_SAVE_FILTER=${PLATFORM:+${RUST_PLAT_DIRS[$PLATFORM]}/debug/deps/}
+PLATFORM_FILTER=${PLATFORM:+${RUST_PLAT_DIRS[$PLATFORM]}/}
+PLATFORM_SUFFIX=${PLATFORM:+-$PLATFORM}
+
 if [ -d "$OBJDIR" ]; then
-  ANALYSIS_DIRS="$(find $OBJDIR -type d -name save-analysis)"
+  # Bail if the build step already performed rust analysis.
+  if [ -f "$OBJDIR/rust-analyzed" ]; then
```
staktrace: This check/bailout seems like it would be better placed around the invocation of this script in mkindex.sh.

asutherland: I am going to temporarily take the punt because my fluency with bash/test conditionals is such that I wouldn't trust the change without running a new try run, especially since it requires effectively inverting the check.  I'll plan to try and make the change and include it in a future try run for future review.

PR #252, opened 2019-11-02
Bug 1593372 - Add proxy_cache and proxy_cache_path settings to searchfox nginx to cache dynamic content.

asutherland:
…


asutherland:
Thanks for the review!  I'll keep an eye on tomorrow's build and iterate if any hiccups appear.

PR #253, opened 2019-11-12
Bug 1595492 - switch over to new taskcluster CI.

asutherland:
Since trunk is broken, I'm going to speculatively land this and the other mozsearch-mozilla fix and then attempt try runs directly from lambda.  I'll ask for after-the-fact review on this and any additional fix-ups I need to perform.


asutherland:
This worked.  Please let me know if you'd like any fix-ups!

PR #256, opened 2019-11-26
Bug 1593833 - rust third_party indexing support.

asutherland:


asutherland:
The approach evolved after feedback in https://github.com/mozsearch/mozsearch-mozilla/pull/65 and those changes landed in https://github.com/mozsearch/mozsearch/pull/299.

PR #257, opened 2019-12-02
Fix code coverage links when lines are selected. 

emilio:


static/js/code-highlighter.js
```diff
@@ -273,10 +273,12 @@ $(function () {
...
       windowHash = windowHash.replace(reCleanup, '');
       history.replaceState(null, '', windowHash);
-      scrollIntoView(windowHash.slice(1), false);
+      let hash = windowHash.slice(1);
+      scrollIntoView(hash, false);
 
-      $("a[data-update-link=true]").each(function(i, elt) {
-        $(elt).attr("href", $(elt).attr("data-link") + windowHash);
+      $("a[data-update-link]").each(function(i, elt) {
+        let extra = elt.attr("data-update-link").replace("{}", hash);
```
staktrace: My jquery-fu is terrible. Does this need to be `$(elt)` instead of `elt`?

static/js/code-highlighter.js
```diff
@@ -273,10 +273,12 @@ $(function () {
     if (windowHash) {
       windowHash = windowHash.replace(reCleanup, '');
       history.replaceState(null, '', windowHash);
-      scrollIntoView(windowHash.slice(1), false);
+      let hash = windowHash.slice(1);
```
staktrace: Maybe call this variable `lineno` so it's more self-documenting

static/js/code-highlighter.js
asutherland: It looks like it does need to be `$(elt)` per https://api.jquery.com/each/ which identifies the second arg as an [Element](https://api.jquery.com/Types/#Element) which is the raw DOM type and [jQuery](https://api.jquery.com/Types/#jQuery) would be the magic type you get from the `$()` wrapping.

emilio: Yeah, whoops, I mistested this.

emilio:
We should probably get rid of jquery, in general, wdyt?

staktrace:
I'd be happy to get rid of it in general.

asutherland:
Yeah, I think given that the searchfox developer-base is all web browser engineers, jQuery just adds confusion.

PR #258, opened 2019-12-05
Bug 1601451 - Move to Ubuntu 18.04 LTS.

asutherland:
This is Emilio's WIP commit (the commit summary slightly changed) from https://github.com/mozsearch/mozsearch/pull/248 plus an additional commit with:
- Updated documentation.  I probably missed some things, but I think I got at least some of the obvious ones!
- Enhancements to the NVME logic.  My hardcoded paths broke immediately.  Thankfully on 18.04 the nvme commands can dump JSON output and `jq` can let us filter it to get the information we need.  We no longer need to do `lsblk` because it's basically the same as the `nvme` calls.
  - We are able to match based on volume id and just the type of volume (instance disk versus EBS).
  - In theory we could also match by the pre-NVME device/mountpoint where the EBS attach would have happened from the "vs" vendor extended info data, but I couldn't get that dumped as part of any JSON output and then we're basically looking at extracting from an `od` style representation embedded in a larger textual representation (and my sed and awk skills are bad, so bad).
- Some package install changes:
  - Add debug info for python so we can get python stacktraces if router.py freaks out again
  - Disabled unattended upgrades so that our debug info is actually up-to-date
  - Added jq for JSON querying
- Some provisioning changes:
  - An extra AWS specific provisioning script for the web server so it can also get the nvme packages installed.
  - I force an unattended upgrade at provisioning time.  The ubuntu base image we were starting from already had like 50 out-of-date packages.  This isn't entirely essential but it's nice to make sure all the packages were installed from approximately the same time.  And if we re-provision without updating the base image, this does start to become essential.  The good news is that we know the unattended part works because I did mess this up initially and in testing experienced a prompt about a grub config file change.

I do have a successful https://dev.searchfox.org/ run up right now, and it all ran using the 18.04 AMI's I provisioned.  Merging this would want to happen concurrently with updating the lambdas, so this is a case where I won't request merging before the morning run.

One disclaimer: I ran the `vagrant up` before I made some additional ubuntu changes, so I or someone else should probably give that a try again before we merge this.  Also, I haven't tried it under virtualbox... I guess I should try that on OS X maybe?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,16 +6,31 @@ Searchfox locally using Vagrant.
 
 ## Vagrant setup for local development
 
-First, [install Vagrant](https://www.vagrantup.com/downloads.html) and
-[VirtualBox](https://www.virtualbox.org/wiki/Downloads) by
-following the instructions for your OS. Then clone Mozsearch and
-provision a Vagrant instance:
+First, [install Vagrant](https://www.vagrantup.com/downloads.html).  This is a
+virtualization automation tool.  It depends on a virtualization backend.
 
+Pick a virtualization backend and install it.  Your options are:
+- libvirt via [vagrant-libvirt](https://github.com/vagrant-libvirt/vagrant-libvirt).
+  If you are on linux and aren't already using VirtualBox, it's the recommended
+  path.
+  - Follow the [installation instructions](https://github.com/vagrant-libvirt/vagrant-libvirt#installation).
+  - Note that if you didn't already have libvirt installed, then a new `libvirt`
+    group may just have been created and your existing logins won't have the
+    permissions necessary to talk to the management socket.  If you do
+    `exec su -l $USER` you can get access to your newly assigned group.
+- VirtualBox.  For platforms that aren't linux.  Visit the
+  [VirtualBox downloads page](https://www.virtualbox.org/wiki/Downloads) and
+  following the instructions for your OS.
+
+Then clone Mozsearch and provision a Vagrant instance:
 ```
 git clone https://github.com/mozsearch/mozsearch
 cd mozsearch
 git submodule update --init
-vagrant up
+# if using libvirt
+vagrant up --provider libvirt
+# if using virtualbox
+vagrant up --provider virtualbox
 ```
 
 The last step will take some time (10 or 15 minutes on a fast laptop)
@@ -101,8 +116,8 @@ requests, we can start the server as follows:
 ```
 
 At this point, you should be able to visit the server, which is
-running on port 80 inside the VM and port 8000 outside the VM. Visit
-`http://localhost:8000/` to do so.
+running on port 80 inside the VM and port 8001 outside the VM. Visit
+`http://localhost:8001/` to do so.
 
 ## Indexing Mozilla code locally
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -6,16 +6,31 @@ Searchfox locally using Vagrant.
...
+
+Pick a virtualization backend and install it.  Your options are:
+- libvirt via [vagrant-libvirt](https://github.com/vagrant-libvirt/vagrant-libvirt).
+  If you are on linux and aren't already using VirtualBox, it's the recommended
+  path.
+  - Follow the [installation instructions](https://github.com/vagrant-libvirt/vagrant-libvirt#installation).
+  - Note that if you didn't already have libvirt installed, then a new `libvirt`
+    group may just have been created and your existing logins won't have the
+    permissions necessary to talk to the management socket.  If you do
+    `exec su -l $USER` you can get access to your newly assigned group.
```
emilio: Could you add a section like this to the docs? I had this locally, but it seems I never pushed it:

> ## Troubleshooting
> 
> If vagrant up times out in the "Mounting NFS shared folders..." step, chances
> are that you cannot access nfs from the virtual machine.
> 
> You probably need to allow `libvirt` to access `nfs`:
> 
> ```
> # firewall-cmd --permanent --add-service=nfs --zone=libvirt
> # firewall-cmd --permanent --add-service=rpc-bind --zone=libvirt
> # firewall-cmd --permanent --add-service=mountd --zone=libvirt
> # firewall-cmd --reload
> ```

I had to do this for a new Fedora 31 machine.

asutherland: Okay, I added this.

infrastructure/aws/index.sh
```diff
@@ -45,16 +77,16 @@ echo "Channel is $CHANNEL"
...
+# remove the/any dash from the volume id.
+JQ_QUERY=".Devices[] | select(.SerialNumber == \"${VOLUME_ID/-/}\") | .DevicePath"
 
 set +o pipefail   # The grep command below can return nonzero, so temporarily allow pipefail
 for (( i = 0; i < 3600; i++ ))
 do
-    COUNT=$(lsblk | grep $EBS_NVME_DEV | wc -l)
-    if [ $COUNT -eq 1 ]
+    EBS_NVME_DEV=$(sudo nvme list -o json | jq --raw-output "$JQ_QUERY")
+    if [[ $EBS_NVME_DEV ]]
```
staktrace: What ends up happening here in the failure case? I would assume `EBS_NVME_DEV` is empty, so you'd want to do `-n "$EBS_NVME_DEV"` in the condition. Or does just putting the variable alone work in bash?

infrastructure/aws/web-serve.sh
```diff
@@ -18,16 +18,53 @@ MOZSEARCH_PATH=$(dirname "$SCRIPT_PATH")/..
...
+# locate the given device.  (We do need to remove any dashes, however.)
+JQ_QUERY=".Devices[] | select(.SerialNumber == \"${VOLUME_ID/-/}\") | .DevicePath"
 
 set +o pipefail   # The grep command below can return nonzero, so temporarily allow pipefail
 for (( i = 0; i < 3600; i++ ))
 do
-    COUNT=$(lsblk | grep $EBS_NVME_DEV | wc -l)
-    if [ $COUNT -eq 1 ]
+    EBS_NVME_DEV=$(sudo nvme list -o json | jq --raw-output "$JQ_QUERY")
+    if [[ $EBS_NVME_DEV ]]
```
staktrace: Ditto

PR #259, opened 2019-12-13
Bug 1603193 - Move bulk of VM update steps into extensible scripts. r=kats

asutherland:
Move from having high-churn dependency updates (rust/spidermonkey) and build
steps being permanently baked into the provisioning-generated `update.sh` and
instead have them live in the mozsearch repo.  This makes it easier for
branches to try things without needing to re-provision.

### DIFF
--- a/infrastructure/indexer-provision.sh
+++ b/infrastructure/indexer-provision.sh
@@ -140,22 +140,6 @@ echo Branch is $BRANCH
 echo Mozsearch repository is $MOZSEARCH_REPO
 echo Config repository is $CONFIG_REPO
 
-# Update Rust (make sure we have the latest version).
-# We need rust nightly to use the save-analysis, and firefox requires recent
-# versions of Rust.
-rustup update
-
-# Install SpiderMonkey.
-rm -rf jsshell-linux-x86_64.zip js
-wget -nv https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/gecko.v2.mozilla-central.latest.firefox.linux64-opt/artifacts/public/build/target.jsshell.zip
-mkdir js
-pushd js
-unzip ../target.jsshell.zip
-sudo install js /usr/local/bin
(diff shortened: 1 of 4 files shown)


asutherland:
Thanks for the review!  I'm copying the existing AMI's for now to names with the "-backup" suffix so that for testing I don't need to mess with the AMI names.  My plan is basically:
- Backup the AMI's.
- Merge this.
- Re-provision.  If the provisioning succeeds, delete the old AMI's with the canonical names and snapshot AMI's from the newly provisioned machines.  If the provisioning fails, iterate on the shell scripts on the machines until they are happy.  Then nuke the machines and re-provision.  Repeat until victory.
- Run a release config.json job from lambda. (The lambdas don't need to be re-provisioned because the AMI names aren't changing.)
- Assuming all goes well, delete the backup AMI's.

asutherland:
(There's no way to rename the AMI's that I'm aware of.  There is an option to register an AMI from an existing AMI but it wasn't totally clear what that was doing.)

asutherland:
Ran into some provisioning trouble.  We use pip to install pygit2 without specifying a version, but we use a hard-coded version of libgit2.  The current pygit2 wants a v0.28.x version of libgit.  It's also the case that the 1.0 release release of pygit2 has officially dropped support for python2.

I think in general with python 2.7 losing all support as of the new year and the known flakiness router.py has that's likely due to the socketserver stuff never getting updates on 2.x, maybe the right answer in the short-medium term is to upgrade to python3 for stuff.

I'm going to take a quick look at installing the previous 0.28.2 release of pygit2.  Assuming that's still available I'll do that as a minimal bustage fix.

asutherland:
To clarify my previous statement a little: bumping the libgit2 version on its own didn't work because the python2 backwards-compat C++ stuff was also gone, there was just no obvious error about it.  And for the bustage fix, I am planning to update us to the latest release of libgit2, v0.28.4.

PR #260, opened 2019-12-13
Bug 1603193 - pygit2 bustage fix. a=bustage

asutherland:
Running pip install with an unconstrainted pygit2 installed pygit2 1.0.0
which needed a more recent version of libgit2 and which also was incompatible
with Python2.7 (but only said the former, not the latter).

This change:
- Forces us to specifically install the last release of pygit2 prior to 1.0.0.
- Upgrades our libgit2 version to the most recent version.  I don't think this
  is strictly necessary with pygit2 at an older version, but I already had done
  the script changes necessary and things built correctly.  (And there doesn't
  seem to be an advantage to staying on an older version, especially given that
  when we do upgrade to pygit2 1.0.0 we will need an 0.28.x series release of
  libgit2 anyways.

### DIFF
--- a/infrastructure/indexer-provision.sh
+++ b/infrastructure/indexer-provision.sh
@@ -95,17 +95,22 @@ rm -rf .cache/bazel
 sudo pip install boto3
 
 # Install pygit2.
-rm -rf libgit2-0.27.1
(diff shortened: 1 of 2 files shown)


asutherland:
I'm merging this as a bustage-fix since I don't think this is likely to be a contentious change, but am request post-facto review.

PR #261, opened 2019-12-14
Bug 1603193 - Provisioning bustage fixes. a=bustage

asutherland:
I forgot to mark the scripts executable.  Whoops!
Also, a ride-along fix here to make sure that any work we had in progress in index-scratch gets moved to the actual permanent EBS store.


asutherland:
The ride-along index-scratch stuff worked.  On my dev branch I broke the indexer and I happily got: /index/interrupted/config.json  /index/interrupted/gecko-blame.tar  /index/interrupted/gecko-dev.tar  /index/interrupted/git_hg.map  /index/interrupted/mozilla-central

Disclaimer: I had to do: `lsblk` to identify the exactly 300G partition and then `sudo mount` that.  I have it on my mental to-do list to update the docs for debugging indexers, but I'm think it might be best if we can just add a startup script or something that understands how to re-attach the /index mount-point as long as the volume is still attached to the indexer.  (As opposed to it having been in the process of being moved to the web-server.)

PR #268, opened 2020-01-06
Bug 1607109 - Upgrade searchfox indexer to LLVM 9/clang 9. r=emilio

asutherland:
There isn't actually a rush to land this.  I'd probably wait until the end of the week to land and re-provision because this seems like the kind of thing that inevitably will cause breakage.  But I need it for the fancy branch, so I thought I'd prepare it.


asutherland:
Not that I'm expecting breakage.  I just am not an optimist when it comes to deploying things ;)

PR #271, opened 2020-01-28
reorder git-cinnabar `pushd` to be correct

asutherland:
Provisioning failed on the checkout because it was in the wrong dir.


asutherland:
I added an `-f` arg to the symlinking because re-running the provisioning after updating this failed on creating the already-existing symlink.  (new commit)

PR #272, opened 2020-01-29
Make NFS settings libvirt-specific

asutherland:
Although our intent when adding libvirt was to add the NFS mount only in the
case of libvirt and not for virtualbox, we did not, so it was also being
added for virtualbox on OS X which did not work.  This corrects the syntax to
only enable NFS on libvirt.

This also has a ride-along of updating to the most recent released image
version, primarily because I tried this as part of my troubleshooting before
realizing what was going on, but it seems like an advantageous change.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,6 +1,6 @@
 Vagrant.configure("2") do |config|
   config.vm.box = "generic/ubuntu1804"
-  config.vm.box_version = "1.9.34"
+  config.vm.box_version = "2.0.6"
 
   config.vm.provision :shell, privileged: false, path: "infrastructure/indexer-provision.sh"
   config.vm.provision :shell, privileged: false, path: "infrastructure/vagrant/indexer-provision.sh"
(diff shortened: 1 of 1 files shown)


asutherland:
The docs about this are at https://www.vagrantup.com/docs/providers/configuration.html#overriding-configuration

Without this change. `vagrant up` fails in the initial stages with an error of the form:
```
NFS requires a host-only network to be created.
Please add a host-only network to the machine (with either DHCP or a
static IP) for NFS to work.
```

If one ran `vagrant up` again after this, it sorta seemed like maybe the problem had been addressed because the error didn't happen again, but the problem was still there, /vagrant would not end up mounted.  In my testing of the libvirt support I'd gotten confused by this and had assumed there were just other virtualbox issues going on because of my prior bad experiences.

I'm going to push this immediately because I'm concerned that other people at the all hands might also have jet lag and be trying to do the setup, but I'd appreciate sanity-checking/review still.

Also note that I'm going to land some doc changes for provisioning made directly via the github UI as part of an attempt to make the process less painful for people.  I'll comment in the chat.mozilla.org channels when I do that so people can help provide feedback or know to directly provide suggested changes, etc.

asutherland:
Argh.  So, provisioning from scratch did now complete on OS X, but I'm not seeing a `/vagrant` directory.  It seems like this config might possibly be clobbering the default "/vagrant" mapping, perhaps?  Part of my trouble-shooting was to use `vagrant init hashicorp/bionic64` and I verified that properly mounted `/vagrant`.  I'll remove the libvirt stuff entirely and halt/up and see if that fixes it to better isolate.

PR #273, opened 2020-01-29
Improve vagrant setup documentation

asutherland:
This is the result of repeated testing of vagrant provisioning on Ubuntu 19.10 and OS X 10.14.6.

I'm setting this up as a pull request to hopefully make it easier to provide feedback on the changes, but I think it's also okay for people to just directly tweak the docs wiki-style if there's a clear improvement.  (I've done this a few times on this doc previously when the changes seemed clear-cut.  These are larger and more substantial changes that I think review would normally be appropriate for, but because of time pressure factors related to the all-hands and the time in Berlin and Toronto,  I'm just going to land this immediately.)

cc'ing @staktrace @emilio

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,43 +6,83 @@ Searchfox locally using Vagrant.
 
 ## Vagrant setup for local development
 
-First, [install Vagrant](https://www.vagrantup.com/downloads.html).  This is a
-virtualization automation tool.  It depends on a virtualization backend.
+### Setting up the VM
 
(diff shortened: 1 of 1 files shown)


asutherland:
Also, I want to call out how great it is that @staktrace added the makefile in the root of the repo.  I think I'd totally missed its existence until the other day I wanted to add something like it, and there it already was!  Setup so very nicely!  (And I'd manually been creating worse shell scripts like a sucker!)

PR #275, opened 2020-02-11
Baby steps for web-server.rs knowing its memory usage

asutherland:
This is where I got the numbers at https://bugzilla.mozilla.org/show_bug.cgi?id=1464297#c14 from.  In general, it seems useful for us to be able to understand where our runtime memory usage lives, so I think this is probably worth landing.  Note that I treat the git2 internals other than the Oids as 0 memory usage.  That's obviously not accurate.  I think jemalloc can tell us the total aggregate allocated memory, so we could certainly try and do some math to note what we explicitly can track versus the overall allocation size.  (And we can also sample the heap sizes at different points while the program is still single-threaded.)  I'm not really wedded to landing this, I just don't want to have to implement it again in the future ;)

There's some more info in the commit messages but the basic idea is that, AIUI (As I Understand It), in order to get MallocSizeOf for BTreeMap and HashMap, the impls have to live in the crate that defines the trait.  wr_malloc_size_of is a cut-down version of servo's malloc_size_of, but it's too cut down and doesn't have BTreeMap and HashMap.  So I fork servo's malloc_size_of and cut it down to what we want/etc.

I also force our rust stuff to use jemalloc.  I mainly did this to get a usable malloc_usable_size function.  Apparently at https://github.com/rust-lang/rust/issues/32838 there's nightly-only magic so that perhaps we could get this with the default system allocator, but I like jemalloc and we use it everywhere, so why not just use jemalloc, right?  (End hand-waving.)

### DIFF
--- a/tools/malloc_size_of/Cargo.toml
+++ b/tools/malloc_size_of/Cargo.toml
@@ -1,0 +1,53 @@
+[package]
+name = "malloc_size_of"
+version = "0.0.1"
+authors = ["The Servo Project Developers"]
+license = "MIT/Apache-2.0"
+publish = false
+
+[lib]
+path = "lib.rs"
+
+[features]
+servo = [
+    "crossbeam-channel",
+    "hyper",
+    "hyper_serde",
+    "keyboard-types",
+    "serde",
+    "serde_bytes",
+    "string_cache",
+    "time",
+    "url",
+    "webrender_api",
+    "xml5ever",
+    "content-security-policy",
+    "uuid",
+    "accountable-refcell",
+]
+
+[dependencies]
+accountable-refcell = { version = "0.2.0", optional = true }
+app_units = "0.7"
+content-security-policy = {version = "0.3.0", features = ["serde"], optional = true}
+crossbeam-channel = { version = "0.3", optional = true }
+cssparser = "0.27"
+euclid = "0.20"
+hashglobe = { path = "../hashglobe" }
+hyper = { version = "0.12", optional = true }
+hyper_serde = { version = "0.11", optional = true }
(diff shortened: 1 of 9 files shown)


tools/malloc_size_of/Cargo.toml
```diff
@@ -10,44 +10,30 @@ path = "lib.rs"
...
-accountable-refcell = { version = "0.2.0", optional = true }
 app_units = "0.7"
-content-security-policy = {version = "0.3.0", features = ["serde"], optional = true}
 crossbeam-channel = { version = "0.3", optional = true }
-cssparser = "0.27"
 euclid = "0.20"
-hashglobe = { path = "../hashglobe" }
 hyper = { version = "0.12", optional = true }
 hyper_serde = { version = "0.11", optional = true }
 keyboard-types = {version = "0.4.3", optional = true}
```
emilio: We should probably rip out all this stuff as well, no point in building it. Or I guess we won't build it because we won't use the `servo` feature? If so, fine.

asutherland: Yeah, more ripping out is possible.  I intentionally left in things I thought we might use (we do use hyper) or could conceivably use (string caches might be neat?), but it also looks like I left some stuff in Cargo.toml that I removed from lib.rs.  I'll do a more thorough pass shortly now that it's all working.

asutherland: Okay, I ripped out everything except for `git2` in Cargo.toml.  I'm pretty confident in my ability to copy-and-paste back anything we end up needing like the limited "hyper" support!

PR #276, opened 2020-02-17
Bug 1616104 - Move line numbers out of X selection on Chrome

Rob--W:
This is sad, but needed to work around https://bugs.chromium.org/p/chromium/issues/detail?id=850685

I limited the condition to non-Firefox for now, but if you prefer, I can invert the check, to specifically target Chrome, and maybe even specifically Linux.

### DIFF
--- a/static/js/code-highlighter.js
+++ b/static/js/code-highlighter.js
@@ -496,3 +496,17 @@ $(function () {
   });
 
 });
+
+// We use user-select:none to hide line numbers from text selections. That
+// does however not work in Chrome (https://crbug.com/850685)
+// or Safari (https://bugzilla.mozilla.org/show_bug.cgi?id=1616104#c1).
+// As a work-around, move all line numbers into pseudo-elements when the user
+// selects something for the first time.
+if (navigator.userAgent.indexOf('Firefox') == -1) {
+  document.addEventListener('selectstart', function() {
+    for (let lineno of document.querySelectorAll('.line-number')) {
+      lineno.dataset.lineNumber = lineno.textContent;
+      lineno.textContent = '';
+    }
+  }, { once: true });
+}
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -148,6 +148,15 @@ td#line-numbers {
   display: flex;
 }
 
+/* Line numbers are text nodes before the actual source lines,
+   and styled with "user-select:none" to hide them from text selections.
+   This does not work in Chrome ( https://crbug.com/850685 ), so we work around
+   the issue by moving the content to an attribute and using generated content.
+ */
+.line-number[data-line-number]::before {
+    content: attr(data-line-number);
+}
+
 .line-number {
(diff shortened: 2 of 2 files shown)


static/js/code-highlighter.js
```diff
@@ -496,3 +496,20 @@ $(function () {
...
+
+$(function() {
+  // We use user-select:none to hide line numbers from text selections. That
+  // does however not work in Chrome - https://crbug.com/850685
+  // As a work-around, we will use generated content to render the line numbers.
+  if (navigator.userAgent.indexOf("Firefox") != -1) {
+    // Work-around is only needed in Chrome.
+    return;
+  }
+  $('#file').one('selectstart', function() {
```
asutherland: This seems like a typo that accidentally worked.  I understand `one` to be akin to passing `once: true` to addEventListener, but I presume this would want to happen every time.

Note: I'm not necessarily endorsing this approach or even fixing this bug, just doing a drive-by.

Rob--W: Not a typo. https://api.jquery.com/one/

asutherland: Ah, I see, you're transforming the entire DOM the first time someone starts to select something.  It'd be good to have a comment to that effect and possibly switch to using addEventListener directly with `{ once: true }`.  (We're planning to rip out jQuery.)

static/js/code-highlighter.js
```diff
@@ -496,3 +496,17 @@ $(function () {
...
 
 });
+
+// We use user-select:none to hide line numbers from text selections. That
+// does however not work in Chrome (https://crbug.com/850685)
+// or Safari (https://bugzilla.mozilla.org/show_bug.cgi?id=1616104#c1).
+// As a work-around, move all line numbers into pseudo-elements when the user
+// selects something for the first time.
+if (navigator.userAgent.indexOf('Firefox') == -1) {
+  document.addEventListener('selectstart', function() {
```
emilio: Can we just move the line-number to be a pseudo-element instead? Maybe a CSS counter?

Rob--W: The line number is already moved to a pseudo-element.

Does a CSS counter have any advantage over the attribute? Improved performance, possibly?

The change would be trivial, so let me know what you prefer. The current PR or just emptying the textContent, combined with the following style:

```
#file {
  counter-reset: lineno;
}
.line-number:empty::before {
  content: counter(lineno);
  counter-increment: lineno;
}
```

emilio: I mean, can we always use a pseudo-element, for all browsers, rather than using the UA string?

Rob--W: We could. It's basically a revert of https://bugzilla.mozilla.org/show_bug.cgi?id=1558942

I didn't measure the impact on performance or other side effects.

The only "feature" that I know of is that the current implementation allows Ctrl-F to find line numbers, whereas browser search won't find content in pseudo-elements.

PR #279, opened 2020-02-21
Show up message about KEEP_WORKING variable

dpino:
Although the KEEP_WORKING variable is meant to be used to preserve the current working directory, an user may rerun "indexer-setup.sh" without defining this variable, causing the removal of all contents inside the WORKING directory (repository checkout, blame information, etc).

This patch asks users to confirm whether they would like to remove the contents of WORKING directory, when KEEP_WORKING is not defined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,6 +279,11 @@ mkdir ~/mozilla-index
 /vagrant/infrastructure/indexer-run.sh ~/mozilla-config ~/mozilla-index
 ```
 
+Note: By default, `indexer-setup.sh` removes the contents of the working
+directory (in the example above, that's `~/mozilla-index`). In case you would
+like to keep the contents of the working directory, define KEEP_WORKING=1
+when calling `indexer-setup.sh`.
+
 ## Background on Mozsearch indexing
 
 The Mozsearch indexing process has three main steps, depicted here:
--- a/infrastructure/indexer-setup.sh
+++ b/infrastructure/indexer-setup.sh
@@ -16,7 +16,7 @@ CONFIG_INPUT="$2"
 export WORKING=$(readlink -f $3)
 
 if [ -z "${KEEP_WORKING:-}" ]; then
-    echo "Removing old contents of $WORKING/"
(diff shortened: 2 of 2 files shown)


staktrace:
Isn't this going to break any automated mechanism that runs this script, because it's going to wait for user input? In particular when the searchfox indexing runs in production it uses this script and so that will just hang.

I think it might be better to just emit a warning to the user informing them that they can use KEEP_WORKING=1 to keep the working directory, and maybe just document this better in the developer docs.

dpino:
To feed the script with input when run automatically, it could be possible to run it like: `$ yes n | indexer-setup.up` (same effect as `KEEP_WORKING=1 indexer-setup.sh ...`). Anyway, since Mozilla is the major user of mozsearch I'm all in for a solution that causes you the less trouble.

I think printing out a message informing abut KEEP_WORKING=1 could work.

staktrace:
> To feed the script with input when run automatically, it could be possible to run it like: `$ yes n | indexer-setup.up` (same effect as `KEEP_WORKING=1 indexer-setup.sh ...`).

While this is technically possible it would be quite inconvenient to have to add this everywhere we invoke this script from.

> I think printing out a message informing abut KEEP_WORKING=1 could work.

Great. Could you update the PR so that the existing "Removing old contents" message mentions the KEEP_WORKING variable? Adding a note to [this section](https://github.com/mozsearch/mozsearch/#testing-basic-changes) and maybe other sections of the README would also be good.

asutherland:
I think it's definitely the case that for humans interactively iterating on things in searchfox that the UX could be much improved.  Right now https://github.com/mozsearch/mozsearch/blob/master/Makefile is the avenue for that, but it would certainly be neat if we could use something like the `mach` tool used in mozilla-central (or some other smart-ish but not overwhelmingly huge dependency) as the preferred means of people hacking on searchfox.  Also, it might help us move away from all the bash scripts...

Also maybe for tooling that doesn't assume an AWS instance that's freshly created every time it runs...

PR #282, opened 2020-03-06
Bug 1620597 - Handle the case where a commit replaces a submodule or …

staktrace:
…folder with a file

If the repo has a folder at a particular path, and somebody adds a commit
that replaces the folder with a file, the blame script would fail because
it would assume that the object at that path in the parent commit was a file
and it would try diffing them. This patch adds a check to ensure that the
object at the path in the parent commit is in fact a file (Blob) before
continuing.


staktrace:
(Just need a single review, whoever gets to it first)

PR #288, opened 2020-03-20
Bug 1623903 - Add support for OptionalExpression and OptionalMemberEx…

staktrace:
…pression.


staktrace:
Good catch, thanks! Updated patch to add `OptionalCallExpression`

PR #289, opened 2020-03-24
Update ipdl_parser (and get nightly-est rustc)

staktrace:


infrastructure/indexer-update.sh
```diff
@@ -22,6 +22,16 @@ set -o pipefail # Check all commands in a pipeline
...
 # versions of Rust.
+# Before we do the update, we remove some components that we don't need and
+# that are sometimes missing. If they are missing, `rustup update` will try
+# to use a previous nightly instead that does have the components, which means
+# we end up with a slightly older rustc. Using rustc from a few days ago is
+# usualy fine, but in cases where we hit ICEs that have been fixed upstream,
+# we want the very latest rustc to get the fix. Removing these components also
+# reduces download time during `rustup update`.
+rustup component remove clippy
+rustup component remove rustfmt
```
emilio: Hmm, we probably want `rustfmt`. Without it Gecko's bindgen-generated files would be all on one line.

Though maybe this is not an issue for the builds on automation, so if it makes your life easier go ahead.

staktrace: Yeah I don't think we use `rustfmt` in the indexer, the formatting must happen in the taskcluster build.

PR #294, opened 2020-04-03
Bug 1627339 - Expose raw analysis files and {repo,objdir}-files. r=kats

asutherland:
Expose the analysis files as "raw-analysis" and the "repo-files" and
"objdir-files" under "file-lists".  This is intended to make debugging
investigations easier as well as to enable extraction of per-file and
per-directory information without having to scrape the HTML output of
output-file.

### DIFF
--- a/scripts/nginx-setup.py
+++ b/scripts/nginx-setup.py
@@ -1,11 +1,28 @@
 #!/usr/bin/env python
 
+# Create our nginx configuration.
+#
+# The general scheme is `TREE/SERVICE/...` where SERVICE is "source" or
+# "raw-analysis" for files available on disk and various dynamic requests that
+# get proxied to per-tree local servers running on localhost.
+#
+# We have a docroot at /home/ubuntu/docroot that provides a place to decide what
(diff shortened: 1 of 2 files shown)


scripts/nginx-setup.py
```diff
@@ -163,7 +200,10 @@ def location(route, directives):
 
     # Handled by router/router.py
     location('/%(repo)s/search', ['proxy_pass http://localhost:8000;'])
+    location('/%(repo)s/sorch', ['proxy_pass http://localhost:8000;'])
     location('/%(repo)s/define', ['proxy_pass http://localhost:8000;'])
+    location('/%(repo)s/symbol', ['proxy_pass http://localhost:8000;'])
```
staktrace, line 205: This hunk shouldn't be here, I think?

asutherland: Argh, sorry, yes.  I was cherry-picking deltas and there was an additional hunk I wanted to get the text/plain comment from without taking other changes in the patch and I checked out the tip revision, forgetting that I'd explicitly avoided cherry-picking certain revisions.

PR #296, opened 2020-04-05
Bug 1627529 - Split the blame-using "searchfox" test config out from the non-blame-using "tests" test config. r=kats

asutherland:


tests/searchfox-config.json
```diff
@@ -0,0 +1,17 @@
+{
+  "mozsearch_path": "$MOZSEARCH_PATH",
+
+  "default_tree": "searchfox",
+
+  "trees": {
+    "searchfox": {
+      "index_path": "$WORKING/searchfox",
+      "files_path": "$MOZSEARCH_PATH",
```
staktrace, line 9: Why the change to `files_path` here? Does this change mean that [this line](https://github.com/mozsearch/mozsearch/blob/ab35fab17cc1d78975094530fb15884a6149944a/tests/searchfox/setup#L8) can be removed?

asutherland: I think I experienced a patherror, said something to myself like "how did this ever work?" without realizing that line existed, and just changed it.  I'll try removing the line and seeing what happens, etc.

Thanks for all the eagle-eye reviews!  Your diligence definitely pays off on these "fancy" path extractions!

Makefile
```diff
@@ -30,6 +30,13 @@ build-test-repo: check-in-vagrant build-clang-plugin build-rust-tools
 	/vagrant/infrastructure/web-server-setup.sh /vagrant/tests config.json ~/index ~
 	/vagrant/infrastructure/web-server-run.sh /vagrant/tests ~/index ~
 
+build-searchfox-repo: check-in-vagrant build-clang-plugin build-rust-tools build-ui
```
asutherland, line 33: whoops, `build-ui` slipped in here, removed locally.

tests/searchfox-config.json
asutherland: Removing the line worked fine.  I'm going to push with this fix.

asutherland:
Okay, so I tested locally (which I was always planning to do) with the line removed, etc. and all is good.  I also updated the docs slightly to make it clear that we populate `~/searchfox-index` rather than `~/index` when I got confused by why things in `~/index` weren't as I expected them when investigating the `files` thing.

@staktrace One thing I did run into is that in a steady-state iteration for `tests` one now really needs to do `CLEAN_WORKING=1 make build-test-repo` instead of `make build-test-repo`.  Perhaps we should make the Makefile specify that environment variable by default?  It's not clear to me that we'd ever want to do anything but that for the tests repo.

asutherland:
(I'm leaving the CLEAN_WORKING=1 thing to another bug/issue/pull request.  But rs=asuth if you want to make that change and/or I'm happy to make the change and request review, etc.  There are a variety of ways to make that happen methinks.)

staktrace:
> (I'm leaving the CLEAN_WORKING=1 thing to another bug/issue/pull request. But rs=asuth if you want to make that change and/or I'm happy to make the change and request review, etc. There are a variety of ways to make that happen methinks.)

I tried running `build-test-repo` twice and the failure I got was due to the symbolic link creation for `tests/files`. Changing `ln -s` to `ln -s -f` seems to fix it for me, as it forces the link creation even if it already exists. I can push a patch with that.

PR #297, opened 2020-04-06
Bug 1627529 follow-up - Force-create the symbolic link to avoid error…

staktrace:
…ing out if build-test-repo is run twice in a row

### DIFF
--- a/tests/tests/setup
+++ b/tests/tests/setup
@@ -5,4 +5,4 @@ set -eu # Errors/undefined vars are fatal
 set -o pipefail # Check all commands in a pipeline
 
 rm -f $INDEX_ROOT/tests
-ln -s $CONFIG_REPO/tests/files $INDEX_ROOT
+ln -s -f $CONFIG_REPO/tests/files $INDEX_ROOT


staktrace:
Landing with rs=asuth from https://github.com/mozsearch/mozsearch/pull/296#issuecomment-609468060, but feel free to post-review if I misunderstood the problem and there's some other reason we might need `CLEAN_WORKING=1`

tests/tests/setup
```diff
@@ -5,4 +5,4 @@ set -eu # Errors/undefined vars are fatal
 set -o pipefail # Check all commands in a pipeline
 
 rm -f $INDEX_ROOT/tests
```
asutherland, line 7: To be clear, I was mainly assuming that the transition from needing to pass KEEP_WORKING to needing to pass CLEAR_WORKING was an undesired transition for the test repo, I hadn't really dug into that assumption.

Maybe this should be `rm -rf` instead?  It seems like that might have been the intent of this rm.  (Which then is effectively CLEAN_WORKING but without the footgun if you don't use it.)

With these changes as-is, the `analysis` directory doesn't get cleaned and so we experience the problem where the C++ analysis files end up containing the union of the current run and all previous runs because of how the C++ indexer merges new data into previous data.  A quick smoke test on this is to do something like add 1 blank line near the top of `big_cpp.cpp` and run make again and then observe that the analysis file is full of duplicate entries.

staktrace: Ah, I see. My intent was that not passing `CLEAN_WORKING=1` should still produce correct results, but I missed that the c++ analysis can get duplicated. I can take a closer look at this and figure out what should happen.

PR #299, opened 2020-04-07
Bug 1593833 - Update path normalizations and apply to rust. r=kats

asutherland:


scripts/rust-analyze.sh
```diff
@@ -35,6 +35,8 @@ PLATFORM_SAVE_FILTER=${PLATFORM:+${RUST_PLAT_DIRS[$PLATFORM]}/debug/deps/}
 PLATFORM_FILTER=${PLATFORM:+${RUST_PLAT_DIRS[$PLATFORM]}/}
 PLATFORM_SUFFIX=${PLATFORM:+-$PLATFORM}
 
+SED_TRANSFORM_EXPR=$4
```
staktrace, line 38: Won't this trigger an error if `$4` isn't defined?

scripts/rust-analyze.sh
```diff
@@ -49,6 +51,11 @@ if [ -d "$OBJDIR" ]; then
   if [ -d "$INDEX_ROOT/rustlib" ]; then
     ANALYSIS_DIRS="$ANALYSIS_DIRS $(find $INDEX_ROOT/rustlib -type d -path \*/${PLATFORM_FILTER}analysis)"
   fi
+  # If a sed normalization command was provided, perform the transform over all
+  # json analysis files.
+  if [[ $SED_TRANSFORM_EXPR ]]; then
+    find $ANALYSIS_DIRS -type f -name \*.json | parallel --halt now,fail=1 sed --in-place -e "\"$SED_TRANSFORM_EXPR\"" {}
```
staktrace, line 57: I'd prefer `"*.json"` to `\*.json` (i.e. quotes instead of backslash. I'm also not sure why the extra backslashed-quotes are required around `$SED_TRANSFORM_EXPR` here.

scripts/rust-analyze.sh
asutherland: It did in fact trigger an error!

scripts/rust-analyze.sh
asutherland: I'll change the wildcarding as I now push again with the $4 fix.  (I plead incompetence.)

When locally playing with this it seemed necessary.  Here's my noodling with all my cool prompt stuff making it hard to read.
```
 visbrero  …  code  searchfox-play  test                     
 $  SED_TRANSFORM_EXPR="s#builds/worker/workspace/build/src/##g;s#z:/task_[0-9]*/build/src/##gI"
 visbrero  …  code  searchfox-play  test                     
 $  /usr/bin/find $ANALYSIS_DIRS -type f -name \*.json | parallel --halt now,fail=1 sed --in-place -e "$SED_TRANSFORM_EXPR" {}
sed: no input files
/bin/bash: s#z:/task_[0-9]*/build/src/##gI: No such file or directory
parallel: This job failed:
sed --in-place -e s#builds/worker/workspace/build/src/##g;s#z:/task_[0-9]*/build/src/##gI foo/kaboom.json
 visbrero  …  code  searchfox-play  test  0  127                     
 0  127  $  /usr/bin/find $ANALYSIS_DIRS -type f -name \*.json | parallel --halt now,fail=1 sed --in-place -e "\"$SED_TRANSFORM_EXPR\"" {}
 visbrero  …  code  searchfox-play  test                     
```

scripts/rust-analyze.sh
```diff
@@ -49,6 +51,11 @@ if [ -d "$OBJDIR" ]; then
   if [ -d "$INDEX_ROOT/rustlib" ]; then
     ANALYSIS_DIRS="$ANALYSIS_DIRS $(find $INDEX_ROOT/rustlib -type d -path \*/${PLATFORM_FILTER}analysis)"
   fi
+  # If a sed normalization command was provided, perform the transform over all
+  # json analysis files.
+  if [[ $SED_TRANSFORM_EXPR ]]; then
+    find $ANALYSIS_DIRS -type f -name "*.json" | parallel --halt now,fail=1 sed --in-place -e "'$SED_TRANSFORM_EXPR'" {}
```
staktrace, line 57: I still find it weird that we need two sets of quotes here, but maybe `parallel` is munging things. I guess this is fine.

asutherland:
Sorry for all the force-push spam.  The verification checks seem happy now but I'm fixing a glitch with the rust stdlib (which I hadn't added integrity checks for yet, and in the process realized it was still broken).

asutherland:
Okay, I'm re-requesting review on both PR's now.  Note that there's now 2 commits in each PR.  The first is for the improved path normalization and improved logging for rust generation, the second is for the "check" operations.  I very cleverly made every possible permutation of error, thereby verifying that we have useful logs and that the check operations correctly detect mess-ups.

Note that the check operation only runs a disk check as part of the indexing, but there is a web check, and you can manually run it on the "asuth" channel web server that has the results right now.  You would run `~/mozsearch/infrastructure/web-server-check.sh ~/config ~/index ~`

asutherland:
(And a successful run is available at https://asuth.searchfox.org/ too, of course.)

scripts/rust-analyze.sh
```diff
@@ -1,58 +1,64 @@
...
   if [ "x$ANALYSIS_DIRS" = "x" ]; then
     exit 0 # Nothing to analyze really.
   fi
-  # If we have rust stdlib sources and analysis data, pick that up too
-  if [ -d "$INDEX_ROOT/rustlib" ]; then
-    ANALYSIS_DIRS="$ANALYSIS_DIRS $(find $INDEX_ROOT/rustlib -type d -path \*/${PLATFORM_FILTER}analysis)"
+
+  # Rust stdlib files use `analysis` directories instead of `save-analysis`, so
+  # even though they live under the same root, it needs a separate find pass
+  # because the above will not have foun them.
```
staktrace, line 51: s/foun/found/

tools/src/bin/rust-indexer.rs
```diff
@@ -32,12 +32,42 @@ pub struct Defs {
...
+    /// prefixed with `__GENERATED__`.
+    generated: &'a Path,
+    /// The searchfox path space prefix for generated.
+    generated_friendly: &'a Path,
+    /// Local filesystem path root for the rust stdlib source.  In the searchfox
+    /// path space presented to users, this means paths prefixed with
+    /// `__GENERATED__/__RUST__`.
     libstd: &'a Path,
+    /// The searchfox path space prefix for the rust stdlib, assumed to be a
+    /// suffix of objdir_friendly.
```
staktrace, line 69: Assuming this comment should refer to `generated_friendly` instead of `objdir_friendly`. Also, not sure if suffix is the right word here. Extension, maybe? Or maybe reword it to say `generated_friendly` is a prefix of this?

scripts/check-helper.sh
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env bash
+
+set -x # Show commands
+set -eu # Errors/undefined vars are fatal
+set -o pipefail # Check all commands in a pipeline
+
+if [[ $# -ne 4 ]]
+then
+    set +x  # Turn off echoing of commands and output only relevant things to avoid bloating logfiles
```
staktrace, line 9: A little unsure that you need the `set +x` here, since in general this path shouldn't get hit much in practice and if it does spewing one extra "echo blah blah" command isn't the end of the world. But I don't mind leaving it here.

tools/src/bin/rust-indexer.rs
asutherland: Yes, your generated_friendly phrasing is much more clear, thank you.  (And indeed there was a late rename from objdir_friendly to generated_friendly that missed this.)

scripts/check-helper.sh
asutherland: I mainly just copied and pasted the existing "set +x" boilerplate to clarify what was happening and in case it got copied onward.  The main intent here was to avoid the echo command line being confused for its output, or rather, distracting the eye (which happened for me).  So I'll leave as-is with the consensus that log filling isn't a real problem for these.

PR #300, opened 2020-04-16
Bug 1630593 - Use the in-tree xpidl parser if there is one.

staktrace:


staktrace:
Works as intended for the test repo, but still testing m-c.

staktrace:
Verified works on m-c as well.

PR #302, opened 2020-04-18
Bug 1631191 - Allow easily restoring interrupted state to local NVME …

staktrace:
…disk

### DIFF
--- a/infrastructure/aws/mkscratch.sh
+++ b/infrastructure/aws/mkscratch.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+
+set -x # Show commands
+set -eu # Errors/undefined vars are fatal
+set -o pipefail # Check all commands in a pipeline
+
+# Create the "index-scratch" directory where each specific tree's indexing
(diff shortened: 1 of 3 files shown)


infrastructure/aws/mkscratch.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
+set -x # Show commands
+set -eu # Errors/undefined vars are fatal
+set -o pipefail # Check all commands in a pipeline
+
+# Create the "index-scratch" directory where each specific tree's indexing
```
staktrace, line 7: This is a straight move from the other file, plus header decoration stuff (hashbang and `set` commands)

staktrace:
My motivation here was actually not for debugging failed indexers, but spinning up an indexer that I can use to do arbitrary testing stuff on. Now I can push a branch with an `exit 1` somewhere in the bash scripts, wait for the failure email, ssh in and run these commands to get a working setup for manual experimentation which is handy.

PR #303, opened 2020-04-18
Rewrite and simplify blame popup.

emilio:
With the goal of making it faster.

Right now the blame popup causes pretty expensive reflows, because it's
hosted inside the line strip.

This is completely unnecessary though, so this patch saves us some
per-line markup and makes it a top-level absolutely-positioned element,
and positions it dynamically.

This also uses aria-owns so that the accessibility semantics are the
same as before.

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -57,10 +57,13 @@ table {
     box-sizing: border-box;
 }
 
-body {
+:root {
+    font: 12px/1.5 Arial, Helvetica, sans-serif;
     background-color: #fff;
     color: #333;
-    font: 12px/1.5 Arial, Helvetica, sans-serif;
+}
+
+body {
     margin: 0;
 }
 h4 {
@@ -622,13 +625,6 @@ span[data-symbols].hovered {
 }
 
 /* ## Blame ## */
-/* the role=cell that holds the role=button `.blame-strip`.  We created a class
-   for this so we could make it `position: relative` so it could create a new
-   containing block for the `.blame-popup` which cannot live beneath the
-   role=button `.blame-strip`. */
-.blame-container {
-  position: relative;
-}
 .blame-strip {
     display: block;
     width: 20px;
@@ -646,12 +642,10 @@ span[data-symbols].hovered {
 }
 
 .blame-popup {
-    display: inline !important;
-    /* positioned inside the .blame-container's containing block created by it
-       being position: relative */
     position: absolute;
+    /* positioned by BlamePopup in blame.js */
     top: 0;
-    left: 20px;
+    left: 0;
     width: 600px;
     padding: 10px;
     background: #404040;
--- a/tools/src/format.rs
+++ b/tools/src/format.rs
@@ -520,10 +520,7 @@ pub fn format_file_data(
             )),
             F::Indent(vec![
                 // Blame info.
-                F::T(format!(
-                    "<div role=\"cell\" class=\"blame-container\"><div{}></div></div>",
-                    blame_data
-                )),
+                F::T(format!("<div role=\"cell\"><div{}></div></div>", blame_data)),
                 // The line number.
                 F::T(format!(
                     "<div id=\"l{}\" role=\"cell\" class=\"line-number\">{}</div>",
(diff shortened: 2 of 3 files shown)


asutherland:
Have you tested on NVDA or similar?  If not (and you can't easily do so), could you trigger a "dev" channel run so I can test on my windows machine with NVDA?  Thanks!

emilio:
> Have you tested on NVDA or similar? If not (and you can't easily do so), could you trigger a "dev" channel run so I can test on my windows machine with NVDA? Thanks!

I've tested with Firefox's a11y inspector and with orca. But more testing is always welcome, will push a run.

emilio:
Hmm, that didn't go as planned:

```
++ pushd /mnt/index-scratch/mozilla-central
/mnt/index-scratch/mozilla-central ~
++ PREEXISTING_HG_REV=
++ '[' -f target.json ']'
++ curl -SsfL --compressed https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/gecko.v2.mozilla-central.latest.firefox.linux64-searchfox-debug/artifacts/public/build/target.json
curl: (22) The requested URL returned error: 502 
++ handle_error
++ mkdir /index/interrupted
++ mv -f /mnt/index-scratch/config.json /mnt/index-scratch/mozilla-central /mnt/index-scratch/nss /index/interrupted
++ /home/ubuntu/mozsearch/infrastructure/aws/send-failure-email.py '[dev/blame-fast]' 
```

That url does seem dead... Any idea @asutherland? Alternatively can you pull this patch and try it locally?

staktrace:
Bugzilla is down too, probably server maintenance.

staktrace:
The URL is working now, so doing a dev push should work again.

static/css/mozsearch.css
```diff
@@ -667,6 +661,7 @@ span[data-symbols].hovered {
     -webkit-user-select: text;
     -ms-user-select: text;
     user-select: text;
+    /* transition: transform .1s ease; */
```
staktrace, line 664: remove if not needed?

emilio:
So this is up in https://dev.searchfox.org/, you can try blaming https://dev.searchfox.org/whatwg-html/source/source to see the difference.

I moved aria-owns to the parent `<div role=cell>` per Jamie's advice in the bug. @asutherland do you want to give it a try on NVDA if you want? On Orca the experience is not great because the `role=button` elements are not focusable, fwiw...

static/css/mozsearch.css
emilio: Yup, this was leftover, I wasn't sure if it'd look good, it doesn't :)

asutherland:
Sorry for the delay, I'd forgotten the Windows machine in question was in the process of being physically migrated across the room in pursuit of being hooked up to even more monitors.  The machine is all set up now, though, so future checks should be much faster.

There seemed to be a regression in the NVDA use-case as I understand it, I've commented at https://bugzilla.mozilla.org/show_bug.cgi?id=1611729#c8 and set a needinfo on :Jamie.

asutherland:
> This looks reasonable to me overall. Very happy to see the code modernization and removal of jquery stuffs too :) Excited to try it out, it might also fix bug 1561076 maybe?

Yes, this fixes that (which makes sense).  Specifically, using your example from that bug with updated line numbers:
- On trunk, go to https://searchfox.org/mozilla-central/source/browser/base/content/browser-fullScreenAndPointerLock.js#579 and hover line 575 and see it's broken
- On dev, go to https://dev.searchfox.org/mozilla-central/source/browser/base/content/browser-fullScreenAndPointerLock.js#579 and hover line 575 and see it's fixed

tools/src/format.rs
```diff
@@ -500,7 +500,7 @@ pub fn format_file_data(
         } else {
             // If we have no blame data, we want the div here taking up space, but we don't want it
             // to bother screen readers.
```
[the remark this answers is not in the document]
emilio: Yup, removed! Yeah, the aria-hidden was broken, should've been `aria-hidden="false"` instead of `role="aria-hidden"`, see the commit message.

PR #308, opened 2020-04-23
Bug 1619751 - Pin blame popup on click, dismiss on scroll or click elsewhere.

staktrace:
I live-edited the dev web-server to test this, so the patch is applied there and you can try it out if you want.

Also note that if you test in Fenix you may run into https://github.com/mozilla-mobile/fenix/issues/8779 which reopens the blame popup after scrolling. Doesn't happen in Fennec, and I intend to track down and fix the bug in Fenix soonish.

### DIFF
--- a/static/js/blame.js
+++ b/static/js/blame.js
@@ -156,15 +156,22 @@ var BlameStripHoverHandler = new (class BlameStripHoverHandler {
     // element will null this out as a means of letting the user get rid of the
     // popup if they didn't actually want to see the pop-up.
     this.mouseElement = null;
+    // Set to true if the user clicks on the blame strip. This will keep the
+    // keep the popup visible until the user clicks elsewhere.
+    this.keepVisible = false;
 
     for (let element of document.querySelectorAll(".blame-strip")) {
       element.addEventListener("mouseenter", this);
       element.addEventListener("mouseleave", this);
-      element.addEventListener("click", this);
     }
 
     BlamePopup.popup.addEventListener("mouseenter", this);
     BlamePopup.popup.addEventListener("mouseleave", this);
+    // Click listener needs to be capturing since whatever is being clicked on
+    // (e.g. a code fragment that displays a context menu) may actually
+    // consume the event.
+    window.addEventListener("click", this, {capture: true});
+    document.getElementById("scrolling").addEventListener("scroll", this, {passive: true});
   }
 
   handleEvent(event) {
@@ -172,6 +179,22 @@ var BlameStripHoverHandler = new (class BlameStripHoverHandler {
     if (ContextMenu.active) {
       return;
     }
+
+    if (this.keepVisible) {
+      if (event.type == "mouseenter" || event.type == "mouseleave") {
+        // Ignore mouseenter/mouseleave events if keepVisible is true
+        return;
+      }
+    }
+
+    let clickedOutsideBlameStrip =
+        event.type == "click" &&
+        !event.target.matches(".blame-strip");
+    if (clickedOutsideBlameStrip && !BlamePopup.blameElement) {
+      // Don't care about clicks outside the blame strip if there's no popup showing.
+      return;
+    }
+
     // Debounced pop-up closer..
     //
     // If the mouse leaves a blame-strip element and doesn't move onto another
@@ -180,8 +203,10 @@ var BlameStripHoverHandler = new (class BlameStripHoverHandler {
     // pop-up.
     if (
       event.type == "mouseleave" ||
-      (event.type == "click" && BlamePopup.blameElement)
+      event.type == "scroll" ||
+      clickedOutsideBlameStrip
     ) {
+      this.keepVisible = false;
       this.mouseElement = null;
       setTimeout(() => {
         if (this.mouseElement) {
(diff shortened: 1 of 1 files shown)


emilio:
> Also note that if you test in Fenix you may run into mozilla-mobile/fenix#8779 which reopens the blame popup after scrolling. Doesn't happen in Fennec, and I intend to track down and fix the bug in Fenix soonish.

I suspect it won't happen with `layout.reflow.synthMouseMove=false`. Seems very similar to https://github.com/mozilla-mobile/fenix/issues/9860.

static/js/blame.js
```diff
@@ -156,22 +156,42 @@ var BlameStripHoverHandler = new (class BlameStripHoverHandler {
...
     for (let element of document.querySelectorAll(".blame-strip")) {
       element.addEventListener("mouseenter", this);
       element.addEventListener("mouseleave", this);
-      element.addEventListener("click", this);
     }
 
     BlamePopup.popup.addEventListener("mouseenter", this);
     BlamePopup.popup.addEventListener("mouseleave", this);
+    window.addEventListener("click", this, {capture: true});
+    document.getElementById("scrolling").addEventListener("scroll", this, {capture: true});
```
emilio, line 171: `passive: true`?

static/js/blame.js
```diff
@@ -156,22 +156,42 @@ var BlameStripHoverHandler = new (class BlameStripHoverHandler {
...
+    if (this.keepVisible) {
+      if (event.type == "mouseenter" || event.type == "mouseleave") {
+        // Ignore mouseenter/mouseleave events if keepVisible is true
+        return;
+      }
+    }
+
+    let clickedOutsideBlameStrip =
+        event.type == "click" &&
+        event.originalTarget.closest(".blame-strip") == null;
```
emilio, line 189: `.originalTarget` is Firefox-specific, probably `.target` works.

emilio: Also probably `!...closest()` instead of `== null`?

static/js/blame.js
```diff
@@ -190,6 +212,12 @@ var BlameStripHoverHandler = new (class BlameStripHoverHandler {
         BlamePopup.blameElement = null;
       }, 100);
     } else {
+      // We run this code on either "mouseenter", or on a "click" event where
+      // the click landed inside the blame strip. In the latter case we set
+      // keepVisible to pin the blame popup open until another click dismisses
+      // it, or a scroll event happens (since the popup doesn't move properly with
+      // scrolling).
+      this.keepVisible = (event.type == 'click');
```
emilio, line 220: nit: I'd remove these parens but your call :)

static/js/blame.js
```diff
@@ -180,8 +200,10 @@ var BlameStripHoverHandler = new (class BlameStripHoverHandler {
     // pop-up.
     if (
       event.type == "mouseleave" ||
-      (event.type == "click" && BlamePopup.blameElement)
+      event.type == "scroll" ||
+      (event.type == "click" && clickedOutsideBlameStrip)
```
emilio, line 204: instead of `(event.type == "click" && clickedOutsideBlameStrip)` just `clickedOutsideBlameStrip`.

static/js/blame.js
emilio: Also why does this need to be capturing? Same for the scroll listener.

staktrace:
> I suspect it won't happen with `layout.reflow.synthMouseMove=false`. Seems very similar to [mozilla-mobile/fenix#9860](https://github.com/mozilla-mobile/fenix/issues/9860).

Indeed you are correct. Thanks for the tip!

static/js/blame.js
emilio: And actually, you need to make it `!....matches(".blame-strip")`. `.blame-strip` elements are empty anyways, so it's cheaper, and the code relies on the `blameElement` being the `.blame-strip` element being the target element.

static/js/blame.js
staktrace: > `passive: true`?

I don't think passive does anything for scroll events. At least APZ doesn't care about whether scroll listeners are passive or not. Maybe I'm mistaken?

staktrace: > Also why does this need to be capturing? Same for the scroll listener.

The click needs to be capturing because if you e.g. click on a code fragment that produces a context menu, it seems to consume the click and then we end up with both the blame popup and the context menu. Having a capture listener here ensures we get the click even in that case. I can add a comment to that effect. The scroll listener probably doesn't need to be capturing, I can drop that.

emilio: > I don't think passive does anything for scroll events. At least APZ doesn't care about whether scroll listeners are passive or not. Maybe I'm mistaken?

Ah, yeah, maybe I was thinking of wheel events, nvm.

> The click needs to be capturing because if you e.g. click on a code fragment that produces a context menu, it seems to consume the click and then we end up with both the blame popup and the context menu. Having a capture listener here ensures we get the click even in that case. I can add a comment to that effect. The scroll listener probably doesn't need to be capturing, I can drop that.

Ah, so this is so that we don't hit the `event.stopPropagation()` in the context-menu code? That should probably be changed to something else... And actually given now we don't always recreate the menu it may not be needed at all anymore.

static/js/blame.js
staktrace: I like the parens so I kept them, but updated for all the rest of your comments. :)

static/js/blame.js
staktrace: I changed /capture/passive/ for the scroll listener. The `passive` won't help but it won't hurt either, so whatever.

For the click - I guess it's the stopPropagation() in the context-menu code, but I didn't actually verify. In general it seems like for things like this a capturing listener seems more robust.

emilio: It seems more reliable for hiding the popup, but not for showing it, right? Anyhow nbd.

staktrace: > It seems more reliable for hiding the popup, but not for showing it, right? Anyhow nbd.

Ah, good point. I was only considering the hiding scenario at the time but you're right it probably is not as good for the showing scenario.

PR #309, opened 2020-04-24
Bug 1632951 - Don't hide the blame popup if the user is expanding the…

staktrace:
… details expando

### DIFF
--- a/static/js/blame.js
+++ b/static/js/blame.js
@@ -194,6 +194,11 @@ var BlameStripHoverHandler = new (class BlameStripHoverHandler {
       // Don't care about clicks outside the blame strip if there's no popup showing.
       return;
     }
+    if (clickedOutsideBlameStrip && BlamePopup.popup.contains(event.target)) {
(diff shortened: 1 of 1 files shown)


staktrace:
It's live on dev, I might apply it to the current `release` web server too.

static/js/blame.js
```diff
@@ -194,6 +194,11 @@ var BlameStripHoverHandler = new (class BlameStripHoverHandler {
       // Don't care about clicks outside the blame strip if there's no popup showing.
       return;
     }
+    if (clickedOutsideBlameStrip && event.target.closest("#blame-popup")) {
```
emilio, line 197: Maybe change the `.closest` call to be `BlamePopup.popup.contains(event.target)`?

staktrace: Ah, thanks. I was looking for something like that but I was looking through `Element` functions on MDN and `contains` is a `Node` function so I didn't see it...

PR #311, opened 2020-04-26
Initial implementation of rusty blame builder

staktrace:
This should be 100% hash-compatible with the python version. The code structure is very similar as well, as I basically ported it from the python code directly. I will be restructuring a bunch of this code later to make it more amenable to parallelization.

### DIFF
--- a/tools/src/bin/build-blame.rs
+++ b/tools/src/bin/build-blame.rs
@@ -8,30 +8,38 @@ extern crate unicode_normalization;
 use std::borrow::Borrow;
 use std::collections::HashMap;
 use std::env;
+use std::io::{BufRead, BufReader, Write};
 use std::path::{Path, PathBuf};
-use std::process::Command;
+use std::process::{Child, Command, Stdio};
 
 use git2::{DiffFindOptions, ObjectType, Oid, Patch, Repository, Sort};
 use tools::config::index_blame;
-use tools::git_ops;
 use unicode_normalization::UnicodeNormalization;
 
-fn get_hg_rev(git_repo: &Repository, git_oid: &Oid) -> Option<String> {
-    // TODO: start a helper process and stream I/O to it instead of spawning a new process every time
-    let output = Command::new("git")
-        .arg("cinnabar")
-        .arg("git2hg")
-        .arg(format!("{}", git_oid))
+fn get_hg_rev(helper: &mut Child, git_oid: &Oid) -> Option<String> {
+    write!(helper.stdin.as_mut().unwrap(), "git2hg {}\n", git_oid).unwrap();
+    let mut reader = BufReader::new(helper.stdout.as_mut().unwrap());
+    loop {
+        let mut result = String::new();
+        reader.read_line(&mut result).unwrap();
+        if result.starts_with("changeset ") {
+            return result.split(' ').nth(1).map(str::trim).map(str::to_string);
+        }
+        if result.trim().chars().all(|c| c == '0') {
+            return None;
+        }
+        // else loop again. cinnabar-helper emits a bunch of other info that we don't care about
+    }
+}
+
+fn start_cinnabar_helper(git_repo: &Repository) -> Child {
+    Command::new("git")
+        .arg("cinnabar-helper")
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
         .current_dir(git_repo.path())
-        .output()
-        .unwrap();
-    if !output.status.success() {
-        panic!("`git cinnabar git2hg {}` returned error code {:?}", git_oid, output.status.code());
-    }
-    match git_ops::decode_bytes(output.stdout).lines().next() {
-        Some(r) if r != "0000000000000000000000000000000000000000" => Some(r.to_string()),
-        _ => None,
-    }
+        .spawn()
+        .unwrap()
 }
 
 fn count_lines(blob: &git2::Blob) -> usize {
@@ -279,6 +287,11 @@ fn main() {
     let git_repo = Repository::open(&args[1]).unwrap();
     let blame_repo = Repository::open(&args[2]).unwrap();
     let use_cinnabar = env::var("CINNABAR").map(|v| v == "1").unwrap_or(false);
+    let mut hg_helper = if use_cinnabar {
+        Some(start_cinnabar_helper(&git_repo))
+    } else {
+        None
+    };
     let blame_ref = env::var("BLAME_REF").ok().unwrap_or("HEAD".to_string());
 
     info!("Reading existing blame map of ref {}...", blame_ref);
@@ -310,11 +323,9 @@ fn main() {
     for git_oid in revs_to_process {
         rev_done += 1;
 
-        let hg_rev = if use_cinnabar {
-            get_hg_rev(&git_repo, &git_oid)
-        } else {
-            // we don't support mapfiles any more.
-            None
+        let hg_rev = match hg_helper {
+            Some(ref mut helper) => get_hg_rev(helper, &git_oid),
+            None => None, // we don't support mapfiles any more.
         };
 
         info!(
@@ -404,5 +415,9 @@ fn main() {
 
         blame_map.insert(git_oid, blame_oid);
         info!("  -> {}", blame_oid);
+    }
+
+    if let Some(mut helper) = hg_helper {
+        helper.kill().unwrap();
     }
 }
--- a/tools/src/bin/build-blame.rs
+++ b/tools/src/bin/build-blame.rs
@@ -286,7 +286,7 @@ fn main() {
     let args: Vec<_> = env::args().collect();
     let git_repo = Repository::open(&args[1]).unwrap();
     let blame_repo = Repository::open(&args[2]).unwrap();
-    let use_cinnabar = env::var("CINNABAR").map(|v| v == "1").unwrap_or(false);
+    let use_cinnabar = env::var("CINNABAR").map_or(true, |v| v != "0");
     let mut hg_helper = if use_cinnabar {
         Some(start_cinnabar_helper(&git_repo))
     } else {
--- a/tools/src/bin/build-blame.rs
+++ b/tools/src/bin/build-blame.rs
@@ -1,0 +1,408 @@
+extern crate env_logger;
+extern crate git2;
+#[macro_use]
+extern crate log;
+extern crate tools;
+extern crate unicode_normalization;
+
+use std::borrow::Borrow;
+use std::collections::HashMap;
+use std::env;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+use git2::{DiffFindOptions, ObjectType, Oid, Patch, Repository, Sort};
+use tools::config::index_blame;
+use tools::git_ops;
+use unicode_normalization::UnicodeNormalization;
+
+fn get_hg_rev(git_repo: &Repository, git_oid: &Oid) -> Option<String> {
+    // TODO: start a helper process and stream I/O to it instead of spawning a new process every time
+    let output = Command::new("git")
+        .arg("cinnabar")
+        .arg("git2hg")
+        .arg(format!("{}", git_oid))
+        .current_dir(git_repo.path())
+        .output()
+        .unwrap();
+    if !output.status.success() {
+        panic!("`git cinnabar git2hg {}` returned error code {:?}", git_oid, output.status.code());
+    }
+    match git_ops::decode_bytes(output.stdout).lines().next() {
+        Some(r) if r != "0000000000000000000000000000000000000000" => Some(r.to_string()),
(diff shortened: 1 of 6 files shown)


tools/src/bin/build-blame.rs
```diff
@@ -0,0 +1,357 @@
...
+        .arg("git2hg")
+        .arg(format!("{}", git_oid))
+        .current_dir(git_repo.path())
+        .output()
+        .unwrap();
+    if !output.status.success() {
+        panic!("`git cinnabar git2hg {}` returned error code {:?}", git_oid, output.status.code());
+    }
+    match git_ops::decode_bytes(output.stdout).lines().next() {
+        Some(r) if r != "0000000000000000000000000000000000000000" => Some(r.to_string()),
```
emilio: May be worth to use `.as_bytes().iter().all()` to future-proof this if the hashes get longer, but can be a follow-up.

tools/src/bin/build-blame.rs
```diff
@@ -0,0 +1,357 @@
...
+        .arg("cinnabar")
+        .arg("git2hg")
+        .arg(format!("{}", git_oid))
+        .current_dir(git_repo.path())
+        .output()
+        .unwrap();
+    if !output.status.success() {
+        panic!("`git cinnabar git2hg {}` returned error code {:?}", git_oid, output.status.code());
+    }
+    match git_ops::decode_bytes(output.stdout).lines().next() {
```
emilio: You can:

```
let line = git_ops::decode_bytes(output.stdout).lines().next()?;
if line != ... {
    return None;
}
Some(line.to_string())
```

tools/src/bin/build-blame.rs
```diff
@@ -0,0 +1,357 @@
...
+        Some(r) if r != "0000000000000000000000000000000000000000" => Some(r.to_string()),
+        _ => None,
+    }
+}
+
+fn count_lines(blob: &git2::Blob) -> usize {
+    let data = blob.content();
+    if data.len() == 0 {
```
emilio, line 46: `is_empty`?

tools/src/bin/build-blame.rs
```diff
@@ -0,0 +1,357 @@
...
+        return 0;
+    }
+    let mut linecount = 0;
+    for b in data {
+        if *b == b'\n' {
+            linecount += 1;
+        }
+    }
+    if data[data.len() - 1] != b'\n' {
```
emilio, line 55: Hmm, this is a bit subtle but I guess it's what `count(splitlines())` did?

tools/src/bin/build-blame.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    if patch.delta().flags().is_binary() {
+        return Ok(unchanged);
+    }
+
+    fn add_delta(lineno: usize, delta: i32) -> usize {
+        ((lineno as i32) + delta) as usize
+    }
+
+    let mut latest_line : usize = 0;
```
emilio, line 83: nit: Remove the space to the left of `:`? That seems more common, and it is consistent with arguments.

tools/src/bin/build-blame.rs
```diff
@@ -0,0 +1,357 @@
...
+    blame_repo: &git2::Repository,
+    blame_parents: &[git2::Commit],
+    path: &Path
+) -> Result<String, git2::Error> {
+    let blob = commit.tree()?.get_path(path)?.to_object(git_repo)?.peel_to_blob()?;
+    let linecount = count_lines(&blob);
+    // TODO: drop this author field entirely, I don't think it gets consumed by anything
+    let commit_id = commit.id();
+    let author = commit.author().name().map(|n| n.nfkd().filter(|c| c.is_ascii()).collect::<String>()).unwrap_or_else(String::new);
+    let mut blame = Vec::new();
```
emilio, line 130: nit: Maybe `Vec::with_capacity(linecount)`.

tools/src/bin/build-blame.rs
```diff
@@ -0,0 +1,357 @@
...
+    commit: &git2::Commit,
+    blame_repo: &git2::Repository,
+    blame_parents: &[git2::Commit],
+    path: &Path
+) -> Result<String, git2::Error> {
+    let blob = commit.tree()?.get_path(path)?.to_object(git_repo)?.peel_to_blob()?;
+    let linecount = count_lines(&blob);
+    // TODO: drop this author field entirely, I don't think it gets consumed by anything
+    let commit_id = commit.id();
+    let author = commit.author().name().map(|n| n.nfkd().filter(|c| c.is_ascii()).collect::<String>()).unwrap_or_else(String::new);
```
emilio, line 129: nit: `unwrap_or_defautl()`.

tools/src/bin/build-blame.rs
```diff
@@ -0,0 +1,357 @@
...
+    for line in 1..=linecount {
+        blame.push(format!("{}:%:{}:{}", commit_id, line, author));
+    }
+
+    for (i, parent) in commit.parents().enumerate().rev() {
+        let parent_path = file_movement.get(&parent.id())
+            .and_then(|m| m.get(&blob.id()))
+            .map(|p| p.borrow())
+            .unwrap_or(path);
+        let parent_blob = match parent.tree()?.get_path(parent_path) {
```
emilio, line 140: nit: This may be a bit cleaner early-continuing, but your call:

```rust
let parent = match parent.tree()?.get_path(parent_path) {
    Ok(p) => p,
    Err(..) => continue,
};

if !parent.kind().map_or(false, |t| t == ObjectType::Blob) {
    continue;
}

let parent = parent.to_object(git_repo)?.peel_to_blob()?;
```

Or such.

tools/src/bin/build-blame.rs
```diff
@@ -0,0 +1,357 @@
...
+            Ok(t) if t.kind() == Some(ObjectType::Blob) => {
+                t.to_object(git_repo)?.peel_to_blob()?
+            }
+            _ => {
+                continue;
+            }
+        };
+        let parent_blame_blob = match blame_parents[i]
```
emilio, line 148: Maybe you could change this loop to be:

```rust
for (parent, parent_blame) in commit.parents().rev().zip(blame_parents.iter()) {
```

Or such, after asserting they're the same length or something. No big deal either way, fine as a followup.

tools/src/bin/build-blame.rs
```diff
@@ -0,0 +1,357 @@
...
+            .tree()?
+            .get_path(parent_path) {
+            Ok(entry) => entry.to_object(blame_repo)?.peel_to_blob()?,
+            _ => continue,
+        };
+        let parent_blame = std::str::from_utf8(parent_blame_blob.content())
```
emilio, line 154: What guarantees that the unwrap() here is sound? Is it just that we always put ascii in the blame repo? If so, we should add a comment. It seems we could avoid relying on that without too much trouble anyhow... Maybe this function should return `Box<[u8]>`?

tools/src/bin/build-blame.rs
```diff
@@ -0,0 +1,357 @@
...
+    Ok(())
+}
+
+fn main() {
+    env_logger::init();
+
+    let args: Vec<_> = env::args().collect();
+    let git_repo = Repository::open(&args[1]).unwrap();
+    let blame_repo = Repository::open(&args[2]).unwrap();
+    let use_cinnabar = env::var("CINNABAR").map(|v| v == "1").unwrap_or(false);
```
emilio: nit: `map_or(false, ...`

tools/src/bin/build-blame.rs
```diff
@@ -0,0 +1,357 @@
...
+    let blame_ref = env::var("BLAME_REF").ok().unwrap_or("HEAD".to_string());
```
emilio, line 264: nit: `unwrap_or_else`? Though arguably not in hot code enough to matter at all.

tools/src/bin/build-blame.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    let mut refobj = blame_repo.find_reference(&blame_ref)
+        .ok()
+        .and_then(|r| r.resolve().ok());
+
+    let mut walk = git_repo.revwalk().unwrap();
+    walk.set_sorting(Sort::TOPOLOGICAL | Sort::REVERSE).unwrap();
+    walk.push(git_repo.refname_to_id(&blame_ref).unwrap()).unwrap();
+    let revs_to_process = walk
+        .map(|r| r.unwrap()) // walk produces Result<git2::Oid> so we unwrap to just the Oid
```
emilio, line 282: You could also filter_map or such.

tools/src/bin/build-blame.rs
```diff
@@ -0,0 +1,357 @@
...
+            let mut movement = HashMap::new();
+            let mut diff = git_repo.diff_tree_to_tree(
+                Some(&commit.parent(0).unwrap().tree().unwrap()),
+                Some(&commit.tree().unwrap()),
+                None
+            ).unwrap();
+            diff.find_similar(Some(DiffFindOptions::new().copies(true).rename_limit(1000000))).unwrap();
+            for delta in diff.deltas() {
+                if !delta.old_file().id().is_zero() && !delta.new_file().id().is_zero() && delta.old_file().path() != delta.new_file().path() {
+                    movement.insert(delta.new_file().id(),
```
emilio, line 315: nit: Indentation here looks a bit odd, maybe rustfmt?

tools/src/bin/build-blame.rs
```diff
@@ -0,0 +1,373 @@
...
+
+fn get_hg_rev(helper: &mut Child, git_oid: &Oid) -> Option<String> {
+    write!(helper.stdin.as_mut().unwrap(), "git2hg {}\n", git_oid).unwrap();
+    let mut reader = BufReader::new(helper.stdout.as_mut().unwrap());
+    loop {
+        let mut result = String::new();
+        reader.read_line(&mut result).unwrap();
+        if result.starts_with("changeset ") {
+            return result.split(' ').nth(1).map(str::trim).map(str::to_string);
+        } else if result.trim() == "0000000000000000000000000000000000000000" {
```
emilio, line 27: nit: No else after return.

tools/src/bin/build-blame.rs
staktrace: Done, except I used `chars().all()`

tools/src/bin/build-blame.rs
staktrace: This code goes away in the next patch so I ignored this.

tools/src/bin/build-blame.rs
staktrace: Done

tools/src/bin/build-blame.rs
staktrace: Yeah, it took me a couple of tries to get it Just Right.

tools/src/bin/build-blame.rs
staktrace: Fixed

tools/src/bin/build-blame.rs
staktrace: Good idea, fixed

tools/src/bin/build-blame.rs
staktrace: Fixed

tools/src/bin/build-blame.rs
staktrace: Good idea, done. (Although the `rev()` goes after the `zip()`)

tools/src/bin/build-blame.rs
staktrace: Yeah we only put ascii in the blame repo. I added a comment. In the future I plan to change drop the ascii-ized author name and start putting other stuff in (that may or may not be ascii) so I'll fix this then.

tools/src/bin/build-blame.rs
staktrace: Done

tools/src/bin/build-blame.rs
staktrace: Yeah this just runs once. Left it as-is.

tools/src/bin/build-blame.rs
staktrace: This I left as-is because using `filter_map` would be cleanest if I did something like `r.ok().and_then(...)` but I don't really want to lose the `unwrap()` here and cause silent dropping of revs in case `libgit2` produces an err for whatever reason. Also it seems easier to read this way where the concerns are separated, although it might be less performant.

tools/src/bin/build-blame.rs
staktrace: Ah yeah, ran rustfmt on the whole thing now.

tools/src/bin/build-blame.rs
staktrace: Fixed

PR #313, opened 2020-04-29
Bug 1634100 - Fix loop to avoid picking up non-word char prior to word

staktrace:


staktrace:
Reporter diagnosed the problem on the bug, makes sense.

PR #314, opened 2020-04-30
Bug 1633260  - Add Java AST parser using JavaParser

makotokato:
Java compiler doesn't seems to have compiler plug-in, but we can parse Java source code using https://github.com/javaparser/javaparser.
    
About cross references over another source files, it has limitations. Because it spends a lot of times and memory (most situations throws out of memory exception), so it references files into same directory files only.
    
- class define.
- method define.
- member variable define.
- variable access.
- member variable access.
- method call.

### DIFF
--- a/java/README.md
+++ b/java/README.md
@@ -1,0 +1,19 @@
+# Generate JSON cross reference files for mozsearch
+
+## Usage
+```
+java -jar mozsearch-java-all.jar <source code path> <output JSON path>
+```
+
+## Build with bazel
+```
+bazel build //:JavaAnalyze
+```
+
+## Run test with bazel
+```
+bazel test //:IndexingTest
+```
+
+## Additional Information
+If `ANDROID_SDK_ROOT` is set, we use android's jar file to resolve symbols (but it spends a lot of memory).
--- a/java/WORKSPACE
+++ b/java/WORKSPACE
@@ -1,0 +1,26 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+RULES_JVM_EXTERNAL_TAG = "3.3"
+RULES_JVM_EXTERNAL_SHA = "d85951a92c0908c80bd8551002d66cb23c3434409c814179c0ff026b53544dab"
+
+http_archive(
+    name = "rules_jvm_external",
+    strip_prefix = "rules_jvm_external-%s" % RULES_JVM_EXTERNAL_TAG,
+    sha256 = RULES_JVM_EXTERNAL_SHA,
+    url = "https://github.com/bazelbuild/rules_jvm_external/archive/%s.zip" % RULES_JVM_EXTERNAL_TAG,
+)
+
+load("@rules_jvm_external//:defs.bzl", "maven_install")
+
+maven_install(
+    artifacts = [
+        "com.github.javaparser:javaparser-core:3.23.1",
+        "com.github.javaparser:javaparser-symbol-solver-core:3.23.1",
+        "org.json:json:20190722",
+        "junit:junit:3.8.1",
+    ],
+    repositories = [
+        "https://jcenter.bintray.com/",
+        "https://repo1.maven.org/maven2",
+    ],
+)
--- a/scripts/java-analyze.sh
+++ b/scripts/java-analyze.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -x # Show commands
+set -eu # Errors/undefined vars are fatal
+set -o pipefail # Check all commands in a pipeline
+
+if [ $# -ne 2 ]
+then
+    echo "Usage: java-analyze.sh config-file.json tree_name"
+    exit 1
+fi
+
+CONFIG_FILE=$(realpath $1)
+TREE_NAME=$2
+
+cd $MOZSEARCH_PATH/java
+bazel run //:JavaAnalyze -- $FILES_ROOT $INDEX_ROOT/analysis
+cd -
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ env
 repos
 /index
 addons
+/java/bazel-*
+/java/build
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ help:
 
 .DEFAULT_GOAL := help
 
-.PHONY: help check-in-vagrant build-clang-plugin build-rust-tools test-rust-tools build-test-repo build-mozilla-repo baseline comparison
+.PHONY: help check-in-vagrant build-clang-plugin build-rust-tools test-rust-tools build-java-tools build-test-repo build-mozilla-repo baseline comparison
 
 check-in-vagrant:
 	@[ -d /vagrant ] || (echo "This command must be run inside the vagrant instance" > /dev/stderr; exit 1)
@@ -23,7 +23,10 @@ build-rust-tools:
 test-rust-tools:
 	cd tools && rustup run nightly cargo test --release --verbose
 
-build-test-repo: check-in-vagrant build-clang-plugin build-rust-tools
+build-java-tools:
+	cd java && bazel build //:JavaAnalyze
+
+build-test-repo: check-in-vagrant build-clang-plugin build-rust-tools build-java-tools
 	mkdir -p ~/index
 	/vagrant/infrastructure/indexer-setup.sh /vagrant/tests config.json ~/index
 	/vagrant/infrastructure/indexer-run.sh /vagrant/tests ~/index
@@ -54,14 +57,14 @@ review-test-repo:
 	INSTA_FORCE_PASS=1 /vagrant/infrastructure/web-server-check.sh /vagrant/tests ~/index "http://localhost/"
 	cargo insta review --workspace-root=/vagrant/tests/
 
-build-searchfox-repo: check-in-vagrant build-clang-plugin build-rust-tools
+build-searchfox-repo: check-in-vagrant build-clang-plugin build-rust-tools build-java-tools
 	mkdir -p ~/searchfox-index
 	/vagrant/infrastructure/indexer-setup.sh /vagrant/tests searchfox-config.json ~/searchfox-index
 	/vagrant/infrastructure/indexer-run.sh /vagrant/tests ~/searchfox-index
 	/vagrant/infrastructure/web-server-setup.sh /vagrant/tests searchfox-config.json ~/searchfox-index ~
 	/vagrant/infrastructure/web-server-run.sh /vagrant/tests ~/searchfox-index ~
 
-build-mozilla-repo: check-in-vagrant build-clang-plugin build-rust-tools
+build-mozilla-repo: check-in-vagrant build-clang-plugin build-rust-tools build-java-tools
 	[ -d ~/mozilla-config ] || git clone https://github.com/mozsearch/mozsearch-mozilla ~/mozilla-config
 	mkdir -p ~/mozilla-index
 	/vagrant/infrastructure/indexer-setup.sh ~/mozilla-config config1.json ~/mozilla-index
@@ -71,7 +74,7 @@ build-mozilla-repo: check-in-vagrant build-clang-plugin build-rust-tools
 
 # This is similar to build-mozilla-repo, except it strips out the non-mozilla-central trees
 # from config.json and puts the stripped version into trypush.json.
-trypush: check-in-vagrant build-clang-plugin build-rust-tools
+trypush: check-in-vagrant build-clang-plugin build-rust-tools build-java-tools
 	[ -d ~/mozilla-config ] || git clone https://github.com/mozsearch/mozsearch-mozilla ~/mozilla-config
 	jq '{mozsearch_path, default_tree, trees: {"mozilla-central": .trees["mozilla-central"]}}' ~/mozilla-config/config1.json > ~/mozilla-config/trypush.json
 	mkdir -p ~/trypush-index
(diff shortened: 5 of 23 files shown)


staktrace:
I tried running this on an indexer, and got an error:

```
Generating references ...
java.nio.file.FileSystemException: /mnt/index-scratch/mozilla-central/gecko-dev/third_party/rust/block: Too many open files
```

So that will need fixing. Also it didn't actually exit with an error code, so the failure was "silent". That can be easily fixed by catching any exception thrown and doing a System.exit(1) or similar. I'll spend some more time poking around later this week.

staktrace:
I'm wondering if the class files have enough information to generate the analysis data that searchfox needs. If so that would be easier to process since you just need to load the one file at a time, and it can be parallelized. Much lower performance footprint in general. Although we'd have to scrape class files out of the m-c build, and I don't know how it would work for mozilla-mobile stuff. So maybe that's not such a great idea after all.

makotokato:
> So that will need fixing. Also it didn't actually exit with an error code, so the failure was "silent". That can be easily fixed by catching any exception thrown and doing a System.exit(1) or similar. I'll spend some more time poking around later this week.

Thanks, I am also ooking this.

> I'm wondering if the class files have enough information to generate the analysis data that searchfox needs.

Yes. This parser spends a lot of times to resolve class reference, but I canno find AST generator from class file with -g argument.

makotokato:
I updated a lot of changes.

- Build system is changed to Bazel.
- Don't traverse symlink since symlink may be nested directory.
- Improve resolving type and performance
- etc

asutherland:
This is very exciting progress!

I've invited you to the "searchfox-aws" mozillians group on people.mozilla.org that is the basis for access to the searchfox AWS setup so that you can trigger test runs.  More details on that at https://github.com/mozsearch/mozsearch/blob/master/docs/aws.md.

I've also created a "java" channel using channel-tool which means that if indexer runs succeed, they'll show up at https://java.searchfox.org/.  I triggered the run I just initiated via the following (which the doc also roughly explains):

```shell
infrastructure/aws/trigger_indexer.py \
  https://github.com/makotokato/mozsearch \
  https://github.com/makotokato/mozsearch-mozilla \
  config1.json \
  java \
  java
```

I see you have a "mobile-only" branch at https://github.com/makotokato/mozsearch-mozilla/tree/mobile-only but the underlying script infrastructure assumes that any branches will have the same name in mozsearch and mozsearch-mozilla.

In the event there's an indexing failure, I think it will email you because the heuristics for choosing a recipient are based on who authored the commit (although now that we have LDAP/individual accounts, maybe we could do better than that).

In terms of additional things that I think are probably desirable / good:
- We'd want the [provisioning scripts](https://github.com/mozsearch/mozsearch/blob/master/docs/aws.md#provisioning-and-cloud-init) and [update scripts](https://github.com/mozsearch/mozsearch/blob/master/docs/aws.md#updating-the-machine-after-startup) to trigger the java indexer build, ideally in an idempotent fashion.  The provisioning scripts do trigger the update scripts (twice; the second time is to test that running the script again when the EC2 instance starts up won't obviously fail), so I think just updating [indexer-update.sh](https://github.com/mozsearch/mozsearch/blob/master/infrastructure/indexer-update.sh) is probably sufficient.
- There's been a new [testing infrastructure](https://github.com/mozsearch/mozsearch/blob/master/docs/testing-checks.md) added since last year, and it would be great to have some coverage in the "tests" repo that exercises some of the Java functionality; in particular to have coverage for situations where symbol names look the same between different languages. 
  - For example, if we expect JNI calls between different languages to work (and that informs our symbol mangling), it would be great to use the "query" command (ex [input](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/inputs/query__big_cpp__abstractart__json) [output](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/snapshots/check_glob%40query__big_cpp__abstractart_beart__json.snap)) which runs against the web-server to provide results from the `crossref` database.
  - That said, the existing whole-file expectation tests in this PR are great and there's no need to duplicate them into the "tests" repo or the check mechanism.  I think the main goal for coverage for java would be:
    - Make sure we have coverage so that regressions in the processing of the Java analysis files through the later stages of the pipeline can be detected.
    - Make sure there is some Java source that can be clicked around when viewing the "tests" repo locally.  (And it's fine to use the existing test files you created; we just don't need to add "check" coverage for every edge case since that's already covered by the in-Java tests.)
- Along those lines, it would be great to update https://github.com/mozsearch/mozsearch/blob/master/docs/analysis.md to explain our symbol mapping for Java, in particular with regards to potential collisions between similarly named C++, JS, rust, and Java classes.
  - While I don't think we have any real commitments to symbol stability (other than using C++ symbols verbatim), it would be nice to avoid churn on this, so it would be great to make sure we've documented the expected ramifications, etc.

java/README.md
```diff
@@ -0,0 +1,19 @@
+# Generate JSON cross reference files for mozsearch
```
asutherland, line 1: nit: I'd call these analysis files.  While it sounds like the underlying mechanism does have to do some cross-referencing of its own, I think the mozsearch terminology is basically that indexers produce analysis files.  (And then `crossref.rs` performs the cross-referencing.)

scripts/java-analyze.sh
```diff
@@ -0,0 +1,18 @@
...
+    echo "Usage: java-analyze.sh config-file.json tree_name"
+    exit 1
+fi
+
+CONFIG_FILE=$(realpath $1)
+TREE_NAME=$2
+
+cd $MOZSEARCH_PATH/java
+bazel run //:JavaAnalyze -- $FILES_ROOT $INDEX_ROOT/analysis
+cd -
```
asutherland, line 18: It'd be great to have a comment here that details or cross-references other documentation (ex: the `java/README.md) that explains the expected parallelism here and any limitations on this.  For example, https://github.com/mozsearch/mozsearch/blob/master/scripts/js-analyze.sh which drives in the similarly happens-during-searchfox-indexing JS analysis stage uses `parallel` which makes it clear how we're achieving parallelism.

If we aren't able to achieve parallelism in this stage, this may be something where we need to move it into the per-tree searchfox taskcluster jobs which currently have a lot of slack timing-wise.  That is, right now every additional minute that this step takes is an additional minute of latency for searchfox, whereas we literally can add hours of additional processing into the taskcluster jobs and it won't affect latency because there's already a large wait there for the code coverage data.

asutherland:
It seems like the indexing run went great and now things like https://java.searchfox.org/mozilla-central/search?q=symbol:org%23mozilla%23geckoview%23GeckoSession&redirect=false work!  Woo!

Even better, it looks like the overhead of running java analyze was fairly trivial per a lightly modified `indexer-logs-analyze.sh` so it recognizes `java-analyze.sh`:
```
├── mozilla-central                                                                                                                                                                           
│   └──                                                                                                                                                                                       
│         script                time since start   apparent duration                                                                                                                          
│        ────────────────────────────────────────────────────────────                                                                                                                         
│         find-repo-files.py    0:00:00            0:01:18                                                                                                                                    
│         build.sh              0:01:18            0:08:29                                                                                                                                    
│         js-analyze.sh         0:09:47            0:03:01                                                                                                                                    
│         idl-analyze.sh        0:12:48            0:00:10                                                                                                                                    
│         ipdl-analyze.sh       0:12:58            0:00:01                                                                                                                                    
│         java-analyze.sh       0:12:59            0:02:15                                                                                                                                    
│         crossref.sh           0:15:14            0:08:13                                                                                                                                    
│         output.sh             0:23:27            0:33:11                                                                                                                                    
│         build-codesearch.py   0:56:38            0:05:47                                                                                                                                    
│         check-index.sh        1:02:25                                                                                                                                                       
│                                                                                                                                                                                             
├── comm-central                                                                                                                                                                              
│   └──                                                                                                                                                                                       
│         script                time since start   apparent duration                                                                                                                          
│        ────────────────────────────────────────────────────────────                                                                                                                         
│         find-repo-files.py    0:00:00            0:01:23                                                                                                                                    
│         build.sh              0:01:23            0:00:00                                                                                                                                    
│         js-analyze.sh         0:01:23            0:03:27                                                                                                                                    
│         idl-analyze.sh        0:04:50            0:00:08                                                                                                                                    
│         ipdl-analyze.sh       0:04:58            0:00:00                                                                                                                                    
│         java-analyze.sh       0:04:58            0:02:27                                                                                                                                    
│         crossref.sh           0:07:25            0:01:57                                                                                                                                    
│         output.sh             0:09:22            0:16:49                                                                                                                                    
│         build-codesearch.py   0:26:11            0:04:36                                                                                                                                    
│         check-index.sh        0:30:47                                                                                                                                                       
│                                                                                                                                                                                             
├── mozilla-mobile                                                                                                                                                                            
│   └──                                                                                                                                                                                       
│         script                time since start   apparent duration                           
│        ────────────────────────────────────────────────────────────                                                                                                                         
│         find-repo-files.py    0:00:00            0:00:38                                     
│         build.sh              0:00:38            0:00:00                                     
│         js-analyze.sh         0:00:38            0:00:46                                     
│         idl-analyze.sh        0:01:24            0:00:00                                     
│         ipdl-analyze.sh       0:01:24            0:00:00                                     
│         java-analyze.sh       0:01:24            0:01:45                                     
│         crossref.sh           0:03:09            0:00:20                                     
│         output.sh             0:03:29            0:01:50                                     
│         build-codesearch.py   0:05:19            0:00:59                                     
│         check-index.sh        0:06:18                                                        
│                                                              
```

scripts/java-analyze.sh
asutherland: As an addendum to my prior comment, it looks like whatever is happening is more than sufficiently performant for our purposes at this time, but it would be great to have this info available here/somewhere.

asutherland (changes requested):
I don't want to leave this PR in an ambiguous state, so I'm marking this as requested changes, but the set of must-have changes is basically the first 2 bullets from https://github.com/mozsearch/mozsearch/pull/314#issuecomment-1003433573 which is mainly 1) we need to hook this up to provisioning (fairly easy), 2) we should have some "tests" repo coverage (easy-medium, depending on build step complexity).  A nice-to-have is the third bullet of more information about the analysis process but that can always be reverse engineered later on.

@makotokato I'm not sure about your availability to do the above set of changes on this.  Both of the must-haves are things I can do fairly easily if it helps get this invaluable functionality landed, but especially if you'd like to be an ongoing searchfox contributor, they're both pretty useful things to gain experience doing (but I am still happy to do if you're not sure when you'd have the time/don't have the time right now.)

scripts/java-analyze.sh
makotokato: Java parser runs on multi-thread and resolving "type" uses multi-thread.  Even if using `parallel`, it doesn't become faster.

When using parallel, 168 secs on my workstation (Ryzen 9 3900), but when not using it, 75 secs.

makotokato:
> For example, if we expect JNI calls between different languages to work

I forget to explain about JNI. GeckoView generates JNI binding to using JVM API (`RegisterNatives`), not using `export`/JNICALL functions. So we cannot resolve/map C++ function from AST parser only.

Also, `mozglue.so` uses [export-style JNICALL method](https://searchfox.org/mozilla-central/rev/fbf1e796ecead9484deced4d99f199f327ba25ab/mozglue/android/APKOpen.cpp#276). So the binding isn't simple in GeckoView.

asutherland:
Thanks for the really useful context about the JNI calls!  That definitely sounds like it would want to happen at a higher level of cross-referencing, like the semantic linkage mechanism I've brainstormed at https://bugzilla.mozilla.org/show_bug.cgi?id=1727789 (and where I added [a comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1727789#c2) to try and quickly capture the results of my superficial investigation of what you referenced).

scripts/java-analyze.sh
asutherland: Excellent!  Thank you for the clarification.  That definitely explains how it's so fast! :)

staktrace:
This is pretty awesome, thank you!

I wanted to mention that since I wrote this comment:

> I'm wondering if the class files have enough information to generate the analysis data that searchfox needs.

I was kinda working on that approach on the back burner in the form of writing a [class file parser](https://crates.io/crates/cafebabe) but after doing that, discovered that class files usually don't have enough information to provide satisfactory analysis information. It would need to be augmented with source AST information anyway, so doing it fully using the source AST as you've done seems better.

java/WORKSPACE
```diff
@@ -0,0 +1,26 @@
...
+
+maven_install(
+    artifacts = [
+        "com.github.javaparser:javaparser-core:3.23.1",
+        "com.github.javaparser:javaparser-symbol-solver-core:3.23.1",
+        "org.json:json:20190722",
+        "junit:junit:3.8.1",
+    ],
+    repositories = [
+        "https://jcenter.bintray.com/",
```
staktrace, line 23: AFAIK bintray has been sunsetted, so you can remove this entry. https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

asutherland:
> I was kinda working on that approach on the back burner in the form of writing a [class file parser](https://crates.io/crates/cafebabe) but after doing that, discovered that class files usually don't have enough information to provide satisfactory analysis information. It would need to be augmented with source AST information anyway, so doing it fully using the source AST as you've done seems better.

Sourcegraph recently rolled out Java LSIF indexing and the description at https://sourcegraph.github.io/lsif-java/docs/design.html says they use a compiler plugin (which then produces SemanticDB data as an intermediary).  I haven't dug into using LSIF meaningfully, but I have filed https://bugzilla.mozilla.org/show_bug.cgi?id=1740290 to track the idea of doing so.  Longer term it might be possible for us to use something like that as well, especially because I think it supports Kotlin (which we use for a lot of mozilla-mobile stuff) and Scala (which I don't think weuse at all).

asutherland:
Also, the ["cafebabe" crate](https://crates.io/crates/cafebabe) seems very cool!

asutherland:
Closing this since we went with the SCIP based approach in https://bugzilla.mozilla.org/show_bug.cgi?id=1490144 and I believe we have consensus in https://bugzilla.mozilla.org/show_bug.cgi?id=1490144#c14 that although this was a very impressive implementation, that the SCIP/semanticdb approach ended up being most pragmatic/suitable for our needs (and given our need to support SCIP for rust-analyzer).

PR #315, opened 2020-04-30
Bug 1634367 - Fix a typo that causes the path value to be written to the query input.

emilio:


emilio:
Kats stamped this on riot.

PR #317, opened 2020-05-01
Remove python blame builder

staktrace:
The rust version is deployed now, we don't need the python version any more. I haven't tested with a reprovisioned indexer yet but I'll do that before I merge this PR. I have some other provisioning changes I want to test as well.


staktrace:
Re-requesting review as I made a few changes:
- Also deleted the pygit2 stuff from the web-server provisioning script
- Added an extra patch to update documentation and a leftover reference to transform-repo.py that I missed.

staktrace:
I also did a test run with a reprovisioned indexer and verified it was fine.

staktrace:
Thanks! :)

PR #320, opened 2020-05-04
Bug 1634955 - Bump cinnabar to fix for glandium/git-cinnabar#252

staktrace:


staktrace:
I'm going to go ahead and merge this without review since it's trivial, and picks up the fix we need for the indexer to not break tomorrow. I'm rebuilding the AMI now.

PR #325, opened 2020-05-08
Bug 1636392 - Update copying ply from gecko trees

staktrace:


staktrace:
I haven't tested this but I'm going to go ahead and merge it anyway so that it's in place for today's indexer run.

PR #326, opened 2020-05-08
Make idl-analyze.sh work with both old and new versions of xpidl.py

staktrace:
This is kinda hacky but should be ok for the forseeable future.

I'm running a test of this now.


staktrace:
This updated version seems to have worked.

PR #331, opened 2020-05-20
Bug 1639631 - Fix calls to subprocess.check_output.

staktrace:
`text=True` can only be used in python3.7 and up, but we're on 3.6. So we should
use the equivalent `universal_newlines=True` argument instead. Also, when catching
errors thrown from subprocess.check_output, let unexpected errors propagate
so we get a stack in the log and generally more obvious failure. Keep catching
the "expected" failure (non-zero return code from process).


staktrace:
Going to merge now so that the mozilla-releases web-server picks this up, as that will spin up shortly.

PR #336, opened 2020-05-26
Bug 1638664 - Update clang-plugin in searchfox repo. r=asuth

staktrace:


staktrace:
r=asuth from m-c landing.

PR #340, opened 2020-05-29
Bug 1641845 - js-analyze.js: Rename Symbol constructor to SymbolTable.Symbol.

ajvincent:
This has absolutely not been tested but it's a very small change.


staktrace:
Verified this doesn't break anything.

PR #344, opened 2020-06-01
Bug 1640915 - Report stray index volumes.

staktrace:
The only time there should be unattached index volumes are (a) immediately
after creation and before attaching to the indexer, and (b) while detaching
from the indexer and reattaching to the web-server. These are both very short
periods of time, so if we find stray index volumes after deleting old web-servers
that's likely a problem. Report these as warnings for further investigation.

### DIFF
--- a/infrastructure/aws/trigger-web-server.py
+++ b/infrastructure/aws/trigger-web-server.py
@@ -172,15 +172,22 @@ if len(terminate):
 for instanceId in terminate:
     awslib.await_instance(ec2, instanceId, None, 'terminated')
 
-# - Delete any old EBS index volumes
+# - Report any old EBS unattached index volumes.
+# Since they are marked as DeleteOnTermination we shouldn't need to delete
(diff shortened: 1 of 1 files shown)


staktrace:
The web-servers that were launched prior to #341 merging (i.e. the ones on the `mozilla-archived`, `asuth`, and `fancy` channels) all had their index volumes NOT set to DeleteOnTermination. So I manually updated those attachments, so that they have DeleteOnTermination=true. So that means spawning a new web-server on those channels after this patch merges should not leak any volumes or emit warnings.

staktrace:
> The only situation where we'd expect this to happen is for an abrupt termination of the indexer during this specific window where it's migrating its indexing volume to the web-server.

Or abrupt termination of the indexer immediately after it creates the volume (start of `attach-index-volume.py` but before it attaches to the volume and marks it DeleteOnTermination (end of `attach-index-volume.py`). But fundamentally similar in that it's a small window of time where and the indexer terminating would be exceptional behaviour.

PR #348, opened 2020-06-16
Bug 1646069 - normalize generated.rs' path. r=kats

asutherland:
Change the user-exposed path of generated.rs (which is created by
build.rs) from
`__GENERATED__/debug/build/files-7e56eb8bb8e7c9bb/out/generated.rs` to
`__GENERATED__/__BUILD_SCRIPT_OUTDIR__/generated.rs` by normalizing
post-`cargo build`.


asutherland:
I'm definitely open to suggestions for better names than `__BUILD_SCRIPT_OUTDIR__`.  Also about the whole approach here.  I think the UX of having a stable path name probably should be the dominant concern.

tests/tests/build
```diff
@@ -20,8 +20,38 @@ export CARGO_TARGET_DIR=$OBJDIR
...
+# "__BUILD_SCRIPT_OUTDIR__".
+BUILD_SCRIPT_ALIAS="__BUILD_SCRIPT_OUTDIR__"
+BUILD_SCRIPT_OUT_ALIASED_PATH="${OBJDIR}/${BUILD_SCRIPT_ALIAS}"
+ln -sf "$BUILD_SCRIPT_OUT_DIR" "$BUILD_SCRIPT_OUT_ALIASED_PATH"
+
+## Normalize the save-analysis paths.
+# Normalize the source root off to make source paths relative.
 NORMALIZE_EXPR='s#/vagrant/tests/tests/files/##g'
+# Normalize the dynamic build script output directory to our new alias.  The
+# OBJDIR prefix removal will be handed by the next normalization.
```
staktrace, line 50: typo: handled

asutherland:
> If we're saying searchfox should be normalizing the output dir, that's fine, but then are we going to do something like this for m-c as well? In particular note that we have generated dirs per-platform (see e.g. [this generated file](https://searchfox.org/mozilla-central/search?q=shaders.rs&path=&case=false&regexp=false)) whereas this patch is assuming a single unified dir. I don't know if this code is intended to be used for m-c at all and if this is a concern.

Yes, I posit that `__GENERATED__/x86_64-unknown-linux-gnu/debug/build/webrender-2a3586aec96fcf3d/out/shaders.rs` is a bad user-exposed path because it diverges from our `__linux64__`/`__win64__`/etc. scheme and is unstable and there are no mitigations other than linking to a search as you have done.  I think the best thing to do is for either the mozilla-central build process or the mozsearch-mozilla scripts that process the output to ensure that it's exposed at a stable URL.

We need not use the same scheme in both places, although it would probably be nice to try.

Right now we have `__GENERATED__/__RUST__` hosting all of the standard library sources which I guess must be part of the normal flat crate namespace so potentially could host other things.  It seems like the general options for platform-specific `__GENERATED__/x86_64-unknown-linux-gnu/debug/build/webrender-2a3586aec96fcf3d/out/shaders.rs` are:
1. `__GENERATED__/__linux64__/__RUST__/webrender/__BUILD_SCRIPT_OUTDIR__/shader.rs`
2. `__GENERATED__/__linux64__/__RUST__/__BUILD_SCRIPT_OUTDIR__/webrender/shader.rs`
3. `__GENERATED__/__linux64__/__BUILD_SCRIPT_OUTDIR__/webrender/shader.rs`

Note that we already expect the stdlib sources to end up under `__GENERATED__/__linux64__/__RUST__` if they vary by platform.  We could also change where the stdlib is exposed under `__RUST__`, such as at `__RUST__/__STDLIB__` or something.

@emilio I expect you have relevant thoughts too.

tests/tests/build
asutherland: Argh, thanks.  My keyboard.io keyswitches really need to be ultrasonically cleaned... thanks to a bad manufacturing process and imperfect software fixes they randomly drop some letters and double some other letters... it keeps me on my toes, but obviously I miss some things... also, sometimes it's my fault ;)

staktrace:
Thanks for clarifying, it all makes sense. I do think that we should keep the rust stdlib and crate-generated files separate, so assuming we keep the platform-specific stdlib files in `__GENERATED__/__linux64__/__RUST__` then I think option 3 of the ones you listed seems like the best one. Or if we're ok with renaming, we can flatten the folder structure a bit and do:

`__GENERATED__/__linux64__/__RUST_STDLIB__/...` for stdlib stuff, and
`__GENERATED__/__linux64__/__RUST_BUILD_SCRIPT__/webrender/shaders.rs` for crate-generated stuff.

asutherland:
Your renaming proposal of `__RUST_STDLIB__` and `__RUST_BUILD_SCRIPT__` sounds best to me as the extra specificity makes the paths self-explanatory in a way that `__RUST__` isn't on its own.

So in that case, the path for this patch becomes `__GENERATED__/__RUST_BUILD_SCRIPT__/files/generated.rs` (because "files" is the name of the crate, which I don't think I sufficiently appreciated before).

I'll take a quick pass at the mozilla-central logic to see whether helper scripts are needed or not.

PR #350, opened 2020-06-29
Redirect /repo-name to /repo-name/source

nhnt11:
Currently it's a 404. It'd be nice to just type searchfox.org/mozilla-central to browse the source. Maybe a nice follow-up would be to also redirect repo-name/path to repo-name/source/path but that's more complicated.

I didn't actually try this locally, but I think it should work? Maybe it needs to be moved below the other directives?

### DIFF
--- a/scripts/nginx-setup.py
+++ b/scripts/nginx-setup.py
@@ -168,6 +168,8 @@ for repo in config['trees']:
         'default_type text/html;',
         'add_header Cache-Control "must-revalidate";',
     ])
+    
+    location('/%(repo)s', ['return 301 $scheme://$host$request_uri/source;`])
 
     location('/%(repo)s/raw-analysis', [
         'root %(doc_root)s;',
--- a/scripts/nginx-setup.py
+++ b/scripts/nginx-setup.py
@@ -169,7 +169,7 @@ for repo in config['trees']:
         'add_header Cache-Control "must-revalidate";',
     ])
     
-    location('/%(repo)s', ['return 301 $scheme://$host$request_uri/source;`])
+    location('/%(repo)s', ['return 301 $scheme://$host$request_uri/source;'])
 
     location('/%(repo)s/raw-analysis', [
         'root %(doc_root)s;',


nhnt11:
I wonder, will this cover e.g. both `/mozilla-central` and `/mozilla-central/`?

staktrace:
So this patch by itself doesn't work for all repos. I tested it for mozilla-beta, and the URL `https://searchfox.org/mozilla-beta` still gives a 404, because the load balancer we have explicitly only routes URLs matching `/mozilla-beta/*` to the web-server with the beta code. i.e. the slash after the repo name is required for the load balancer to route it properly. In general I think is good practice if we ever end up with a repo name that's a prefix of another repo name. But it also means that for any repo not hosted on the primary web server it would violate this practice to have `/repo-name` automatically redirect to `/repo-name/source`.

The patch will work for requests to the URL `https://searchfox.org/mozilla-beta/` (with the trailing slash) but it redirects to `https://searchfox.org/mozilla-beta//source` (double slash) because you're plopping the entire request URI into the redirect. And also it creates an infinite redirect loop for URLs like `https://searchfox.org/mozilla-beta/?foo=bar` until the browser terminates it at a URL like `https://searchfox.org/mozilla-beta/?foo=bar/source/source/source/source/source/source/source/source/source/source`.

staktrace:
FYI there are instructions on how to set up a local test instance of searchfox in the README, so if you want to experiment with this locally to come up with a working patch please do that. It wouldn't have caught the load balancer problem but at least you can get it working for URLs with the trailing slash.

staktrace (changes requested):
Per comments on the PR

nhnt11:
Nice, so I do have to do this "properly", hehe. Thanks for the review! I'll try and find some time to set things up and build a more sophisticated solution later this week. Meanwhile if anyone is interested in picking this up please feel free.

asutherland:
As an aside, how are you ending up with URL's that look like `/repo-name`?  I could imagine starting with an autocomplete of `https://searchfox.org/mozilla-beta/source` and thinking that you don't want the source but you want a search page that has the same kind of "query language" details you would find at `https://searchfox.org/`, finding out the hard way there is no such thing exposed to the web.  (And I wonder if there should be such a thing.  Like should the "mozilla-beta" entry from `https://searchfox.org/` link to mozilla-beta's help.html that should get exposed now?)

nhnt11:
I have ended up trying to use searchfox, without thinking about it, as a code browser e.g. to check what something looks like in mozilla-beta or when I need to browse the source trying to find where something lives, or when my clone is old and I want to look at a more up-to-date version of the tree.

I do suppose I could fix this locally, but I filed this because 404 just didn't feel right.

asutherland:
All efforts to refine searchfox are appreciated!  I also use searchfox as my primary code browser.  I just wanted to make sure I understood the steps-to-reproduce as when the awesomebar/megabar/autocomplete enters the picture, typing a URL becomes much more complex/inconsistent between users.

Also, it's always nice to understand the use case and problem to be solved!

One thing this issues really makes obvious to me is that the searchfox 404 page is... very bad.  It seems like it could make sense to configure nginx to display the help.html page we see at https://searchfox.org/ for any 404 in the top-level directory.  Per-repo 404's would seem to be more complex as for paths under REPO/source I think we might want to fall back to a dynamic "CGI" revision control lookup to say "hey, that file last existed at HASH" or "nothing ever existed here".

I wonder if it would be better to just do the 404 thing for now to avoid creating autocomplete "muscle memory" for the /REPO and /REPO/ paths redirecting to /REPO/source.  At least in Firefox, I don't think the autocomplete mechanism gets feedback from redirects, which I think would mean that /REPO would get all the frecency, not its redirect target.  (My interest would be in retaining /REPO and /REPO/ as something future UI enhancements could use.)

asutherland:
Between our coverage support and https://tests.firefox.dev/ that we want to potentially help expose, plus some possibilities with helping with code review, I think we're pretty close to using the tree root pages for things, so I'm going to close this now.  Also, :arai's tree-switcher and Nicolas' letting us browse historical revision directories also reduces the need to manually mess with URLs.

PR #352, opened 2020-07-09
Bug 1616362 - Rework how the find-repo-files.py script works.

staktrace:
Instead of each tree having its own find-repo-files script (which are all
pretty much identical), this changes things around so that the main find-repo-files
script lives in the mozsearch scripts/ folder. This version contains the base
functionality, and allows individual tree configs to augment the behaviour
via specific hooks.

The base script (now at scripts/find-repo-files.py) is pretty much the same
as the old one, but upgraded to work with python3 (really just converting bytes
to strings in a few places).


staktrace:
Goes with https://github.com/mozsearch/mozsearch-mozilla/pull/104

PR #353, opened 2020-07-13
Provisioning updates

staktrace:


staktrace:
> The trade-off is one can't re-run the provision scripts to update an existing VM

Oh, so that's why the scripts were set up with `rm -rf`! I have never re-run the provision scripts (always destroy and rebuild the VM) so I didn't even realize that was a possibility. But yeah I agree that seems prone to breakage and easy to work around manually.

PR #354, opened 2020-07-16
Optimize the build-blame tool

staktrace:
This series of patches separates the work that build-blame does into "compute" and "writing to repo". The compute step is then parallelized across n-1 threads, leaving 1 thread for the "writing to repo". If n=2 the bottleneck is the compute side, but on the indexer we generally run with n=8 which makes the writing to repo part the bottleneck. So some additional patches then reduce the amount of time spent in the get_path function which was showing up as expensive in profiles. There's still more that can/should be done to optimize the writing thread, but I need to go do other things now.

Note that I initially parallelized using rayon, but it turns out that creating and destroying the `git2::Repository` objects is quite expensive (and they can't be shared across threads) so I wanted each worker thread to have a dedicated `git2::Repository` that it could use. I couldn't use rayon in this manner so I rolled my own simple round-robin threadpool thing which seems to work fine.

I've verified that the behaviour is still correct (produces hash-compatible blame repos with what we already have) and is much faster. How much faster depends on the repo, but usually on the order of 2-4x.

### DIFF
--- a/tools/src/bin/build-blame.rs
+++ b/tools/src/bin/build-blame.rs
@@ -480,6 +480,10 @@ fn main() {
         None
     };
     let blame_ref = env::var("BLAME_REF").ok().unwrap_or("HEAD".to_string());
+    let commit_limit = env::var("COMMIT_LIMIT")
+        .ok()
+        .and_then(|x| x.parse::<usize>().ok())
+        .unwrap_or(0);
 
     info!("Reading existing blame map of ref {}...", blame_ref);
     let mut blame_map = if let Ok(oid) = blame_repo.refname_to_id(&blame_ref) {
@@ -498,10 +502,18 @@ fn main() {
     walk.set_sorting(Sort::TOPOLOGICAL | Sort::REVERSE).unwrap();
     walk.push(git_repo.refname_to_id(&blame_ref).unwrap())
         .unwrap();
-    let revs_to_process = walk
+    let mut revs_to_process = walk
         .map(|r| r.unwrap()) // walk produces Result<git2::Oid> so we unwrap to just the Oid
         .filter(|git_oid| !blame_map.contains_key(git_oid))
         .collect::<Vec<_>>();
+    if commit_limit > 0 && commit_limit < revs_to_process.len() {
+        info!(
+            "Truncating list of commits from {} to specified limit {}",
+            revs_to_process.len(),
+            commit_limit
+        );
+        revs_to_process.truncate(commit_limit);
+    }
     let rev_count = revs_to_process.len();
 
     let num_threads: usize = num_cpus::get() - 1; // 1 for the main thread
--- a/tools/src/bin/build-blame.rs
+++ b/tools/src/bin/build-blame.rs
@@ -253,6 +253,7 @@ fn build_blame_tree(
     git_repo: &git2::Repository,
     commit: &git2::Commit,
     tree_at_path: &git2::Tree,
+    parent_trees: &[Option<git2::Tree>],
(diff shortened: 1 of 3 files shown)


tools/src/bin/build-blame.rs
```diff
@@ -114,7 +114,7 @@ fn unmodified_lines(
 }
 
 fn blame_for_path(
-    file_movement: &HashMap<Oid, HashMap<Oid, PathBuf>>,
+    file_movement: &Option<HashMap<Oid, PathBuf>>,
```
emilio: nit: Passing `Option<&HashMap<..>>` is a bit more idiomatic (you'd move the `as_ref()` to the caller.

tools/src/bin/build-blame.rs
```diff
@@ -162,27 +163,93 @@ fn blame_for_path(
...
+// modified in `commit`, relative to all the parents. The results are populated
+// into the `results` HashMap.
+fn find_unmodified_lines(
+    file_movement: &Option<HashMap<Oid, PathBuf>>,
+    git_repo: &git2::Repository,
+    commit: &git2::Commit,
+    mut path: PathBuf,
+    results: &mut HashMap<(git2::Oid, PathBuf), Vec<(usize, usize)>>,
+) -> Result<(), git2::Error> {
+    let tree_at_path = if path == PathBuf::new() {
```
emilio, line 197: A bit surprising there's no `is_empty()` in path or pathbuf tbf.

tools/src/bin/build-blame.rs
```diff
@@ -411,17 +414,70 @@ fn compute_diff_data(
...
+}
+
+fn compute_thread_main(
+    query_rx: Receiver<git2::Oid>,
+    response_tx: Sender<DiffData>,
+    git_repo_path: String,
+) {
+    let git_repo = Repository::open(git_repo_path).unwrap();
+    loop {
+        match query_rx.recv() {
```
emilio, line 471: `while let Ok(rev) = query_rx.recv() {`

tools/src/bin/build-blame.rs
```diff
@@ -306,12 +311,24 @@ fn build_blame_tree(
             }
             Some(ObjectType::Tree) => {
                 let mut entry_builder = blame_repo.treebuilder(None)?;
+                let mut parent_subtrees = Vec::new();
```
emilio, line 314: nit: Maybe `Vec::with_capacity(parent_trees.len());`

tools/src/bin/build-blame.rs
```diff
@@ -306,12 +311,24 @@ fn build_blame_tree(
...
+                for parent_tree in &parent_trees {
+                    let parent_subtree = match parent_tree {
```
emilio, line 316: May be worth a comment that we rely on the `parent_trees` vector to be the same length as blame_parents, for which we need to keep the `None`s in the vector?

tools/src/bin/build-blame.rs
```diff
@@ -254,18 +254,23 @@ fn build_blame_tree(
     git_repo: &git2::Repository,
     commit: &git2::Commit,
     tree_at_path: &git2::Tree,
+    parent_trees: Vec<Option<git2::Tree>>,
```
emilio, line 257: You don't seem to be moving or mutating it so maybe just `&[Option<git2::Tree>]`?

tools/src/bin/build-blame.rs
staktrace: This particular instance gets removed later but I updated the other one that remains.

PR #357, opened 2020-07-27
Bug 1653986 - Display information about disabled and failing tests on those tests' pages. r=kats

asutherland:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1653986#c8 for WPT-related examples and https://bugzilla.mozilla.org/show_bug.cgi?id=1653986#c4 for previous non-WPT examples.  https://asuth.searchfox.org/ should now have a fully happy config.json run now with full mozilla-central data.  The "dev" channel is currently at 5h48m on "mozilla-releases.json".

I did most of the things I said I would do in comment 8 but did not clean up the JSON any further.  There are likely places where `unwrap_or` could have been used instead of a match, or perhaps `into_object`.  In derive-file-info I was intentionally trying to consume the JSON as it was processed which meant as_foo() wasn't generally viable because they return borrows, but in output-file, I was trying to work with borrows.  In previous discussions with Nika she indicated we should probably be using serde for JSON instead of the rustc_serialize stuff, so I'd probably prefer deferring any deep idiomatic cleanups until we make that conversion.

Relatedly, I didn't run rustfmt/cargo format on this patch yet.  I was a little worried there may be underlying drift that would have complicated the patch.  I'll plan to run a total reformat in a second commit is part of this PR prior to landing if that works.  (Or more specifically, I'm fine with doing anything.)


tools/src/bin/derive-per-file-info.rs
```diff
@@ -0,0 +1,320 @@
...
+        all_info.insert("bugzilla-components".to_string(),
+                        data.remove("components").unwrap());
+
+        if let Some(bz_root) = data.get("paths") {
+            traverse_and_store_bugzilla_map(
+                bz_root.as_object().unwrap(),
+                all_info.get_mut("root").unwrap().as_object_mut().unwrap());
+        }
+    } else {
+        println!("No bugzilla-components.json file found");
```
staktrace, line 241: Should we make this a `warn!` instead so we get notified of problems? Ditto other missing files. And I guess the rest of the `println!` calls should really be `info!`s but that's a pre-existing problem so no need to fix that here.

tools/src/bin/output-file.rs
```diff
@@ -57,6 +54,110 @@ fn get_bugzilla_component<'a>(
...
+                    // element is a null or the condition string and the 2nd is
+                    // the bug link.
+                    match cond.as_array().unwrap_or(&vec![]).as_slice() {
+                        // null means there was no condition, it's always disabled.
+                        [Json::Null, Json::String(b)] => Some(("ALWAYS".to_string(), b.to_string())),
+                        [Json::String(a), Json::String(b)] => Some((a.to_string(), b.to_string())),
+                        // I guess this is just covering up our failures?  I'm
+                        // sorta tired of this patch though, so... cover up our
+                        // failures.
+                        _ => None,
```
staktrace, line 130: Covering up is fine, but maybe add a `warn!` of some sort here so that if it happens we'll get a notification about it.

tools/src/bin/output-file.rs
```diff
@@ -320,17 +427,184 @@ fn main() {
...
+                if !linked_meta {
+                    list_nodes.push(F::T(format!(
+                        r#"<li>This test has a <a href="{}">WPT meta file</a> for some reason."#,
+                        meta_url,
+                    )));
+                }
+            }
+
+            list_nodes.push(F::T(format!(
+                "<li>This test ran {} times in the preceding 7 days with an average run time of {:.2} secs.</li>",
```
staktrace, line 573: I noticed "inf secs" being reported on some pages, so maybe make the text after "preceding 7 days" conditional on the number of unskipped runs being > 0.

staktrace:
In terms of formatting, yeah doing a rustfmt as a separate commit on top of everything is fine by me.

tools/src/format.rs
```diff
@@ -323,6 +323,7 @@ pub fn format_file_data(
     cfg: &config::Config,
     tree_name: &str,
     panel: &[PanelSection],
+    info_boxes: Vec<InfoBox>,
```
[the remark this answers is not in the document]
asutherland: There was an ownership issue related to the `F`s that I failed to document.  I've pushed separate commit https://github.com/mozsearch/mozsearch/pull/357/commits/b961218b2f9e5ee0d594680ff997d7ad79cb6d89 that adds a clone mechanism, but I'm open to other fixes.  The only real opinion I have about the logic is it's nice to defer the formatting to as late as possible so it's nice for InfoBox to take a formatting object.  (Otherwise the indentation would be messed up unless we treat F to do appropriate replacements on newlines, but that's hacky/potentially breakage-prone for script tags with inline contents.)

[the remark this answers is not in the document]
staktrace: I guess if there were an enum variant similar to `Indent` but took a reference then we wouldn't need the clone. But yeah we can leave that for a future followup

tools/src/bin/output-file.rs
```diff
@@ -127,7 +129,10 @@ fn read_test_info_from_file_info(per_file_info: Option<&json::Object>) -> Option
                         // I guess this is just covering up our failures?  I'm
                         // sorta tired of this patch though, so... cover up our
                         // failures.
-                        _ => None,
+                        _ => {
+                            warn!("Unhandled disabled condition JSON branch! {:}", cond);
```
staktrace, line 133: typo in format string, i think

tools/src/format.rs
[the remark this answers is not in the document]
asutherland: Going to take you up on the deferral for now.  For future work, I think it's worth considering whether we could move to an existing third-party crate that might look more like React's JSX.  That is, a means of using macros to more directly embed something that looks like HTML and get some type-checking that the HTML is valid rather than a bunch of manual string juggling with format! calls throughout.

Unfortunately, it looks like the existing options are either unmaintained (ex: https://github.com/bodil/typed-html which seems nice, https://github.com/utkarshkukreti/markup.rs also seems interesting) or part of a massive transpiling-to-WASM web framework (ex: [yew's html! macro](https://yew.rs/docs/concepts/html)).

tools/src/bin/output-file.rs
asutherland: Ah yeah, I'd meant for this to be `{:?}` to get the debug rendering of the condition, unless you meant something else?  I've changed it to `{:?}` locally.

staktrace: Yeah that's what I meant, sorry for unclarity!

PR #358, opened 2020-07-29
Add some support for classifying third-party code.

tomprince:
This adds some support for classifying files as third-party code. It would be possible to do better by looking at `tools/rewriting/ThirdPartyPaths.txt`. I was looking for use of our vendored python code in-tree, and it would be handy to filter out the third-party code on searchfox, and this would be enough to cover that case. I've not had to deal with any of the other third-party code in-tree so I don't have a feeling for whether the classification of the other paths in  `tools/rewriting/ThirdPartyPaths.txt` would be useful.

I've not setup vagrant to test this, so this code is untested.

### DIFF
--- a/router/router.py
+++ b/router/router.py
@@ -119,7 +119,7 @@ class SearchResults(object):
 
     max_count = 1000
     max_work = 750
-    path_precedences = ['normal', 'test', 'generated']
+    path_precedences = ['normal', 'test', 'generated', 'thirdparty']
     key_precedences = ["Files", "IDL", "Definitions", "Assignments", "Uses", "Declarations", "Textual Occurrences"]
 
     def categorize_path(self, path):
@@ -138,6 +138,8 @@ class SearchResults(object):
 
         if '__GENERATED__' in path:
             return 'generated'
+        elif path.startswith("third_party/"):
+            return "thirdparty"
         elif is_test(path):
             return 'test'
         else:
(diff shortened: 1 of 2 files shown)


staktrace:
> I've not setup vagrant to test this, so this code is untested.

Are you intending to do this?

tomprince:
> > I've not setup vagrant to test this, so this code is untested.
> 
> Are you intending to do this?

I'd definitely prefer not to need to figure out how to get an dev environment setup.

staktrace:
I applied this patch locally and it basically works. However I noticed that you put the "thirdparty" as the last section in the search results, after "generated". I think it would be better to move it up to either right after "normal" or right after "tests". In terms of what people care about I imagine generated files have the least importance, and it's a bit of a tossup between thirdparty code and tests.

People who are currently used to seeing the thirdparty results mixed in with the "normal" results might be surprised if they get moved down past the tests, and might be misled into thinking those results aren't there. So that's an argument for putting the "thirdparty" section just below "normal". @asutherland any opinion?

asutherland:
I strongly agree thirdparty should come before generated.  I also agree that `['normal', 'thirdparty', 'test', 'generated']` is probably the best incremental change to avoid people perceiving false negative search results.

staktrace:
Closing as I updated the patch in #361 and merged it. (I didn't have permissions to edit this in-place)

tomprince:
Thanks!

PR #359, opened 2020-07-29
Fix re-provisioning.

emilio:
If that directory exists already and you run `vagrant provision`, then
the provisioning script errors which is bad.

### DIFF
--- a/infrastructure/web-server-provision.sh
+++ b/infrastructure/web-server-provision.sh
@@ -73,6 +73,7 @@ fi
 
 # Install gRPC python libs and generate the python modules to communicate with the codesearch server
 sudo pip3 install grpcio grpcio-tools
+rm -rf livegrep-grpc3
 mkdir livegrep-grpc3
 python3 -m grpc_tools.protoc --python_out=livegrep-grpc3 --grpc_python_out=livegrep-grpc3 -I livegrep/ livegrep/src/proto/config.proto
 python3 -m grpc_tools.protoc --python_out=livegrep-grpc3 --grpc_python_out=livegrep-grpc3 -I livegrep/ livegrep/src/proto/livegrep.proto


staktrace:
Hm, this came up on https://github.com/mozsearch/mozsearch/pull/353 and the (without in-depth discussion) conclusion was that generally doing a re-provision of an existing instance isn't a good idea. In particular after that PR we avoid doing redundant work, but that also means if you re-provision and say we updated the cinnabar revision, your reprovisioning won't update to the new cinnabar version because it will detect cinnabar as already installed.

Maybe we should make re-provisioning fail early instead, with an error that it's not supported?

/cc @asutherland too in case he has further thoughts on this.

asutherland:
My take would be that if we want to support re-provisioning, then we should really support it, which means tests or having some invariants about the provisioning.  Rather than manually writing tests that would likely be error prone, it would make sense in that case to move towards using something like chef/puppet/whatever has succeeded them which using a declarative mechanism and can verify the result state and perform upgrades in a less ad-hoc fashion.  My most recent experience with this is very old and is https://github.com/asutherland/deuxdrop/tree/develop/deploy, but since in general we want to be converging with taskcluster, presumably we'd want to find out what taskcluster or related projects do.  For treeherder, it seems like they use docker-compose maybe? https://treeherder.readthedocs.io/installation.html#server-and-full-stack-development

That seems like more work than failing early, so my default vote would be to fail early and with an apology and potentially make sure our vagrant configuration isn't doing anything that would suppress caching of the base image.  The reality of contribution to searchfox is that most contributors are likely to be a subset of web browser developers, and web browser developers tend to need to have a robust internet connection, so the added cost of re-downloading parts of the image should hopefully not be an actual barrier to contribution.

staktrace:
I've written a (completely untested) patch that explicitly fails on reprovisioning, https://github.com/staktrace/mozsearch/commit/95df2ccab0f64696b870fc2d70aa89ecf3e4465c - I'll test it and put up a PR in a bit.

staktrace:
See https://github.com/mozsearch/mozsearch/pull/360

PR #361, opened 2020-07-31
Add some support for classifying third-party code.

staktrace:
Replaces #358 since I can't push to that branch. This is the same patch (and retains authorship) but tweaks the order of sections in the results page, changes a pair of " quotes to ' quotes for consistency, and adds a third_party/ folder to the test repo for testing purposes.


staktrace:
r=me (original patch by tom prince)

PR #362, opened 2020-08-06
Updates robots.txt to allow homepage indexing

karlcow:
As the homepage has no specific name we need to disable more things.

The current robots.txt also disallow everything

Allow: /mozilla-central/source/

which means the homepage of mozilla-central (`https://searchfox.org/mozilla-central/source`) is not accessible but the subdirectories are. (`https://searchfox.org/mozilla-central/source/*`) So I'm fixing this too.

### DIFF
--- a/static/robots.txt
+++ b/static/robots.txt
@@ -1,6 +1,8 @@
-# Disallow everything
+# Disallow almost everything, but make the homepage accessible
 User-Agent: *
-Disallow: /
+Disallow: */search?q=*
+Disallow: */source*
+Disallow: /static/
 
 # ... except the nginx-served pregenerated content for the repos
 # that Mozilla is the canonical owner for. For things like whatwg-html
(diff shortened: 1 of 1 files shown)


karlcow:
This probably needs a followup on 
https://github.com/mozsearch/mozsearch-mozilla/blob/master/help.html
and/or
https://github.com/mozsearch/mozsearch/blob/b906106ce547fc6bca13ea5e2c0dc0550eb44375/scripts/output.js

To have a better search results in search engines out there. Something that will be meaningful in the results.

karlcow:
This can help you test.
And the why is that I didn't remember the name of searchfox today and searched on duckduckgo and google and could not find directly searchfox.org

https://technicalseo.com/tools/robots-txt/

Capture d’écran 2020-08-06 à 20 38 31

staktrace:
> And the why is that I didn't remember the name of searchfox today and searched on duckduckgo and google and could not find directly searchfox.org

What kind of search terms did you use, out of curiosity?

Thank you for the patch though. If I understand correctly the main goal is to allow indexing of the main "searchfox.org/" page which doesn't have any specific "file" component in the URL, and so you have to do these hacky workarounds in the robots.txt file. I think it would be better to just have an /index.html path set up in the nginx config, and allow that specific file in the robots.txt.

As for the `/mozilla-central/source` vs `/mozilla-central/source/` URLs, both of those have the same content. But the links on the main page point to `.../source` and we can update those to point to `.../source/` which will allow the crawler to crawl them based on the current robots.txt file. Does that seem like an acceptable solution to you?

staktrace:
I put up https://github.com/mozsearch/mozsearch/pull/363 and https://github.com/mozsearch/mozsearch-mozilla/pull/109 which should obsolete this patch (based on my previous comment). We can still certainly improve the content on the help page, perhaps adding meta tags and sprinkling SEO dust. For that I'd want to know what kinds of queries @karlcow  used to search.

karlcow:
😭 

hehe. So I tried 
https://www.google.com/search?client=firefox-b-d&q=firefox+source+code+search

and 
https://www.google.com/search?client=firefox-b-d&q=gecko+source+code+search

which gives links to documentation which eventually links to searchfox, but no direct links to searchfox.

PR #363, opened 2020-08-07
Improve indexing

staktrace:
Obsoletes part of #362.

### DIFF
--- a/scripts/nginx-setup.py
+++ b/scripts/nginx-setup.py
@@ -222,6 +222,12 @@ location('= /', [
     'add_header Cache-Control "must-revalidate";',
 ])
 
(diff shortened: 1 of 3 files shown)


staktrace:
Yeah I figured the googlebot will hit `/index.html` and regular users will hit `/`. There might be some value in redirecting regular users from `/` to `/index.html` but as a user I prefer the clean `/` URL in my URL bar :)

asutherland:
> Yeah I figured the googlebot will hit `/index.html` and regular users will hit `/`. There might be some value in redirecting regular users from `/` to `/index.html` but as a user I prefer the clean `/` URL in my URL bar :)

Agreed!

PR #366, opened 2020-08-07
Bug 1656009 - Refine the case-sensitivity checkbox behaviour

staktrace:
This prevents URLs from overriding the user's saved default behaviour,
and also fixes the reported bug where the checkbox state doesn't match
the actual results.


asutherland:
Great comment!

PR #367, opened 2020-08-11
Bug 1566874 - Initial direct code coverage integration. r=kats,emilio

asutherland:
Test run is up at https://asuth.searchfox.org/.  I'm expecting potentially a lot of discussion about what we want this to look like and perhaps also about CSS selector performance.  You can definitely feel the theme changes at https://asuth.searchfox.org/mozilla-central/source/dom/indexedDB/ActorsParent.cpp when changing the themes via the select boxes I crammed into the navigation panel.  I'm planning to wait for reviews from both of you unless Emilio has gone on vacation for a month :)

It probably makes sense to discuss styling more on the bug proper at https://bugzilla.mozilla.org/show_bug.cgi?id=1566874 but maybe selector nitty gritty here?


emilio:
Interesting, I'd have gone for "Fixed no gap", but on the left and with slightly lighter colors.

asutherland:
I'm using https://colorbrewer2.org/#type=diverging&scheme=RdBu&n=11and https://colorbrewer2.org/#type=diverging&scheme=PiYG&n=11 as the basis for the color schemes now, which does allow some latitude for changing the foreground colors while sticking to the themes, etc.

staktrace:
> Interesting, I'd have gone for "Fixed no gap", but on the left and with slightly lighter colors.

Agree. Having it on the right is fine too if that makes the code simpler, but I do like "Fixed no gap" the most. Having a hover tooltip would be good too, so the bar's meaning is more discoverable.

tools/src/output.rs
```diff
@@ -375,6 +375,33 @@ pub fn generate_panel(
             F::S(
                 r#"<label class="panel-accel"><input type="checkbox" id="panel-accel-enable" checked="checked">Enable keyboard shortcuts</label>"#,
             ),
+            F::S(r#"<div><label class="panel-theme">"#),
```
staktrace, line 378: I assume we're just going to pick a theme instead of offering this selector :)

tools/src/bin/output-file.rs
```diff
@@ -168,8 +168,17 @@ fn read_test_info_from_file_info(per_file_info: Option<&json::Object>) -> Option
...
 
-fn get_per_file_info<'a>(all_info: &'a json::Object, path: &str) -> Option<&'a json::Object> {
-    let mut cur_obj = all_info.get("root")?.as_object()?;
+/// Per-file info derived from the concise and detailed info for a given file.
+/// Everything in here is optional data, but this structure will be available
+/// for every file to simplify control-flow.
+struct PerFileInfo {
+    bugzilla_component: Option<(String, String)>,
+    test_info: Option<TestInfo>,
+    coverage: Vec<i32>,
```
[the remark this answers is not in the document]
asutherland: Yeah, semantically `Option<Vec<i32>>` probably makes more sense.  The current structuring was based on me knowing there was no way that I'd create dummy coverage files for every file and that the coverage data uses `-1` to convey that a line isn't covered, so as far as the output logic is concerned, it just treats a lack of data as a `-1`.  But obviously it can do this if there's no array just as well.

I'll plan to make this change, as it likely makes sense to also eventually include more information about the coverage which would live next to the Vec inside the Option.  Like when the coverage run ran, etc.

tools/src/output.rs
asutherland: Yeah, as much fun as changing things is, this takes up too much space, etc.

asutherland:
> > Interesting, I'd have gone for "Fixed no gap", but on the left and with slightly lighter colors.
> 
> Agree. Having it on the right is fine too if that makes the code simpler, but I do like "Fixed no gap" the most. Having a hover tooltip would be good too, so the bar's meaning is more discoverable.

I'll plan to add a "fixed no gap left" option in the next revision, hopefully after getting feedback from :Jamie, as I expect the hover or its content will also want to be part of the accessibility tree story.

If the coverage is a hover target, would we want it to be the same width as the blame bar?  I made a CSS variable that controls the width of the fixed bar, `--cov-strip-width: 5px`, which is almost certainly too small for hover purposes.

staktrace:
> If the coverage is a hover target, would we want it to be the same width as the blame bar? I made a CSS variable that controls the width of the fixed bar, `--cov-strip-width: 5px`, which is almost certainly too small for hover purposes.

I was thinking 5px is ok. It really just needs to be hovered once by a person, to discover what the bar is. After that presumably the person will remember. (As opposed to blame, where one often wants to hover to see the blame for a particular line, which warrants making it easier to hover).

docs/index-directory-contents.md
```diff
@@ -0,0 +1,168 @@
...
+  `mkindex.sh`.
+- `jumps`: Lookup table that maps from machine symbol names to their canonical
+  definition point.  Produced by `crossref.rs` and consumed by `output-file.rs`
+  so that the context menus can in the HTML files can generate definition links
+  without having to involve any server queries.  See [crossref.md](crossref.md)
+  for more info.
+- `linux64.*`: A bunch of per-platform files downloaded by
+  `fetch-tc-artifacts.sh` that we retain for debugging
+  `process-gecko-analysis.sh`.
+- `livegrep.idx`
```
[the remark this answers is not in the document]
asutherland: Thanks!  Using this and all your above changes.

tools/src/bin/derive-per-file-info.rs
```diff
@@ -196,7 +225,43 @@ fn traverse_and_store_bugzilla_map(bz_dir: &json::Object, all_node: &mut json::O
...
+///
+/// Each node can contain the following keys:
+/// - "children": An object whose keys are file/directory names and whose values
+///   are nodes of the self-same type.
+/// - "coverage": An array where each entry corresponds to a line of the source
+///   file with `-1` indicating an uninstrumented line, `0` indicating an
+///   instrumented line with no coverage hits, and any positive integers
+///   indicating a line with that number of hits.
+/// - "coveragePercent": Coverage percent in the node and all its children as a
+///   floating point value in the range [0, 100] which 2 digits of precision
```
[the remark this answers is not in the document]
asutherland: Changed this to "Coverage percent in the node and all its children as a floating point value in the range [0, 100] to 2 decimal places."

asutherland:
@emilio @staktrace For the hover, does it make sense to repurpose a slightly generalized blame popup?

staktrace:
Assuming by that you mean you're going generalize the blame popup code a bit and use the generalized version for the code coverage hover popup - that sounds good to me!

asutherland:
Yes, that.  Thanks!  (And thanks to Emilio for having cleaned that all up!)

asutherland:
@staktrace @emilio So I think this is pretty much ready for review.  I kicked off an indexing run just now but I think it'll take 4 hours.  The only real differences between the current https://asuth.searchfox.org/ and the next run are:
- typo fix, some dead CSS removed
- the gzip enhancement
- being able to click on the blame/coverage strips to make them go away

My primary review concern at this time is the attempted magic using CSS variables to make the coverage detail toggle happen when hovering on the coverage strip.  My hope was that unexplained magic would make things more performant than changing an attribute on the root of the file which would cause tons of style recomputation.  (As witnessed with the original draft of this patch with the theme selector, that's definitely not magically performant.)

But when I test on our worst-case file of https://asuth.searchfox.org/mozilla-central/source/dom/indexedDB/ActorsParent.cpp I can see that the hover magic takes about a second to toggle on/off.  This makes me wonder if it would be better to instead do something that instead just sets/clears an attribute on the coverage cells that are currently displayed on the viewport plus a lip.  (This could be integrated with the position: sticky logic that dynamically determines what's stuck so we centralize the sketchiness.)

I had thought this might also be doing bad things for the NVDA screen reader case, but as I do some A/B testing now, it appears that NVDA seems to be having trouble when we clear the blame popup and the problem isn't related to the coverage hover styling at all.  That is, I can induce jank by just hovering over the blame bar and then trying to unhover and not involving the coverage cells at all.  A delay of approximately 2-3 seconds is added on https://asuth.searchfox.org/mozilla-central/source/dom/serviceworkers/ServiceWorkerPrivate.cpp for me.  It's possible this could be related to and improved/fixed by https://bugzilla.mozilla.org/show_bug.cgi?id=1668707 ?

That said, for the screen reader case, the expected use of searchfox doesn't actually result in the popups being popped up much, and the enhancements I've made to the aria-labels will ideally further reduce use of the popups.

scripts/compress-outputs.sh
```diff
@@ -0,0 +1,67 @@
...
+#
+# It turns out mozilla-central has exactly this case in the guise of
+# `devtools/client/styleeditor/test/simple.css` and its gzipped twin
+# `devtools/client/styleeditor/test/simple.css.gz`.  Right now (pre this patch)
+# if you view the ".gz" file you'll see a useless single gibberish line of text.
+#
+# Our fix:
+# - We observe that if we process the existing "FOO.gz" file before "FOO", then
+#   when FOO's gzipping overwrites "FOO.gz", there's no harm because it's
+#   overwriting a zero-length file that only needed to exist so its filename
```
staktrace, line 33: Technically FOO.gz is not going to be a zero-length file. It will contain the original FOO.gz contents (which got compressed into FOO.gz.gz) with the timestamp updated via touch to match that of FOO.gz.gz. But your point remains valid that FOO.gz is only used to satisfy the try_files requirement and not actually read when the user requests FOO.gz, so it's ok that the file contains the compressed version of FOO instead.

scripts/compress-outputs.sh
```diff
@@ -0,0 +1,67 @@
...
+# - We use `gzip -f` so that gzip doesn't care when it overwrites the zero
+#   length file.
+# - We note that although we try and be responsible with the timestamps in this
+#   script, they don't actually matter at all because they're the timestamps for
+#   generated HTML files and don't have timestamps corresponding to the
+#   underlying revision controlled files.  (Also, the timestamp that gets served
+#   to users is that of the gzip file anyways and so the overwrite case doesn't
+#   do any harm there either.)
+
+echo "Compressing files in ${INDEX_ROOT}/file/ with zero-length marker for try_files"
```
staktrace, line 45: Looking at the ~/index/index-log file on your asuth server, this whole compression process takes about 17 minutes for m-c. While not bad, it seems eminently parallelizable, as long as you process all the `*.gz` files in one pass and then do the rest of the files. That might also be better from a robustness point of view, to not rely on the `sort -r` behaviour of listing .gz files first.

This can be spunoff as a followup though as it's a relatively contained and minor enhancement. My completely untested rough approach would be:

```
find . -type f > all-files
grep ".gz$" all-files | parallel --halt now,fail=1 "gzip -f {} && touch -r {}.gz {}"
grep -v ".gz$" all-files | parallel --halt now,fail=1 "gzip -f {} && touch -r {}.gz {}"
```

(with appropriate quoting, which is the hard part...)

staktrace (commented):
Comments from first two patches. I'll review the other two soon but not immediately.

static/js/blame.js
```diff
@@ -348,7 +348,24 @@ var BlameStripHoverHandler = new (class BlameStripHoverHandler {
...
+      // We could check `keepVisible` here but that creates a weird situation
+      // when activating via mouse where we get the sequence 1) hover triggers
+      // popup, 2) click maintains popup and makes it keepVisible, and then 3)
+      // second click hides popup.  By not checking, we get 1) hover triggers
+      // popup, 2) click hides popup.
+      if (isClick && this.mouseElement === event.target) {
+        this.keepVisible = false;
+        this.mouseElement = null;
+        BlamePopup.blameElement = null;
+        return
```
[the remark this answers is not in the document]
asutherland: Ah, yeah, I've gotten a bit dependent on eslint and vscode automatically running it as I type.  It would be great to enable that.

scripts/compress-outputs.sh
asutherland: I think you're describing what happens if we do `gzip -k` for keep.  If we have a file `FOO` and run `gzip FOO`, then afterwards we only have a file `FOO.gz`.  The zero length file is created by using `touch`.

asutherland (commented):
I actually had trouble sleeping in this morning because my brain was fixated on how I probably should have parallelized the operation.  Unfortunately I foolishly didn't check for review comments before starting, but the good news is that I think we ended up on the same page, although I've now done `s/;/&&/` seeing your rough approach handling that better.  The only thing I had to add was handling for the case when that initial grep doesn't find a ".gz" and has an exit code of 1.

asutherland:
Okay, https://dev.searchfox.org/ up with the results of the parallel run... mozilla-beta took 12 minutes still?  I guess that shouldn't be totally surprising that things could end up I/O bound.  I guess we need to make output-file directly write gzipped output-file.  Hopefully that's somewhat I/O-bound right now so we'll have CPU to burn on the gzipping and then things won't be I/O bound because they'll all be writing 1/10th of what they were writing before.

asutherland:
So, the attempt to use flate2 with a Cargo.toml directive to use the system zlib via:
```
flate2 = { version = "1.0", features = ["zlib"], default-features = false }
```

resulted in an error of:
```
double free or corruption (fasttop)            
Aborted (core dumped)
```
which seems bad.  I'm retrying using the allegedly slower pure rust zlib impl.

asutherland:
I re-ran and got a backtrace and it seems like maybe I shouldn't have upgraded libgit2 in order to fix the zlib-sys version mismatch issue:
```
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff6c788b1 in __GI_abort () at abort.c:79
#2  0x00007ffff6cc1907 in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7ffff6deedfa "%s\n") at ../sysdeps/posix/libc_fatal.c:181
#3  0x00007ffff6cc897a in malloc_printerr (str=str@entry=0x7ffff6df0a88 "double free or corruption (fasttop)") at malloc.c:5350
#4  0x00007ffff6cd0074 in _int_free (have_lock=0, p=0x555556317900, av=0x7ffff7023c40 <main_arena>) at malloc.c:4230
#5  __GI___libc_free (mem=0x555556317910) at malloc.c:3124
#6  0x00005555559f1e8f in stdalloc__free (ptr=0x555556317910) at libgit2/src/allocators/stdalloc.c:104
#7  0x00005555559b6b78 in git_mwindow_close_lru_window () at libgit2/src/mwindow.c:269
#8  0x00005555559b6dbb in new_window (fd=8, size=5324050218, offset=2076982455) at libgit2/src/mwindow.c:337
#9  0x00005555559b6fd4 in git_mwindow_open (mwf=0x555556338e60, cursor=0x7fffffff3b88, offset=2076982455, extra=20, left=0x7fffffff3b10) at libgit2/src/mwindow.c:407
#10 0x00005555559bf892 in git_packfile_unpack_header (size_p=0x7fffffff3ba0, type_p=0x7fffffff3b7c, mwf=0x555556338e60, w_curs=0x7fffffff3b88, curpos=0x7fffffff3b90) at libgit2/src/pack.c:455
#11 0x00005555559bfdf2 in pack_dependency_chain (chain_out=0x7fffffff3c70, cached_out=0x7fffffff3c40, cached_off=0x7fffffff4550, small_stack=0x7fffffff3cd0, stack_sz=0x7fffffff3c48, 
    p=0x555556338e60, obj_offset=2076982455) at libgit2/src/pack.c:581
#12 0x00005555559bffe8 in git_packfile_unpack (obj=0x7fffffff4530, p=0x555556338e60, obj_offset=0x7fffffff4550) at libgit2/src/pack.c:637
#13 0x0000555555a04d95 in pack_backend__read (buffer_p=0x7fffffff45e0, len_p=0x7fffffff45e8, type_p=0x7fffffff45f0, backend=0x555556316040, oid=0x7fffffff4aa0) at libgit2/src/odb_pack.c:399
#14 0x00005555559b032d in odb_read_1 (out=0x7fffffff46b8, db=0x555556316140, id=0x7fffffff4aa0, only_refreshed=false) at libgit2/src/odb.c:1065
#15 0x00005555559b053e in git_odb_read (out=0x7fffffff46b8, db=0x555556316140, id=0x7fffffff4aa0) at libgit2/src/odb.c:1116
#16 0x000055555598d50f in git_object_lookup_prefix (object_out=0x7fffffff47a8, repo=0x555556314c20, id=0x7fffffff4aa0, len=40, type=GIT_OBJECT_BLOB) at libgit2/src/object.c:222
#17 0x000055555598d5dd in git_object_lookup (object_out=0x7fffffff47a8, repo=0x555556314c20, id=0x7fffffff4aa0, type=GIT_OBJECT_BLOB) at libgit2/src/object.c:253
#18 0x00005555559a5776 in git_blob_lookup (out=0x7fffffff47a8, repo=0x555556314c20, id=0x7fffffff4aa0) at libgit2/src/object_api.c:122
#19 0x000055555595437a in git2::repo::Repository::find_blob (self=0x7fff3332e200, oid=...) at /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/git2-0.13.11/src/repo.rs:1044
#20 0x00005555556e0058 in tools::git_ops::compute_line_map (repo=0x7fff3332e200, new_oid=..., old_oid=...) at src/git_ops.rs:96
#21 0x00005555556e2445 in tools::git_ops::map_to_previous_version (git_data=0x7fff3332e200, rev=..., target_file=0x7fff030378e0, cache=...) at src/git_ops.rs:284
#22 0x00005555556e2bfe in tools::git_ops::find_prev_blame (git_data=0x7fff3332e200, rev=..., target_file=0x7fff030378e0, lineno=4, cache=0x7fffffff7110, diff_cache=...) at src/git_ops.rs:312
#23 0x00005555556bf47e in tools::format::format_file_data (cfg=0x7fffffff8c60, tree_name=..., panel=..., info_boxes=..., commit=0x7fffffff9060, blame_commit=0x7fffffff8ee0, path=..., data=..., 
    jumps=0x7fffffff8d50, analysis=..., coverage=0x7fffffff9ac0, writer=..., diff_cache=...) at src/format.rs:487
#24 0x00005555555e7983 in output_file::main () at src/bin/output-file.rs:796
```

gonna try randomly downgrading and/or upgrading things until they become happier.

asutherland:
On the indexer, reverting all changes to Cargo.lock and Cargo.toml and just adding `flate2 = "1"` seems to result in a stable output-file run but uses the slower pure rust gzip impl.  I'm going to update the push and retry and we'll see how it goes in terms of total indexing runtime for mozilla-central.

asutherland:
Okay, I think we can declare victory even without native zlib.  On my mozilla-central-only "asuth" run that just completed, measuring from the first date directly after `mkindex.sh` for the repo through the date right before the `+ /home/ubuntu/mozsearch/scripts/check-index.sh /mnt/index-scratch/config.json mozilla-central filesystem ''` line which happens after compress-output.sh (when present), we get:
- This code: 52.8 minutes
- Trunk: 62.3

So with the gzip fixes, it's now 10 minutes faster with compression.  The caveat is that we know that things mysteriously slow down as we index each repo.  So I'll trigger a full config.json run now on "asuth" so we can do a more apples-to-apples comparison.

asutherland:
@staktrace @emilio Okay, I think this is good for final review for reals now.  There's more that we could do in terms of having merge-analyses or other pieces of the analysis flow pre-gzip the analysis files and the `dir` files, but assuming the "asuth" full config.json run I just kicked off clocks in at less than the trunk runs, it seems like something that could be landed and later refined.

In terms of the hover styling change performance issue I raised, I'd say for landing the decision is now whether to disable the hover styling change or not.  It'll be at least a week before I think I can find the time to try and optimize it further.  I think the 1sec performance on the worst-case IndexedDB ActorsParent.cpp on a developer-class machine is sufficiently survivable and the extra information is sufficiently useful, but it's easy enough to disable the functions that make the change.  (Note that I don't recommend trying to do any a11y testing on that file, at least last time I tried to open the a11y tree for that file, Firefox locked up forever.)

scripts/compress-outputs.sh
staktrace: Ah yes, you're right, my mistake!

tools/src/bin/output-file.rs
```diff
@@ -199,6 +199,98 @@ fn read_detailed_file_info(path: &str, index_path: &str) -> Option<json::Object>
...
+///   information about the AST or nesting contexts might help
+///
+/// Our arbitrary decisions here are:
+/// - Leave the starts and ends of file uncovered.  This is more realistic but
+///   at the cost of this realism making it less obvious that interpolation is
+///   present in the rest of the file, potentially leading to bad inferences.
+///   - We attempt to mitigate this by making sure the hover information makes
+///     it clear when interpolation is at play so if someone looks into what's
+///     going on they at least aren't misled for too long.
+/// - Maximally interpolate hits over misses.  Our goal is that people's eyes
```
staktrace, line 239: 👍

tools/src/bin/output-file.rs
```diff
@@ -199,6 +199,98 @@ fn read_detailed_file_info(path: &str, index_path: &str) -> Option<json::Object>
...
+            let next_val = raw[j];
+            if next_val == -1 {
+                continue;
+            }
+            // We've found a value which means that both last_noninterp_val and
+            // next_val are >= 0.  (last_noninterp_val can never be -1 because
+            // we start the loop with have_interp_val=true.)
+
+            // Favor hits over misses (see func doc block for rationale).
+            if val > 0 || last_noninterp_val > 0 {
```
staktrace, line 287: Shouldn't this be `next_val > 0` instead of `val > 0`? Tangentially, this function looks like a nice candidate for a few unit tests.

tools/src/format.rs
```diff
@@ -413,6 +419,37 @@ pub fn format_file_data(
     for (i, line) in output_lines.iter().enumerate() {
         let lineno = i + 1;
 
+        // Compute the coverage data for this line (if any)
+        let coverage_data: String = if let Some(ref coverage) = coverage {
+            // There's 2 levels of not having data for a line here:
+            // 1. We had no coverage data, coverage is None.  In that case,
+            //    we'll map to -5.
```
staktrace, line 426: Doesn't look like we actually use -5 anywhere, it just emits `cov-no-data` directly. So update the comment I guess?

scripts/compress-outputs.sh
```diff
@@ -42,26 +42,33 @@ set -o pipefail # Check all commands in a pipeline
...
-echo "Compressing files in ${INDEX_ROOT}/file/ with zero-length marker for try_files"
-pushd ${INDEX_ROOT}/file/
+function compress_dir_with_touch {
+  echo "Compressing files in $1 with zero-length marker for try_files"
+  pushd $1
+  find . -type f > ../compress-file-list
+  # It's possible there won't be matches, in which case grep will exit with 1,
+  # which we want to eat.
+  { egrep '\.gz$' ../compress-file-list || test $? = 1; } | parallel --halt now,fail=1 'gzip -f {} && touch -r {}.gz {}'
+  egrep -v '\.gz$' ../compress-file-list | parallel --halt now,fail=1 'gzip -f {} && touch -r {}.gz {}'
```
staktrace, line 52: Might be worth adding the `|| test` bit here too, in case the repo has only gzip files or no files at all.

scripts/compress-outputs.sh
```diff
@@ -42,26 +42,33 @@ set -o pipefail # Check all commands in a pipeline
...
+# We could also have parameterized the above to take the command but this seemed
+# less confusing / harder to mess up.
+function compress_dir_only {
+  echo "Compressing files in $1 without zero-length marker"
+  pushd $1
+  find . -type f > ../compress-file-list
+  # It's possible there won't be matches, in which case grep will exit with 1,
+  # which we want to eat.
+  { egrep '\.gz$' ../compress-file-list || test $? = 1; } | parallel --halt now,fail=1 'gzip -f {}'
+  egrep -v '\.gz$' ../compress-file-list | parallel --halt now,fail=1 'gzip -f {}'
```
staktrace, line 63: Ditto

staktrace (approved):

tools/src/bin/output-file.rs
asutherland: Ah, yes, you're absolutely right on both counts.  When I implemented this and it seemed to work immediately I didn't ask too many questions and I think I got fixated on other things that were more obviously on fire, but I absolutely should have implemented an in-file test func.  Thanks very much for calling this out!

Your specific example is now caught by the:
```rust
        vec![
            vec![1, -1, 0, -1, 1],
            vec![1, -2, 0, -2, 1]
        ],
```
case.  With the failure having been reported as:
```
---- test_interpolate_coverage stdout ----
thread 'test_interpolate_coverage' panicked at 'assertion failed: `(left == right)`
  left: `[1, -2, 0, -3, 1]`,
 right: `[1, -2, 0, -2, 1]`', src/bin/output-file.rs:339:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The following test case also uncovered the lack of a `break` at the bottom of the loop!
```rust
        // combine all of the above (except for whole file), single interp each.
        vec![
            vec![-1, -1, 0, -1, 0, -1, 1, -1, 1, -1, 1, -1, 0, -1],
            vec![-1, -1, 0, -3, 0, -2, 1, -2, 1, -2, 1, -2, 0, -1]
        ],
```

Manifesting the failure:
```
---- test_interpolate_coverage stdout ----
thread 'test_interpolate_coverage' panicked at 'assertion failed: `(left == right)`
  left: `[-1, -1, 0, -3, 0, -3, 1, -2, 1, -2, 1, -2, 0, -1]`,
 right: `[-1, -1, 0, -3, 0, -2, 1, -2, 1, -2, 1, -2, 0, -1]`', src/bin/output-file.rs:339:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #368, opened 2020-08-17
panel: Make clicking on the permalink panel when already in there copy to clipboard

emilio:
Rather than pushing duplicate entries to session history.

I think this is nicer than context-menu + copy link location, or click +
copy urlbar, which are the other two alternatives to do this.

### DIFF
--- a/static/js/panel.js
+++ b/static/js/panel.js
@@ -20,15 +20,17 @@ var Panel = new (class Panel {
     );
 
     if (this.permalinkNode) {
+      this.permalinkNode.title += " (double-click to copy)";
       this.permalinkNode.addEventListener("click", event => {
         if (event.altKey || event.ctrlKey || event.metaKey || event.shiftKey) {
           return;
         }
-        window.history.pushState(
-          { permalink: event.target.href },
-          window.title,
-          event.target.href
-        );
+        let url = event.target.href;
+        if (window.location.href == url) {
+          navigator.clipboard.writeText(url);
+        } else {
+          window.history.pushState({ permalink: url }, window.title, url);
+        }
         event.preventDefault();
       });
     }


static/js/panel.js
```diff
@@ -20,15 +20,17 @@ var Panel = new (class Panel {
     );
 
     if (this.permalinkNode) {
+      this.permalinkNode.title += " (double-click to copy)";
```
staktrace, line 23: I'd prefer the wording " (click twice to copy)" since "double-click" specifically refers to the thing where clickCount=2 and that's not what you're checking for.

asutherland:
This seems like awkward UX.  One idiom Mozillians might already be familiar with is showing a copy/clipboard icon to the right of a link.  For example, the [treeherder job for today's searchfox linux job](https://treeherder.mozilla.org/#/jobs?repo=mozilla-central&searchStr=searchfox&selectedTaskRun=cAxC5VBNSfG-Okh3GXoICA.0) has a clipboard icon with a tooltip that when clicked on copies to the clipboard.  For accessibility it could also be in the tab sequence as a full button.  The [task UI](https://firefox-ci-tc.services.mozilla.com/tasks/cAxC5VBNSfG-Okh3GXoICA) itself has rows like "Created" which show the mobile version of the copy icon with the 2 overlapping squares, and clicking anywhere on the row results in a copy operation taking place, with a time-based animation showing feedback.

Github similarly displays a clipboard icon to the right of the "Clone with FOO" icon if you click on the "Code" dropdown on https://github.com/mozsearch/mozsearch.

If we expect people actually want to change to the given revision / have a usable link, then we probably want the separate button which should take up less space than the extra unusual text.  If we think people only ever want to copy the URL, then having the whole row's action be copying would be best.

Given how things currently work and given that people may continue clicking around to change the selected line numbers, the separate copy button approach seems best.

That said, it likely does make sense to suppress injecting additional history entries if people mash the permalink button.

emilio:
Closing in favor of  #377

PR #371, opened 2020-10-10
Make the rust code successfully index the llvm repo

staktrace:


asutherland (approved):
I've been starting the instance like a sucker!  I keep forgetting ssh can do that!

PR #372, opened 2020-10-12
Bug 1669776 - Add hgrev support

staktrace:


tools/src/bin/web-server.rs
```diff
@@ -94,20 +96,47 @@ fn handle(
...
+                    .arg("hg2git")
+                    .arg(&rev)
+                    .current_dir(&git_path)
+                    .output();
+                match output_result {
+                    Ok(output) if output.status.success() => {
+                        rev = git_ops::decode_bytes(output.stdout).trim().to_string();
+                    }
+                    Ok(output) => {
+                        println!("Cinnabar exit code: {:?}", output.status.code());
```
emilio, line 130: `eprintln!` or something like that?

tools/src/bin/web-server.rs
```diff
@@ -94,20 +96,47 @@ fn handle(
...
-    match &kind[..] {
-        "rev" => {
+    match kind {
+        "rev" | "hgrev" => {
             if path.len() < 3 {
                 return not_found();
             }
 
-            let rev = &path[2];
+            let mut rev = path[2].to_string();
```
emilio, line 109: Maybe just `Cow::Borrowed(&path[2])` then in the branch `rev = Cow::Owned(git_ops...)`. But very in the micro-optimization category so not a big deal either way.

tools/src/bin/web-server.rs
```diff
@@ -94,20 +96,47 @@ fn handle(
...
+                };
+                let output_result = Command::new("git")
+                    .arg("cinnabar")
+                    .arg("hg2git")
+                    .arg(&rev)
+                    .current_dir(&git_path)
+                    .output();
+                match output_result {
+                    Ok(output) if output.status.success() => {
+                        rev = git_ops::decode_bytes(output.stdout).trim().to_string();
```
emilio, line 127: I wonder if it'd be better to just redirect to the git revision... Stuff like the permalink things / blame / etc will still point to git revs...

Would also potentially be nice to have in the menu a "HG link" or something, but that can be future work or what not.

tools/src/bin/web-server.rs
```diff
@@ -94,20 +96,47 @@ fn handle(
...
                 return not_found();
             }
 
-            let rev = &path[2];
+            let mut rev = path[2].to_string();
             let path = path.clone().split_off(3);
             let path = path.join("/");
 
+            if kind == "hgrev" {
+                let tree_config = cfg.trees.get(*tree_name).unwrap();
```
emilio, line 114: `&cfg.trees[*tree_name]`?

tools/src/bin/web-server.rs
emilio: Also may be worth to just format the whole `output` if possible, so that we have something more than the error code.

emilio (approved):

tools/src/bin/web-server.rs
asutherland: If our goal is to maintain the git revisions as our canonical revisions, then I strongly agree here that /hgrev/ should likely just be a permanent redirect.  If we want to move mozilla-central and other hg-using repositories over to being primarily addressed using hg revisions, then I think it makes to possibly support both, but strongly bias such repositories to using hg.

Given that the people previously most focused on enhancing mercurial are no longer at Mozilla (ex: :gps), and that the CVS history of mozilla-central cannot be expressed in hg revisions (I think?), and that mercurial periodically frustrates me personally (the most important heuristic! ;), it seems like it probably makes sense to continue being git-centric for now.  (And therefore doing the redirect.)

asutherland: In terms of the HG link, I agree that that seems like a nice affordance.  That seems like it probably wants to be a separate bug and we should probably reach consensus around the UX for the "I probably just want the URL in my clipboard" use-case and conventions there I discuss in https://github.com/mozsearch/mozsearch/pull/368#issuecomment-674979638 for Emilio's patch for that.

staktrace:
Yeah a permanent redirect makes more sense. I'll redo the patch accordingly.

tools/src/bin/web-server.rs
emilio: Yeah, I plan to look at that soonish :)

staktrace:
@emilio Patch updated to do the redirect

emilio (approved):
LGTM!

PR #373, opened 2020-10-13
Bug 1630646

staktrace:


asutherland (approved):
Insert ill-conceived joke here about Firefox version numbering scheme and version number inflation.

asutherland:
It'd be neat if this also fixes whatever that segfault was that we'd seen last time the fork-bomb happened!

staktrace:
> Insert ill-conceived joke here about Firefox version numbering scheme and version number inflation.

Hah. bazel is now on 3.6.0 so they've probably gone through more released version than Firefox over the same time span.

Also FYI the latest livegrep will include https://github.com/livegrep/livegrep/pull/281 so we'll start getting more warnings in the daily run emails. But they're not actually new warnings, they're just getting caught in the grep match now. At some point I want to add a per-repo warning ignore list to filter out known warnings that are safe to ignore.

staktrace:
New AMIs deployed, today's mozilla-releases web server will get this fix and the rest of the channels will get it with tomorrow's run. I'll kick off a mozilla-archived run as well.

PR #374, opened 2020-10-16
Bug 1670314 - Treat all channel names starting with release as releas…

staktrace:
…e channels

This leaves the existing mozilla-archived and mozilla-releases channels as
release channels too because I'm not sure how long it will take to migrate those
over to new names. (I'm working on this sporadically as time permits)


asutherland (approved):
Restating:
- case, per https://www.gnu.org/software/bash/manual/html_node/Conditional-Constructs.html uses standard [Pattern Matching](https://www.gnu.org/software/bash/manual/html_node/Pattern-Matching.html#Pattern-Matching) which means the `*` "Matches any string, including the null string."

PR #377, opened 2020-10-19
Add a "copy to clipboard" button to (some) panel entries.

emilio:
More explicit alternative to #368.

### DIFF
--- a/static/css/icons.css
+++ b/static/css/icons.css
@@ -72,16 +72,45 @@ a.type {
     padding: 0.5rem
 }
 
+.panel .accel,
+.panel .copy {
+    font-size: 10px;
+    border-radius: 2px;
+    padding: 1px 3px;
+    margin: 0 10px;
+    border: 1px solid #ddd;
+}
+
 .panel .accel {
     /* simulate a keyboard key effect */
     font-family: monospace;
-    font-size: 10px;
-    padding: 1px 3px;
-    border: 1px solid #ddd;
-    border-radius: 2px;
-    box-shadow: 0 1px #999;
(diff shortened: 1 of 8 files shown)


staktrace (approved):
The code looks fine to me. Is it deployed anywhere I can try it out?

asutherland (approved):
I tested this locally with `make build-searchfox-repo` and it seems great!  Love the generalization so that this can be used on many different panel links!

It would probably be good to sanity check with a run against mozilla-central before landing/shipping just to sanity check.  I'll see if one of us is triggering a run on matrix.

asutherland:
Forgot to mention that I particularly appreciate the attention-to-detail in having the checkmark copy confirmation!

staktrace:
So it's up on dev.searchfox.org now (thanks @asuth!) and I noticed that if I click on the icon for the permalink, it just copies it to the clipboard. But if I click on the icon for e.g. the log or raw links, it **also** navigates to the link after copying. It's not clear to me why the behaviour is different for the permalink link vs the other ones.

emilio:
> So it's up on dev.searchfox.org now (thanks @asuth!) and I noticed that if I click on the icon for the permalink, it just copies it to the clipboard. But if I click on the icon for e.g. the log or raw links, it also navigates to the link after copying. It's not clear to me why the behaviour is different for the permalink link vs the other ones.

Yup, noticed that too, that's because the permalink is handled in JS. Added a patch that should address that (and live-edited the server so if you shift-reload you should see that fixed).

emilio:
(Assuming that with that fixed there's no objection to this)

PR #378, opened 2020-10-19
Bug 1672014 - Provision the yq tool

staktrace:
Apparently it only comes in snap form.


asutherland (approved):
> Apparently it only comes in snap form.

Huh, interesting.  But probably wise given how unsafe yaml has turned out to be!

Also, I really should have checked my Thunderbird inbox for this PR before hitting submit on the other one!  My comment about yq was already out of date by 3 minutes! ;)

asutherland:
And 38 minutes from Bugzilla!

PR #381, opened 2020-10-20
Bug 1672004 - Format mjs files as JS

staktrace:


emilio (approved):

asutherland:
If we want semantic analysis, seems like we'd also want https://github.com/mozsearch/mozsearch/blob/d3c7ae324acce8dad9630c9a9e92ce9c85805fe5/scripts/find-repo-files.py#L73 to think they're js too?

PR #383, opened 2020-10-28
Bug 958852 - Allow walk_submodules to be overridden by the config.

staktrace:


asutherland (approved):
Restating:
- livegrep has a walk_submodules JSON config option it understands.  We previously always set it to true, we're now propagating the value from our tree config definition if it's present or defaulting to True otherwise.
- For rust, we're not checking out the submodules because of the points made in https://bugzilla.mozilla.org/show_bug.cgi?id=958852#c7 which is namely that we lose blame if we checkout submodules due to current limitations in blame processing.  But when we don't check them out, codesearch / livegrep freaks out if we tell it to walk the submodules, so we want to be able to tell it not to.

PR #385, opened 2020-11-02
Bug 1674616 - Fix raw-analysis URL endpoint.

staktrace:
The /raw-analysis/ URLs also use try_files, and therefore need the zero-length
non-gzip'd files to exist.

### DIFF
--- a/scripts/compress-outputs.sh
+++ b/scripts/compress-outputs.sh
@@ -66,4 +66,4 @@ function idempotently_gzip_dir_no_touch {
 }
 
 compress_dir_with_touch "${INDEX_ROOT}/dir/"
-idempotently_gzip_dir_no_touch "${INDEX_ROOT}/analysis/"
+compress_dir_with_touch "${INDEX_ROOT}/analysis/"


staktrace:
I tested this on a local run of the test repo but haven't tried it on a full indexer yet. Will do that later today or tomorrow.

asutherland:
When you say local run, do you mean you ran a server-check https://github.com/mozsearch/mozsearch/blob/0e82dd8c5a6f3c9534b4513345306e7d9fed8d5d/scripts/check-helper.sh#L76 ?

Unfortunately the built-in check doesn't do a web-server check, but maybe we should figure out how to do that https://github.com/mozsearch/mozsearch/blob/00a523e80411b576c0bd0c87bac5df9426951802/scripts/mkindex.sh#L82

asutherland (approved):
Check aside, I can tell by inspection you're absolutely right per https://github.com/mozsearch/mozsearch/blob/00a523e80411b576c0bd0c87bac5df9426951802/scripts/nginx-setup.py#L176.  That said, I guess maybe we don't need to use try_files?  That said that said (not sure one can keep appending that phrase, but why not?), consistency is always nice.  (And my suspicion that we weren't consistent here led me down a wrong path!)

staktrace:
> When you say local run, do you mean you ran a server-check

No, I meant a `make build-test-repo` followed by loading `http://localhost:16995/tests/raw-analysis/big_cpp.cpp`. But yeah doing a web-server check as part of the checks script would be nice. I'll see how hard it is.

staktrace:
> But yeah doing a web-server check as part of the checks script would be nice. I'll see how hard it is.

I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1675270 for this so I don't forget.

PR #387, opened 2020-11-04
Bug 1418001 - Add checks for the include file linkage

staktrace:


scripts/check-helper.sh
```diff
@@ -8,6 +8,7 @@ if [[ $# -ne 4 ]]
 then
     set +x  # Turn off echoing of commands and output only relevant things to avoid bloating logfiles
     echo "usage: $0 <check-disk> <server-url> <searchfox user exposed path> <canonical symbol name as found in data-symbols>"
+    echo "Set CHECK_SKIP_DATA_SYMBOL=1 in the environment to skip the check for the symbol in the generated HTML"
```
asutherland, line 11: Maybe add ", currently necessary for zero-length symbols" or something like that?  The `tests/tests/check` comment is fantastic at providing the context here, and I think it would be great to propagate more of a hint of here for why this would be desirable.

asutherland (approved):
Thanks!

scripts/check-helper.sh
staktrace: Done!

PR #388, opened 2020-11-05
Run checks on web-server

staktrace:


asutherland (approved):
Woo!  Yes!  Thank you!

Also, good design call on having the web check just be a web check, allowing the indexer to run the check as part of its health check after waiting for the web server to become responsive and the daemons to have completed their potentially lengthy startup!  As you might have inferred, I had been thinking we'd run the check on the web server and consult the filesystem there too, but the SERVER_SYMBOL_SEARCH_URL check depends on the very long process of the web servers starting up, and I'd sort of given up after realizing that would be additional logistical work.  (In the case of the local "make build-test-repo" I don't think we have an explicit wait in place, and router.py doesn't start accepting connections until after initializing, but I'd presume between the relatively small crossref database, the fact that we only get to that fetch last, and nginx potentially taking time to get to making that proxy_pass connection, we probably have a fair amount of slack time.  That said, we might need to add a `watch` at some point or something.)

staktrace:
> (In the case of the local "make build-test-repo" I don't think we have an explicit wait in place, and router.py doesn't start accepting connections until after initializing, but I'd presume between the relatively small crossref database, the fact that we only get to that fetch last, and nginx potentially taking time to get to making that proxy_pass connection, we probably have a fair amount of slack time. That said, we might need to add a `watch` at some point or something.)

Ah good point. I didn't run into any problems when testing locally but I agree that there may be a bit of a race there that might bite us later.

PR #389, opened 2020-11-17
Bug 1677731 - androidTest directory should be recognized as test files.

makotokato:
geckoview-junit files are into mobile/android/geckoview/src/androidTest,
but these files isn't recognized as test files.

### DIFF
--- a/router/router.py
+++ b/router/router.py
@@ -124,9 +124,9 @@ class SearchResults(object):
 
     def categorize_path(self, path):
         def is_test(p):
-            # Except /unit/, all other paths contain the substring 'test', so we can exit early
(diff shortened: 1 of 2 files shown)


router/router.py
```diff
@@ -128,6 +128,8 @@ def is_test(p):
             # in case it is not present.
             if '/unit/' in p:
                 return True
+            if '/androidTest/' in p:
```
staktrace, line 131: Please combine this with the previous `/unit/` check, and update the comment on line 127 as well. thanks!

makotokato: Thank you, I updated this.

staktrace (approved):
Perfect, thanks!

PR #390, opened 2020-12-20
Bug 1674601 - Make the blame/coverage strips a bit more touch friendly

staktrace:
Instead of using the touch-generated mouse events to control the blame popup,
add touch event listeners to manage it directly. This patch allows the user
to drag their finger around on the blame strip, and have the blame popup update
to whatever is under the cursor. The touch-action CSS property is used to
disable browser-based panning while the finger is being dragged in such a
manner. This UX makes it somewhat easier to trigger the blame popup for a
given line using touch, where the finger is less precise of a pointer than
a mouse cursor. When triggered via touch, the blame popup remains visible
until it is dismissed by scrolling or "clicking" (i.e. tapping) elsewhere.

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -601,6 +601,8 @@ span[data-symbols].hovered {
     height: 16px;
     padding: 0;
     margin: 0;
+    touch-action: none; /* fallback for FF pre-85 which didn't support touch-action:pinch-zoom */
+    touch-action: pinch-zoom;
 }
 /* blame zebra stripes: we alternate colors whenever the revision a line is from
    changes between lines. */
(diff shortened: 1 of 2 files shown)


staktrace:
https://kats.searchfox.org/mozilla-central/source/gfx/layers/apz/src/APZCTreeManager.cpp can be used to test out the behaviour. It's a web-server from ~45 days ago but has this patch applied manually.

asutherland (approved):
Thanks for the fix and making sure to have a site up for easy testing!  Searchfox is surprisingly usable on Firefox Nightly on Android in landscape mode with this.  (In portrait mode the fundamental lack of screen width is a problem, but the blame popup effective font size also ends up noticeably larger than the source code font size it seems).

PR #391, opened 2020-12-28
Bug 1684371 - Update provisioner to get yq from pip instead of the snap

staktrace:
snaps are evil because they auto-update all by themselves, potentially
breaking downstream users.


asutherland (approved):
Thanks!

PR #393, opened 2020-12-30
Switch build-blame to using git-fast-import

staktrace:
This switches the build-blame tool to write to a git-fast-import stream instead of using the libgit2 API. This improves performance and dramatically reduces temporary disk usage requirements.

### DIFF
--- a/tools/src/bin/build-blame.rs
+++ b/tools/src/bin/build-blame.rs
@@ -9,7 +9,8 @@ extern crate unicode_normalization;
 use std::borrow::Borrow;
 use std::collections::HashMap;
 use std::env;
-use std::io::{BufRead, BufReader, Write};
+use std::fmt;
+use std::io::{BufRead, BufReader, BufWriter, Read, Write};
 use std::path::{Path, PathBuf};
 use std::process::{Child, Command, Stdio};
 use std::sync::mpsc::{channel, Receiver, Sender};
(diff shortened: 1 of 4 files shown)


tools/src/bin/build-blame.rs
```diff
@@ -659,78 +742,88 @@ fn main() {
...
+                write!(
+                    import_stream,
+                    "from 0000000000000000000000000000000000000000\n"
+                )
+                .unwrap();
+            }
+            for additional_parent in blame_parents.iter().skip(1) {
+                write!(import_stream, "merge {}\n", additional_parent).unwrap();
+            }
+            write!(import_stream, "deleteall\n").unwrap();
```
asutherland, line 796: Potential comment: Start from a clean slate for simplicity. https://git-scm.com/docs/git-fast-import#_filedeleteall

asutherland (approved):
Very clean and extremely well documented!  I'm very excited at the prospect of all the fancy new history-aware features this massively improved pipeline will allow thanks to these changes and your other recent enhancements!

I've added a bunch of nit-level comment suggestions to link to the relevant docs simply because I found myself consulting the docs as I read the diff and I figured the links might be handy, but if you think they impair readability or some familiarity with the docs should be presumed anyways, I have no problem if they don't get added.

staktrace:
Thanks! I added the comments as you indicated, with slight wording changes. I didn't actually do a full indexer run with these changes so it's possible these still run into the git2 upgrade problem you were seeing in #367. I'll try that now and verify nothing breaks.

staktrace:
Verified the git2 upgrade doesn't appear to break anything. I'll be around tomorrow to keep an eye out for breakages so I'll go ahead and merge this now.

PR #395, opened 2021-01-07
bug 1584120 - Remove the link to dxr

sylvestre:


static/js/dxr.js
```diff
@@ -1,553 +0,0 @@
...
-      }
-    }
-  }
-
-  var fileCount = 0;
-  for (var pathkind in data) {
-    for (var qkind in data[pathkind]) {
-      fileCount += data[pathkind][qkind].length;
-    }
-  }
```
emilio, line 392: Why did you remove this? This will completely break searchfox lol.

tools/src/output.rs
```diff
@@ -271,7 +271,6 @@ pub fn generate_footer(
     }
 
     let scripts = [
-        "dxr.js",
```
emilio, line 274: Same here, did you test these changes? `dxr.js` was originally from dxr, though maybe we should rename it to `search.js` or something? It was heavily changed.

tools/src/config.rs
```diff
@@ -20,7 +20,6 @@ pub struct TreeConfigPaths {
     pub git_blame_path: Option<String>,
     pub objdir_path: String,
     pub hg_root: Option<String>,
-    pub dxr_root: Option<String>,
```
emilio, line 23: These changes are fine.

emilio (changes requested):

static/js/dxr.js
sylvestre: oups, I went full in :)
I will rename this script to make it obvious that it isn't dxr.

asutherland:
This will also want a patch to https://github.com/mozsearch/mozsearch-mozilla to remove the "dxr_root" fields from the 4 configN.json files.

asutherland:
Also, @sylvestre can you confirm if you ran this locally in a Vagrant VM or is this still done by inspection?

sylvestre:
@asutherland  https://github.com/mozsearch/mozsearch-mozilla/pull/134

scripts/output.js
```diff
@@ -122,7 +122,7 @@ function generate(content, opt)
   <span id="data" data-root="/" data-search="/${tree}/search" data-tree="${tree}"></span>
   <span id="state" data-offset="${stateOffset}" data-limit="${stateLimit}" data-result-count="${resultCount}" data-eof="${eof}"></span>
 
-  <script src="/static/js/dxr.js"></script>
+  <script src="/static/js/search.js"></script>
```
staktrace, line 125: Note that when this merges we'll likely need to retrigger the config3.json indexer (manually) since that will have generated files still linking to `dxr.js`

staktrace:
@sylvestre plz ignore the failure email you got. I'm testing out this patch but (a) had some stale stuff in my push causing the failure and (b) forgot to add a dummy commit so that the results email comes to me

staktrace (approved):
I tested the patches and they look good. I'll r+ but still merge a bit later so I can schedule deployment (see my previous comment)

staktrace:
I symlinked dxr.js to search.js on the latest release1 web-server, and kicked off new indexing runs for release2, release3, and release4 with this PR applied. So those should all deploy with references to search.js which will resolve using the symlink. And then when the next regularly-scheduled release1 run happens the web-server from that will have search.js instead of dxr.js and everything should be migrated happily. There's a chance that the release2 run I just kicked off won't be done before the next release1 is done, in which case those repos might be temporarily broken.

asutherland:
Thanks all!

PR #396, opened 2021-01-09
Small context menu event handling fix

buttercookie42:
Curiously enough this works both in current-ish Firefoxes and Chrome, but is it really legal JS?


staktrace (approved):
Thanks! The code works because of https://developer.mozilla.org/en-US/docs/Web/API/Window/event but the change makes the code better.

PR #397, opened 2021-01-17
Bug 1687185 - Disable daily cron jobs. r=kats

asutherland:
updatedb can end up tying up the indexing partition in a bad way, so disable
it and all the other daily and weekly cron jobs that we weren't actually
running until we just recently started trying to run jobs for the 22:00 UTC
mozilla-central nightly.


asutherland:
I haven't run an indexer job with this, but I did cut and paste the operative command on an actively running indexer and it didn't explode.

The rationale for this choice is:
- chmod is non-destructive but accomplishes the goal, so we can see what we disabled.
- This avoids messing with the vagrant VMs
- I don't have to re-provision anything!

Open to further alternative approaches!

staktrace (approved):
LGTM!

PR #399, opened 2021-02-17
Bug 1693352 - Increase web-server.rs thread count to 4.

asutherland:
This also adds $request_time to the nginx HTTP log output which indicates how
long the request took in seconds with a millisecond resolution.

I checked locally that I can see the extra logging and that the web-server process goes from 4 threads to 6 threads.  I'm assuming the other threads are interesting support threads that don't feel the name to give themselves interesting names.

### DIFF
--- a/scripts/nginx-setup.py
+++ b/scripts/nginx-setup.py
@@ -118,7 +118,7 @@ if nginx_cache_dir:
     print('')
 
 print('''# we are in the "http" context here.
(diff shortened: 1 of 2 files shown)


asutherland:
I did briefly look at adding the duration of the rust web-server.rs responses from inside itself, but figured that it's probably better to be looking at the full access log with latencies across all responses, as that could also show the machine experiencing paging hiccups or other problems.

emilio (approved):
Looks reasonable to me, thanks!

PR #400, opened 2021-02-22
Bug 1694210 - Update VMs to Ubuntu 20.04 LTS

asutherland:
The primary motivation is to get a newer version of curl with relevant
bugfixes.  The secondary motivation is that our lives are made simpler
by staying up to date with releases.

The only necessary change for the Vagrant VM was to remove the GNU
parallel citation workaround, see
https://bugzilla.mozilla.org/show_bug.cgi?id=1694210#c1

Other notes not part of the commit message:
- I've only run this locally on Vagrant libvirt so far.  I would ideally like to merge this and then do the provisioning so I can provision from master rather than on my own branch.
  - I did look at switching to official Ubuntu images for vagrant, but if you run the search https://app.vagrantup.com/boxes/search?utf8=%E2%9C%93&sort=downloads&provider=&q=ubuntu+20.04 you'll find that the official "ubuntu/focal64" only has a virtualbox image and no libvirt (or hyperv or parallels or vmware_desktop) images.
- I may need to iterate on the provisioning, but I figure that will all be trivial given the local `vagrant destroy; vagrant up` and the `make build-test-repo` and `make build-searchfox-repo` things checked out fine.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,6 +1,6 @@
 Vagrant.configure("2") do |config|
-  config.vm.box = "generic/ubuntu1804"
-  config.vm.box_version = "2.0.6"
+  config.vm.box = "generic/ubuntu2004"
+  config.vm.box_version = "3.2.6"
 
   config.vm.provision :shell, privileged: false, path: "infrastructure/vagrant/indexer-provision.sh"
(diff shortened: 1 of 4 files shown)


emilio (approved):
Looks sensible, thanks!

PR #401, opened 2021-02-22
Bug 1694210 - Update AMI image name refs, use python2 for logs

asutherland:
The VM triggering scripts need to know the revised AMI name that we're
going to use.  Also, during provisioning, the awslogs-agent-setup.py
script failed to execute properly because the embedded hash-bang was
for "python", so we now explicitly run it with python2 because our
python3 is too new.

After this I'll also regenerate the lambda jobs.

### DIFF
--- a/docs/aws.md
+++ b/docs/aws.md
@@ -286,7 +286,7 @@ new target group.
 
 The EC2 instances for indexing and web serving are started using a
 custom Amazon Machine Image (AMI). This is the disk image used for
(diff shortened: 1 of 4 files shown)


asutherland:
I'm going to merge this as this is all the expected flow from https://github.com/mozsearch/mozsearch/pull/400 including the minor fix-up and AMI name changes (which I should have called out in the previous commit as planned, however).

PR #403, opened 2021-02-24
Bug 1637010 - Fix how the scroll position is computed for position:sticky line clicks

kashav:
The nesting container is rooted at the stuck line's original offset, so
scrolling up by the difference between the two correctly jumps to the line
number that was clicked.


asutherland (approved):
YAYYYYY!  Thank you!  I've applied this to the current mozsearch web-server and it works great!  Wooooooo!

PR #404, opened 2021-02-26
Bug 1683563 - Make it easier to rebuild the blame tarballs [requires lambda update with merging!]

staktrace:


Makefile
```diff
@@ -60,6 +60,12 @@ trypush: check-in-vagrant build-clang-plugin build-rust-tools
 	/vagrant/infrastructure/web-server-setup.sh ~/mozilla-config trypush.json ~/trypush-index ~
 	/vagrant/infrastructure/web-server-run.sh ~/mozilla-config ~/trypush-index ~
 
+nss-reblame: check-in-vagrant build-rust-tools
+	[ -d ~/mozilla-config ] || git clone https://github.com/mozsearch/mozsearch-mozilla ~/mozilla-config
```
asutherland, line 64: I've never been happier or sadder to know that `[` is actually a command!  (Which is to say, I yearn for the simpler days when I would have gone: "what on earth is going on here?!?".  This makes sense though and is something it makes sense to presume contributors will learn quickly if they don't already know.)

infrastructure/aws/index.sh
```diff
@@ -110,7 +92,7 @@ esac
 $AWS_ROOT/send-warning-email.py "[$CHANNEL/$BRANCH]" "$DEST_EMAIL"
 
 gzip -k ~ubuntu/index-log
-$AWS_ROOT/upload.py ~ubuntu/index-log.gz indexer-logs "$(date -Iminutes)_${CHANNEL}_${CONFIG_INPUT%.*}"
+$AWS_ROOT/upload.py ~ubuntu/index-log.gz indexer-logs "index-$(date -Iminutes)_${CHANNEL}_${CONFIG_INPUT%.*}.gz"
```
asutherland, line 95: Nice enhancement!

infrastructure/aws/main.sh
```diff
@@ -22,7 +22,9 @@ handle_error() {
     # was. We only do this if we got far enough to actually start indexing.
     if [ -d "/index" ]; then
         mkdir -p /index/interrupted
-        mv -f /mnt/index-scratch/* /index/interrupted
+        if [ -d "/mnt/index-scratch" ]; then
+            mv -f /mnt/index-scratch/* /index/interrupted
+        fi
```
asutherland, line 27: Also nice!  I won't miss these errors/warnings!

infrastructure/aws/main.sh
```diff
@@ -41,15 +43,24 @@ handle_error() {
     exit 1
 }
 
-# See index.sh for the arguments to this script
+# Pull out the first two arguments, which are for consumption by this
+# main.sh script. The rest of the arguments get passed as arguments to
+# TARGETSCRIPT, so we leave them in $*
+TARGETSCRIPT=$1; shift
+MAXHOURS=$1; shift
```
asutherland, line 50: Also^2 Nice!

infrastructure/aws/main.sh
```diff
@@ -68,10 +79,26 @@ cat > ~/.aws/config <<"STOP"
...
+echo "Disabling daily and weekly cron jobs for this indexing run"
+sudo chmod -x /etc/cron.daily/* /etc/cron.weekly/*
+
+echo "Creating index-scratch on local instance SSD"
+${AWS_ROOT}/mkscratch.sh
 
-# Run indexer with arguments supplied to this script.
-$AWS_ROOT/index.sh $*
+# Run target script with arguments supplied to this script.
+${AWS_ROOT}/${TARGETSCRIPT} $*
```
asutherland, line 104: Also^3 nice!

infrastructure/aws/trigger_blame_rebuild.py
```diff
@@ -0,0 +1,77 @@
...
+
+def trigger(mozsearch_repo, config_repo, config_input, branch, channel):
+    ec2 = boto3.resource('ec2')
+    client = boto3.client('ec2')
+
+    running = ec2.instances.filter(Filters=[{'Name': 'tag-key', 'Values': ['blame-builder']},
+                                           {'Name': 'tag:channel', 'Values': [channel]},
+                                           {'Name': 'instance-state-name', 'Values': ['running']}])
+    for instance in running:
+        print("Terminating existing running indexer %s for channel %s" % (instance.instance_id, channel))
```
asutherland, line 19: nit: s/indexer/blamer/ (or blame-builder ;) but fine to leave as-is

infrastructure/aws/trigger_blame_rebuild.py
```diff
@@ -0,0 +1,77 @@
...
+        instance.terminate()
+
+    user_data = '''#!/usr/bin/env bash
+
+cd ~ubuntu
+sudo -i -u ubuntu ./update.sh "{branch}" "{mozsearch_repo}" "{config_repo}"
+sudo -i -u ubuntu mozsearch/infrastructure/aws/main.sh rebuild-blame.sh 168 "{branch}" "{channel}" config "{config_input}"
```
asutherland, line 26: The 168 is less intuitive here than the 10 for the indexer.  I think it would be great to have a comment in the script here or have the value extracted out into a constant with a comment.  In particular, I think it would be great for the comment to capture how close we are to 168 hours at the time of writing and what date that was written.  (I realize there may be the caveat that maybe you're just doing this locally and haven't run on EC2?)

infrastructure/aws/trigger_blame_rebuild.py
```diff
@@ -0,0 +1,77 @@
...
+'''.format(branch=branch, channel=channel, mozsearch_repo=mozsearch_repo, config_repo=config_repo, config_input=config_input)
+
+    block_devices = []
+
+    images = client.describe_images(Filters=[{'Name': 'name', 'Values': ['indexer-18.04']}])
```
asutherland, line 31: s/18.04/20.04/ (and maybe we should delete or rename the 18.04 AMI's now?  I left them around for faster rollback.)

asutherland (approved):
This looks fantastic but needs the s/18.04/20.04/ fix.  Thank you so much for your continued efforts to improve blame rebuilding!

infrastructure/aws/trigger_blame_rebuild.py
staktrace: Ah, good point. And yeah I haven't actually done a full run on EC2 yet so I don't know how long it will actually take. So for now maybe I'll land it with the 168 extracted into a variable, and then after I do a full run I can write a better comment.

infrastructure/aws/trigger_blame_rebuild.py
staktrace: Should be safe to delete them now, I think. We've had a few days of running with the new ones and no issues seem to have cropped up.

asutherland: I deleted the 18.04 AMI's so now we're down to just the web+indexer 20.04 AMI's.

PR #405, opened 2021-02-26
Bug 1695242 - Don't hard-code blame strip box heights

emilio:
16px is not guaranteed to fill the line (and it fact it does not if you use a big font-size, like webkit-search.igalia.com, or if you're zoomed in in some platforms).


asutherland (approved):
Thanks!

PR #406, opened 2021-03-01
code-hightlighter: Ensure we account for float height when computing whether something is stuck.

emilio:
Otherwise rounding error accumulates and random stuff happens.


asutherland (approved):
Thanks!

PR #407, opened 2021-03-13
Bug 1698236 - Use clang 11 and build c++17-style

asutherland:
I tested this with `vagrant destroy` followed by `vagrant up` followed by (in the VM via `vagrant ssh`) `cd /vagrant; make build-test-repo`.  Note that I did experience weirdness with a missing symbol on the ItaniumMangleContext during clang analysis which I believe ended up being due to the "make" in /vagrant/clang-plugin deciding not to actually rebuild things under clang 11 because we didn't express a dependency on the compiler and those object files/the binary existing outside of the VM itself because we don't use a bindir.  So if you're testing and not using a fresh checkout, you may need to first do a "make clean" in /vagrant/clang-plugin.

I'll plan to re-provision the indexer once this lands even though it hopefully doesn't matter.

### DIFF
--- a/infrastructure/indexer-provision.sh
+++ b/infrastructure/indexer-provision.sh
@@ -8,10 +8,19 @@ set -o pipefail # Check all commands in a pipeline
 # will be used by the Firefox build process.  If you have a "mach bootstrap"ped
 # system then you can see the current version locally via
 # "~/.mozbuild/clang/bin/clang --version"
-CLANG_VERSION=9
+#
+# Note that for the most recent LLVM/clang release (ex: right now v13), you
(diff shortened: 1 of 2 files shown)


emilio (approved):

staktrace (approved):
The only other thing that comes to mind to watch out for is that we actually build the `nss` repo on the indexer, so we should check that it doesn't fail compilation with the new clang. I assume it doesn't but ya never know!

asutherland:
Thanks for raising the NSS potential issue; I manually triggered a config1 indexer and there were no errors and analysis data seems to have been successfully generated.

PR #409, opened 2021-03-17
Bug 1558691 - Make the search result headers headings

asutherland:
https://bugzilla.mozilla.org/show_bug.cgi?id=1558691 is a follow-up to https://bugzilla.mozilla.org/show_bug.cgi?id=1511279 and its https://github.com/mozsearch/mozsearch/pull/185 

I added some extra comments in the loop as I was getting the lay of the land because I don't find our "pathkind" and "qkind" terms self-explanatory.  For hand-waving purposes, my rationale for going with the span proposal over a proper h3 is that I expect we might overhaul this all in the near future.

### DIFF
--- a/static/js/search.js
+++ b/static/js/search.js
@@ -302,7 +302,12 @@ function populateResults(data, full, jumpToSingle) {
       chooseIcon(fileResult.path) +
       " icon-container'></div></td>";
 
-    html += "<td>";
+    // This span exists for a11y reasons.  See bug 1558691 but the core idea is:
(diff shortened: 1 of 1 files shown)


asutherland:
I've re-tagged an existing backup release1 web-server into the "dev" channel, put it in the "dev" target group, and checked out this branch.  https://dev.searchfox.org/mozilla-central/search?path=&q=ServiceWorkerPrivate and https://dev.searchfox.org/ should now be a live example of this patch (since it was only runtime JS stuff).

staktrace (approved):

PR #412, opened 2021-03-30
Bug 1701848 - Add script to aid in analyzing searchfox web-server dynamic query performance

asutherland:
Example of output can be found in the bug.

This depends on https://github.com/rcoh/angle-grinder which can be installed
via `cargo install ag`.  My current assumption is that this is being run
locally, not in a VM, by someone who has already done this.

I'm not sure I have a longer term strategy for this, I just wanted to archive
the results of my investigation in a way that's slightly better than searching
for a bugzilla comment and then copying and pasting from that.

### DIFF
--- a/scripts/weblog-analyze.sh
+++ b/scripts/weblog-analyze.sh
@@ -1,0 +1,76 @@
+#!/bin/bash
+
+# Arguments: <log path>
+
+# This script uses https://github.com/rcoh/angle-grinder in order to process our
+# searchfox logs from /var/log/nginx/searchfox.log
+
+# We don't want to show commands, so no: set -x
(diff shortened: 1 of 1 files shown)


asutherland:
In terms of https://github.com/rcoh/angle-grinder I found it fun/easy to get something going initially, but trying to deal with the optionally present search query got ugly fast.  In the end I ended up just forking those paths.  The tool seems very promising though in terms of:
- Written in rust!
- Lots of tests, both inline and "structured"
- Seems quite reasonably extensible
- It does a pretty cool realtime UI on the console as it processes stuff.

staktrace (approved):

PR #413, opened 2021-04-01
Bug 1702319 - Update title based on last selected line or current sticky line.

asutherland:
https://asuth.searchfox.org/ has this using a config1 spare server from yesterday.

### DIFF
--- a/static/js/code-highlighter.js
+++ b/static/js/code-highlighter.js
@@ -8,6 +8,232 @@
  * 3) Highlight lines when page loads, if window.location.hash exists
  */
 
+/**
+ * Decides what the document title should be.
+ */
+var DocumentTitler = new (class DocumentTitler {
+  constructor() {
+    this.originalTitle = document.title;
+    this.currentTitle = this.originalTitle;
+
+    this.stickyTitle = null;
+    this.selectionTitle = null;
+  }
+
+  updateTitle() {
+    let bestTitle;
+
+    // If we have a better title than the original title, use it, but making
+    // sure to include the original filename because this is important for
+    // finding the tab again in the awesomebar via its filename.
+    if (this.selectionTitle) {
+      bestTitle = `${this.selectionTitle} (${this.originalTitle})`;
+    } else if (this.stickyTitle) {
+      bestTitle = `${this.stickyTitle} (${this.originalTitle})`;
+    } else {
+      bestTitle = this.originalTitle;
+    }
+
+    // Debounce setting the title out of a fear of slowing down scrolling but
+    // not wanting to to do the legwork to figure out if this would pose a
+    // problem.  I am assuming this shouldn't result in synchronous reflows.
+    if (bestTitle === this.currentTitle) {
+      return;
+    }
+    document.title = this.currentTitle = bestTitle;
+  }
+
+  /**
+   * Recognizing that namespaces can be a little verbose and use up precious
+   * space, flatten things that look like namespaces to be a single-character
+   * delimited by a single colon.
+   *
+   *
+   * Good example transforms:
+   * - `Foo` => `Foo`
+   * - `Foo::Bar` => `Foo::Bar`
+   * - `mozilla::Foo::Bar` => `m:Foo::Bar`
+   * - `mozilla::dom::quota::Foo::Bar` => `m:d:q:Foo::Bar`
+   *
+   * Sketchy example transforms:
+   * - `mozilla::Foo` => `m:Foo`
+   * - `mozilla::dom::Foo` => `m:d:Foo`
+   *
+   * The sketchy transforms are sketchy because we're assuming that lowercase is
+   * indicative of a namespace.  Note that, similarly to the comments in
+   * `_findBestPrettySymbolInSourceLineElem`, this heuristic should simply be
+   * mooted by having the symbol dictionary provide a pre-computed concise
+   * pretty name for a symbol.  That mechanism can leverage knowing what is and
+   * isn't a namespace/class and also having global knowledge of the names that
+   * are in the codebase so that additional tokens can be used in cases where
+   * ambiguity exists, etc.
+   */
+  _shortenNamespaces(pretty) {
+    const pieces = pretty.split("::");
+    // Nothing to do if we don't have anything to split.
+    if (pieces.length < 2) {
+      return pretty;
+    }
+
+    // We want to use the last two components if they are both initial-caps,
+    // but not for the "mozilla::Foo" case, or "mozilla::dom::Foo" case, where
+    // we still want to collapse the lower-cased namespace.  In that case, we
+    // collapse everything but the last piece.
+    let splitPoint;
+    // Our regexp assumes it's a namespace if it's:
+    // - All ASCII lowercase.
+    // - And therefore has no underscores (which helps avoid us getting tricked
+    //   by nested functions named_like_this in some hypothetical world where
+    //   we understand Python).
+    if (/^[a-z]+$/.test(pieces[pieces.length - 2])) {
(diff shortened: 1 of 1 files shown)


static/js/code-highlighter.js
```diff
@@ -8,6 +8,198 @@
...
+    const pieces = pretty.split("::");
+    // Nothing to do if we don't have anything to split.
+    if (pieces.length < 2) {
+      return pretty;
+    }
+
+    // We want to use the last two components if they are both initial-caps,
+    // but not for the "mozilla::Foo" case, or "mozilla::dom::Foo" case, where
+    // we still want to collapse the lower-cased namespace.  In that case, we
+    // collapse everything but the last piece.
```
emilio, line 81: I think I'd drop the collapsing of the `mozilla::` in `mozilla::Foo` (or at least special-case that instead).

Otherwise you incorrectly collapse classes that start with a lower-case letter when you're inside a method, which is very common in older code (I was trying e.g., `nsPresContext` in your instance).

static/js/code-highlighter.js
```diff
@@ -8,6 +8,198 @@
...
+    }
+
+    const nsPieces = pieces.slice(0, splitPoint);
+    const fullPieces = pieces.slice(splitPoint);
+
+    const nsTransformed = nsPieces.map((piece) => {
+      return piece[0];
+    });
+
+    if (nsTransformed.length === 0) {
```
emilio, line 96: I'd write this as `if (!nsTransformed.length)`, but no big deal either way.

static/js/code-highlighter.js
```diff
@@ -8,6 +8,198 @@
...
+    if (/^[a-z]/.test(pieces[pieces.length - 2])) {
+      splitPoint = -1;
+    } else {
+      splitPoint = -2;
+    }
+
+    const nsPieces = pieces.slice(0, splitPoint);
+    const fullPieces = pieces.slice(splitPoint);
+
+    const nsTransformed = nsPieces.map((piece) => {
```
emilio, line 92: nit: You can technically omit the parenthesis here, so just `piece => {`.

emilio (approved):
Other than that this looks really good!

static/js/code-highlighter.js
asutherland: @emilio Ah, yeah, I see I already wrote a (better) check like this before in the fancy branch that has a [regexp /^ns|moz/](https://github.com/asutherland/mozsearch/blob/188fd265611efbe01fbc1b373551f067b5a35929/ui/src/grokysis/frontend/kb/symbol_info.js#L10) to decide things aren't namespaces.  But maybe a better hacky approach is checking that the entirety of the piece is 1) lowercase and 2) doesn't include underscores.

asutherland: My specific motivation here was for "mozilla::dom::indexeddb::HelperFunction" where "indexeddb" wasn't getting collapsed and was relatively long.  I could also probably just try and gather up the finite set of namespaces we have in the codebase, but I'm a little worried about people filing a bunch of bugs like "add my namespace to your big hacky regexp" when the real fix involves the structured records landing and some follow-ons.

emilio: Yeah, checking that it's all lowercase seems like a reasonable compromise. We do have some mixed-case namespaces like `Foo_Bindings` etc for the DOM, but it's less common to explore autogenerated code I think.

static/js/code-highlighter.js
asutherland: > I'd write this as `if (!nsTransformed.length)`, but no big deal either way.

This is a stylistic thing I've been doing (after reading a thing I no longer remember where I read it) where the explicit zero check feels more explicit/literate/in-domain even if it ends up equivalent to the negation.  I will sometimes just do the negation when it's more of a falsey check for whether optional arguments were provided, etc.

That said, I'm definitely amenable to conforming to prettier/eslint rules that cover this, especially if we check in a vscode config.  I will leave this as-is for now.

static/js/code-highlighter.js
asutherland: Implemented, thanks!

asutherland:
Checking out the `nsPresContext` use-case, I added another "stop looking for symbols" heuristic to handle the case `class nsPresContext : public nsISupports, public mozilla::SupportsWeakPtr {` where the heuristic was picking and displaying `m:SupportsWeakPtr`.  We now stop looking if we've seen both "class" and " : ".

asutherland:
And I added one final change to the heuristics to deal with the situation for the [constructor](https://asuth.searchfox.org/mozilla-central/rev/9043e515e9608cc55b252a40cb2dfb6f767bcffd/dom/workers/WorkerPrivate.cpp#1087) of `class WorkerPrivate::MemoryReporter final : public nsIMemoryReporter`.  In this case our searches list had `nsIMemoryReporter::nsIMemoryReporter` for the 2nd to last search and the actual `MemoryReporter::MemoryReporter` that we wanted as the last one.  My use of `find(...)` looking for the `MemoryReporter` token in the search, which ended up being true for the interface.  I've addressed this by having us traverse the entire list of searches, updating our choice potentially every time, so we'll end up picking MemoryReporter, as can be seen now [here again this is on asuth.searchfox.org](https://asuth.searchfox.org/mozilla-central/rev/9043e515e9608cc55b252a40cb2dfb6f767bcffd/dom/workers/WorkerPrivate.cpp#1087).

I did run with these changes using a recent Firefox nightly on Win10 using the most recent NVDA with default-ish settings because I was worried it might report every document title change, which would be spammy.  It did not report every document title change, so I think the change is neutral or even an improvement for accessibility purposes given that we're improving tab titles.

I'm going to land this now so it shows up in the UTC22 nightly.  I'm very happy to further iterate on this in response to feedback from :kats in a follow-up PR, etc.

static/js/code-highlighter.js
```diff
@@ -8,6 +8,232 @@
...
+        }
+      }
+
+      // Extract the most appropriate pretty data from the searches.
+      // Specifically, we are looking for "pretty" text in the searches that
+      // contains the textContent from the semantic token.  We do this to
+      // compensate for the implicitly invoked field constructors which
+      // currently end up coalesced into the constructor's symbol/point.
+      const visibleToken = symElem.textContent;
+      const data = window.ANALYSIS_DATA[symElem.getAttribute("data-i")];
```
staktrace, line 166: Are there any semantic differences between reading the attribute and reading `.dataset.i`? Just wondering, since I generally prefer the latter as it fails sooner/better.

staktrace (commented):
Patch looks good to me!

static/js/code-highlighter.js
asutherland: Honestly, I had the same question but was cargo-culting from [code that already was using getAttribute](https://github.com/mozsearch/mozsearch/blob/989908c990abeee6a68df7dc0daa273c135eef81/static/js/context-menu.js#L48).  I also was a little perturbed by the lack of explicit parsing into an integer, but took what I'll call the "shrug, YOLO JS" approach where it clearly worked and certainly was terse and explicit (and if arrays are going to coerce as needed, why not).

rs=asuth for sure on comprehensive cleanup of this idiom that stops me from cargo culting bad stuff in the future ;)

NeptuneUlt (commented):
> https://github.com/mozsearch/mozsearch/blob/dynamic-title- - [ ] /static%2Fjs%2F**_[]()@ _**code-highlighter.js ``~~~~

NeptuneUlt (commented):
```@
Y ```

```
- **Echoes* ******

Meme19881401:
Hello,Word

PR #414, opened 2021-04-04
CSS cleanup

emilio:
I went through our CSS to see how much work is really adding a dark theme to searchfox. Turns out we can remove a lot of CSS and simplify a lot more.

See individual commits for details.


staktrace (approved):
Awesome, thank you for all this cleanup!

asutherland (approved):
Woo!  Yes, thank you for this excellent cleanup and the very well factored patch stack!

PR #415, opened 2021-04-05
WIP dark-mode theming

mystor:
VSCode really wanted to format the CSS with prettier based on the `.prettierrc`, so the first patch in this series is reformatting everything with that tool. It'll be easier to see the specific changes by looking at the second patch.

This patch still has some remaining issues, such as issues with the ccov bar's colouring, less accessible diff colouring, and a super janky setup for the copy button, so isn't quite ready to merge yet.

### DIFF
--- a/scripts/output.js
+++ b/scripts/output.js
@@ -69,6 +69,7 @@ function generate(content, opt)
   <link href="/static/icons/search.png" rel="shortcut icon">
   <title>${title}</title>
 
+  <link href="/static/css/theme.css" rel="stylesheet" media="screen"/>
   <link href="/static/css/mozsearch.css" rel="stylesheet" media="screen"/>
   <link href="/static/css/icons.css" rel="stylesheet" media="screen" />
 </head>
--- a/static/css/icons.css
+++ b/static/css/icons.css
@@ -75,11 +75,12 @@ a.type {
 
 .panel .accel,
 .panel .copy {
-  font-size: 10px;
-  border-radius: 2px;
-  padding: 1px 3px;
-  margin: 0 10px;
-  border: 1px solid #ddd;
+  background-color: var(--toolbarbutton-background);
+  color: var(--toolbarbutton-color);
+  box-sizing: content-box;
+  border: 1px solid var(--theme-toolbar-separator);
+  height: 1em;
+  padding: 3px;
 }
 
 .panel .accel {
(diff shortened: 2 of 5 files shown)


asutherland:
This can now be seen on https://dev.searchfox.org/ and we should be able to update the git checkout as the CSS changes.

One immediate thing I noticed was that in the non-dark-mode case, I miss the gradient on the search box.  Not opposed to doing away with it for cause, but it seems like we should retain the existing non-dark-mode visual look unless there's intent to change.

mystor:
The main reason I also changed the non-darkmode theme a bit with the refactoring was to keep as much of the theming logic out of the specific components' CSS as possible. In general, I tried to use css variables from the imported devtools theme over defining new colours, which means that some colours from the normal theme were definitely impacted. 

For something significant like the gradient on the searchbox we can definitely add in a special case, though. If you notice other areas where the colours were tweaked which you don't like, lmk.

asutherland:
Gotcha.  I just wanted to make sure we weren't losing it accidentally because the focus was on the dark mode.  I personally have enjoyed the gradient as "feels premium" hack like gluing weights inside physical products to make them feel hefty, but if kats and emilio don't share such nostalgia, I'm fine with losing that.

In general, my primary concern for colors is just ensuring that we don't create contrast problems, colorblindness problems, or legibility problems.  The improvement in contrast in the search header that describes what revision is indexed for example is good, even if it means we might want to rethink showing the largely useless commit message and maybe even move that information elsewhere.

I think the one coloration concern I have is over the change to the (light mode) position:sticky'd lines might make it harder to realize that the stuck lines are not continuous with the actually displayed source (which falls in the legibility bucket).  This problem came up in my original styling of the position:sticky implementation; relevant comments on that are in [Bug 1557139](https://bugzilla.mozilla.org/show_bug.cgi?id=1557139) from [:mccr8](https://bugzilla.mozilla.org/show_bug.cgi?id=1557139#c0) and [:smaug](https://bugzilla.mozilla.org/show_bug.cgi?id=1557139#c6).

mystor:
> Gotcha. I just wanted to make sure we weren't losing it accidentally because the focus was on the dark mode. I personally have enjoyed the gradient as "feels premium" hack like gluing weights inside physical products to make them feel hefty, but if kats and emilio don't share such nostalgia, I'm fine with losing that.

it also wouldn't be a big deal at all to bring the gradient back. The main reason it's missing is that I don't personally care much about it and it would've taken ever so slightly more effort.

> In general, my primary concern for colors is just ensuring that we don't create contrast problems, colorblindness problems, or legibility problems. 

This is also one of my major concerns. With some colours like the diff colours, I made special effort to keep the old colours around in light mode because the light devtools colours weren't as readable as the dark/our normal ones. We should still probably change the colours in dark mode to not be red/green (for colourblind folks, e.g. https://dev.searchfox.org/mozilla-central/diff/f722be009c02f0aa46c03dc0d99237ea38dc06ac/docshell/base/BrowsingContext.h#199), but I am no good at picking colours.

> The improvement in contrast in the search header that describes what revision is indexed for example is good, even if it means we might want to rethink showing the largely useless commit message and maybe even move that information elsewhere.

I suppose it could be moved into the navigation box on the right? I don't think that needs to be done in this patch.

> I think the one coloration concern I have is over the change to the (light mode) position:sticky'd lines might make it harder to realize that the stuck lines are not continuous with the actually displayed source (which falls in the legibility bucket).

Do you think the dark mode version is fine as-is? We can definitely switch the light mode version back to the old colour, but I'm not the best at coming up with new colours which fit with the devtools colourscheme, so I've generally been avoiding it :-P

asutherland:
>> The improvement in contrast in the search header that describes what revision is indexed for example is good, even if it means we might want to rethink showing the largely useless commit message and maybe even move that information elsewhere.
>
> I suppose it could be moved into the navigation box on the right? I don't think that needs to be done in this patch.

Yeah, absolutely doesn't need to (/ shouldn't) happen in this patch.  Just meant to convey that's an example of something where correcting the colors is the correct thing to do and then we can iterate from there on the fallout.

> Do you think the dark mode version is fine as-is? We can definitely switch the light mode version back to the old colour, but I'm not the best at coming up with new colours which fit with the devtools colourscheme, so I've generally been avoiding it :-P

I think the dark mode background color probably has enough contrast, but I think we're missing the delineating effect of the box-shadow on `.last-stuck` which is still black in the dark mode theme.  I randomly used the devtools color picker to change the shadow to `#ffffff82` and that sorta works?  Like, it's not great, but the delineating effect is good and I think demonstrates we want some variation on it.  My mental model of what we're doing for this is like a ground effect light from cars in a fast and the furious movie, so maybe a color is better than just some white/gray.

Specifically in terms of the not great for my choice, the current interaction of me doing that with the coverage coloring seems to create some kind of optical illusion contradiction that hurts my brain, but maybe if we change the coverage colors to be darker variants that will get better?  To address that, from the 11-color variation from https://colorbrewer2.org/#type=diverging&scheme=RdBu&n=11 we could jump to the darker red (#67001f or #b2182b)/blue (#053061 or #2166ac) variants instead of the lighter ones.

emilio:
Let's close given https://github.com/mozsearch/mozsearch/pull/464 landed. Thanks @mystor!

PR #416, opened 2021-04-05
Format css with prettier

mystor:


emilio (approved):
Looks fine :)

Not a fan of gradients in multiple lines etc, but no big deal, the more consistent the better.

PR #417, opened 2021-04-12
Bug 1414954 - Add secret codesearch support for getting lines of context.

asutherland:
This implements an undocumented "context:N" search directive understood by
the parser (if it precedes the search query proper, as-is required by the
implementation of parse_search) that we propagate to livegrep which in turn
generously returns the context for us.

I'm calling this secret because I don't think the UX as provided by this
patch is appropriate for a supported feature, but I think this could be
useful for experimentation and iteration.  In particular, in this patch:
- Typing out context:N at the start of the search but having things not work
  if you put it at the end as an afterthought isn't great.
- I don't think adding a separate form input for lines of context is a great
  initial choice either.
- This doesn't work with any of the semantic responses, and doing so is both
  non-trivial and not something I think we should be freshly implementing in
  Python.
- On the actual result line where we have bounds, we normalize off leading
  whitespace whereas we leave the whitespace intact on the context lines and
  this looks weird.

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -274,6 +274,17 @@ body {
 }
 /* Search results */
 
+/* Make context lines more subtle. */
+.results tr.before-context-line,
+.results tr.after-context-line {
+  opacity: 0.5;
+}
+
+/* Delineate the transition between consecutive groups. */
+tr.after-context-line + tr.before-context-line {
+  border-top: 1px solid gray;
+}
+
 .results tr:hover,
 .file tr:hover {
   background: none;
--- a/router/codesearch.py
+++ b/router/codesearch.py
@@ -20,16 +20,24 @@ def collateMatches(matches):
         if m.tree == 'mozilla-subrepo':
             path = 'mozilla/' + path
 
-        paths.setdefault(path, []).append({
+        line = {
             'lno': m.line_number,
             'bounds': [m.bounds.left, m.bounds.right],
             'line': m.line
-        })
+        }
+
+        if len(m.context_before):
+            line['context_before'] = list(m.context_before)
+        if len(m.context_after):
+            line['context_after'] = list(m.context_after)
+
+        paths.setdefault(path, []).append(line)
     results = [ {'path': p, 'icon': '', 'lines': paths[p]} for p in paths ]
     return results
 
-def do_search(host, port, pattern, fold_case, file):
-    query = livegrep_pb2.Query(line = pattern, file = file, fold_case = fold_case)
+def do_search(host, port, pattern, fold_case, file, context_lines):
+    query = livegrep_pb2.Query(line = pattern, file = file, fold_case = fold_case,
+                               context_lines = context_lines)
     log('QUERY %s', repr(query).replace('\n', ', '))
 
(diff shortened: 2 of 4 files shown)


asutherland:
This is up on https://asuth.searchfox.org/ and an example search is https://asuth.searchfox.org/mozilla-central/search?q=context%3A5+indexedDB.open&path=browser&case=false&regexp=false

asutherland:
And just one more limitation which is probably obvious:
- No effort is made to coalesce overlapping sibling contexts / merge them into a single run of lines.

static/css/mozsearch.css
```diff
@@ -274,6 +274,17 @@ body {
 }
 /* Search results */
 
+/* Make context lines more subtle. */
+.results tr.before-context-line,
+.results tr.after-context-line {
+  color: gray;
```
emilio, line 280: Maybe just `opacity: .5` would work with dark mode as well?

emilio (approved):
Looks good!

static/css/mozsearch.css
asutherland: Yeah, I think that even works better since it also makes the context line numbers more subtle too.

staktrace:
> example search is https://asuth.searchfox.org/mozilla-central/search?q=context%3A5+indexedDB.open&path=browser&case=false&regexp=false

On this search, the context lines *above* the actual result are shown in reverse order. i.e. if you look at the first hit (ActivityStreamStorage.jsm) what it claims is line 70 in the context is actually line 66.

staktrace (commented):
Patch looks fine, modulo the fix for the bug I mentioned above.

asutherland:
I guess the context_before lines must be provided in reverse order, which could make some level of sense.  I'll compensate for this.

It also turns out that I guess the livegrep server must have always been using a default of 3 lines, including if we specify a context_lines of 0.  I did a bad job of local testing on this... unfortunately I ended up wasting a bunch of time trying to figure out why the `log()` lines weren't working in the forked search processes in the local VM (ex: the QUERY lines), so I rushed the testing somewhat.  (And in the end, I didn't figure out what was going on, but the QUERY lines do show up on the real server, so it's likely something annoying related to the VM.)

asutherland:
Yeah, the [default is 3](https://github.com/livegrep/livegrep/blob/d0148a0fe662d2fc10aebcbfcaac87cbeb8b17e6/src/tools/grpc_server.cc#L31) and [there's logic that explicitly upgrades values that are <= 0](https://github.com/livegrep/livegrep/blob/d0148a0fe662d2fc10aebcbfcaac87cbeb8b17e6/src/tools/grpc_server.cc#L148-L150) to whatever the `FLAGS_context_lines` value is, with [gflags.h from gflags/gflags](https://github.com/gflags/gflags/blob/master/src/gflags.h.in) seeming to provide the relevant macrology.

asutherland:
Okay, I addressed both problems in https://github.com/mozsearch/mozsearch/pull/418 which is up on https://asuth.searchfox.org/ where I validated this.

I'm going to pivot the recent bugged config1 into place after `rm -rf`-ing `~/index/nginx-cache/*` (which will have tainted cached values) and restarting the web-servers.

PR #420, opened 2021-04-28
Bug 1708021 - Advance by word rather than character when looking for …

kashav:
…token position

Advancing by character results in doing things like matching "URI" to the
latter part of "nsIURI", which results in bad column numbers. This fixes
that by just skipping words whose first |len(token)| characters don't
match the token.

### DIFF
--- a/scripts/idl-analyze.py
+++ b/scripts/idl-analyze.py
@@ -13,7 +13,7 @@ def find_line_column(text, token, pos):
         if text[pos] == '\n':
             return (0, 0)
 
-        pos += 1
+        pos = text.find(' ', pos) + 1
 
     line = 0
     while pos > linebreaks[line]:


kashav:
Other maybe interesting attributes that this affected:

- [`MIMEInfo` in nsIExternalHelperAppService.idl](https://searchfox.org/mozilla-central/rev/4648b6ee31c2519b1753023e4f4853b14fdd16e5/uriloader/exthandler/nsIExternalHelperAppService.idl#115)
- [`XULBrowserWindow` in nsIAppWindow.idl](https://searchfox.org/mozilla-central/rev/4648b6ee31c2519b1753023e4f4853b14fdd16e5/xpfe/appshell/nsIAppWindow.idl#137)
- [`result` in nsIWebBrowserPersist.idl](https://searchfox.org/mozilla-central/rev/4648b6ee31c2519b1753023e4f4853b14fdd16e5/dom/webbrowserpersist/nsIWebBrowserPersist.idl#98)

kashav:
And changes the analysis output for the following files:

imgIRequest.idl
nsIAlertsService.idl
nsIAppWindow.idl
nsIChannel.idl
nsIExternalHelperAppService.idl 
nsIHttpAuthenticableChannel.idl
nsIIncrementalDownload.idl
nsILoadInfo.idl
nsIPrincipal.idl
nsISHEntry.idl
nsIWebBrowserPersist.idl
nsIWebProgress.idl
nsIWebSocketChannel.idl

all of which appear to be expected (most just have a `nsIURI` attribute named `URI`).

asutherland:
Thanks for filing and fixing this and digging into the other situations where the current naive heuristics go wrong!

asutherland:
Given the testing flags we use in phabricator now, I do feel compelled to add a testing description:

Testing: exception other
- Our check scripts aren't yet capable of expressing changes in offsets.  They just do boolean checks.  I have a (very) WIP for https://bugzilla.mozilla.org/show_bug.cgi?id=1707282 (noting that in general people should never be worried about bit-rotting me, as my [eyes are frequently bigger than my stomach](https://en.wiktionary.org/wiki/have_eyes_bigger_than_one%27s_stomach)).
- We don't actually have any .idl test cases in [tests/tests/files](https://github.com/mozsearch/mozsearch/tree/master/tests/tests/files) yet either.  (And the whole mechanism by which we fetch the xpidl parser from the tree is generally weird to me.  And since we won't generate bindings, it is hard to test in the mozsearch repo, etc.)
-

asutherland:
I've triggered a config1 indexer job so we should see the results in ~3 hours.

PR #422, opened 2021-05-17
Bug 1707282 - Implement searchfox-tool command line tool.

asutherland:
This implements "searchfox-tool" a new binary, as proposed in https://bugzilla.mozilla.org/show_bug.cgi?id=1707282 but does not yet do the "insta" snapshot testing or otherwise replace the check-helper.sh mechanism yet, but I think it's quite possibly ready to do so.

Given that this is already a sizable PR and the searchfox-tool thing is already a MVP and I won't be able to get to the testing stuff until the next weekend, I'm putting this up for review now.

## Example usages:

Here we see us filtering the contents of the (raw) analysis files to "source" (via record type) records that are a "def" (via kind) for a specific identifier "outerNS::Thing::Thing" via both local index from disk and localhost server:
```
> cargo run --bin searchfox-tool -- '--server http://localhost/ --tree tests filter-analysis big_cpp.cpp -k def -i outerNS::Thing::Thing -r source'
{"loc":"00147:2-7","nestingRange":"149:20-151:2","pretty":"function outerNS::Thing::Thing","source":1,"sym":"_ZN7outerNS5ThingC1Ei","syntax":"def,function"}

> cargo run --bin searchfox-tool -- '--server /home/vagrant/index/config.json --tree tests filter-analysis big_cpp.cpp -k def -i outerNS::Thing::Thing -r source'
{"loc":"00147:2-7","nestingRange":"149:20-151:2","pretty":"function outerNS::Thing::Thing","source":1,"sym":"_ZN7outerNS5ThingC1Ei","syntax":"def,function"}
```

Here we see the same commands, but now we pipe the output of the filtering to the "show-html" command which extracts the source lines for which we had records.
```
> cargo run --bin searchfox-tool -- '--server /home/vagrant/index/config.json --tree tests filter-analysis big_cpp.cpp -k def -i outerNS::Thing::Thing -r source | show-html'
<div role="row" id="line-147" class="source-line-with-number nesting-sticky-line">
  <div role="cell"><div class="cov-strip cov-no-data"></div></div>
  <div role="cell"><div class="blame-strip"></div></div>
  <div role="cell" class="line-number" data-line-number="147"></div>
  <code role="cell" class="source-line">  <span class="syn_def" data-symbols="_ZN7outerNS5ThingC1Ei" data-i="23" >Thing</span>(<span class="syn_reserved" >int</span> <span data-symbols="V_328b2e25607c3527_812a493f256" >baseHP</span>)
</code>
</div>

> cargo run --bin searchfox-tool -- '--server http://localhost/ --tree tests filter-analysis big_cpp.cpp -k def -i outerNS::Thing::Thing -r source | show-html'
<div role="row" id="line-147" class="source-line-with-number nesting-sticky-line">
  <div role="cell"><div class="cov-strip cov-no-data"></div></div>
  <div role="cell"><div class="blame-strip"></div></div>
  <div role="cell" class="line-number" data-line-number="147"></div>
  <code role="cell" class="source-line">  <span class="syn_def" data-symbols="_ZN7outerNS5ThingC1Ei" data-i="23" >Thing</span>(<span class="syn_reserved" >int</span> <span data-symbols="V_328b2e25607c3527_812a493f256" >baseHP</span>)
</code>
</div>
```

## General Operation

- `searchfox-tool.rs` is currently the only CLI interface, it takes the OS args and is assuming there is only a single argument now which contains a jq-style-ish query that's pipe-delimited.  (In other words, something that looks like a normal shell pipeline then gets quoted so that it has a single argument.)
- `cmd_pipeline/builder.rs`:
  - Takes the single argument and uses the `shell_words` crate to create a normal argument stream which then gets split on `|` and we feed the sub-slices to the structopt-built argument parser which understands a single command.  Each command is mapped into a pipeline command.  These exist as `cmd_pipeline/cmd_*.rs` of which there are 2 real ones right now and 2 stub ones.
  - It treats the first set of arguments as special, which is where we extract the global arguments from, specifically deciding what server to use.
  - There is a local-index and a remote-HTTP server option with all of the stuff living under `abstract_server`.  Both use async tokio.  We use tokio because it seems like we could potentially use some of this command pipeline in the future as part of web-server.rs, and in that case we want to be responsible async users.
- `searchfox-tool.rs` takes the pipeline and runs the commands in succession, feeding the output of the previous step to the input of the next step.  These steps in the above example are:
  - `cmd_pipeline/cmd_filter_analysis.rs`: Uses serde_json JSON-reps (our current use of `rustc_serialize::json` is not a modern best practice, so I figured I'd start the transition here to get some experience with it) and the structopt parameters to support various filtering.
  - `cmd_pipeline/cmd_show_html.rs`: Uses `lol_html` so that we can use a selector-based mechanism to pick out the lines of interest.  There isn't currently an excerpting mechanism so we use a hack so that we have a stateful OutputSink based on the lines we've seen.  Because of how the hack works and me forgetting about the edge-case until just now, we can't actually excerpt the last line of a file, but I'm willing to live with that for now...

### DIFF
--- a/tools/src/abstract_server/local_index.rs
+++ b/tools/src/abstract_server/local_index.rs
@@ -1,0 +1,126 @@
+use async_trait::async_trait;
+use flate2::read::GzDecoder;
+use serde_json::{from_str, Value};
+use std::io::Read;
+use tokio::fs::File;
+use tokio::io::AsyncReadExt;
+
+use super::server_interface::{AbstractServer, ErrorDetails, ErrorLayer, Result, ServerError};
+
+use crate::config::{load, TreeConfigPaths};
+
+/// IO errors amount to a 404 for our purposes which means a sticky problem.
+impl From<std::io::Error> for ServerError {
+    fn from(err: std::io::Error) -> ServerError {
+        ServerError::StickyProblem(ErrorDetails {
+            layer: ErrorLayer::ServerLayer,
+            message: err.to_string(),
+        })
+    }
+}
+
+/// Read newline-delimited JSON that's been gzip-compressed.
+async fn read_gzipped_ndjson_from_file(path: &str) -> Result<Vec<Value>> {
+    let mut f = File::open(path).await?;
+    // We read the entirety to a buffer because
+    // https://github.com/serde-rs/json/issues/160 suggests that the buffered
+    // reader performance is likely to be much worse.
+    //
+    // When we want to go async here,
+    // https://github.com/rust-lang/flate2-rs/pull/214 suggests that we want to
+    // use the `async-compression` crate.
+    let mut buffer = Vec::new();
+    f.read_to_end(&mut buffer).await?;
+
+    let mut gz = GzDecoder::new(&buffer[..]);
+
+    let mut raw_str = String::new();
+    gz.read_to_string(&mut raw_str)?;
+
+    // let mut raw_str = String::new();
+    // f.read_to_string(&mut raw_str).await?;
+
+    raw_str
+        .lines()
+        .map(|s| from_str(s).map_err(|e| ServerError::from(e)))
+        .collect()
+}
+
+#[derive(Debug)]
+struct LocalIndex {
+    // We only hold onto the TreeConfigPaths portion of the config because the
+    // git data is not `Sync`.  Specifically, the compiler says:
+    //
+    // "within `TreeConfig`, the trait `Sync` is not implemented for
+    // `*mut libgit2_sys::git_repository`"
+    //
+    // When we need to do local git stuff, we will be able to accomplish this by
+    // creating a new `git2::Repository` on demand from the git path.  This is
+    // already done in `build-blame.rs` for its compute threads and that's
+    // likely the model we should use.
+    config_paths: TreeConfigPaths,
+    tree_name: String,
+}
+
+#[async_trait]
+impl AbstractServer for LocalIndex {
+    async fn fetch_raw_analysis(&self, sf_path: &str) -> Result<Vec<Value>> {
+        let full_path = format!("{}/analysis/{}.gz", self.config_paths.index_path, sf_path);
+        read_gzipped_ndjson_from_file(&full_path).await
+    }
+
+    async fn fetch_html(&self, sf_path: &str) -> Result<String> {
+        let full_path = format!("{}/file/{}.gz", self.config_paths.index_path, sf_path);
+
+        // If we were dealing with uncompressed files.
+        /*
+        let mut f = File::open(full_path).await?;
+        let mut raw_str = String::new();
+        f.read_to_string(&mut raw_str).await?;
+        */
+
+        let mut f = File::open(full_path).await?;
+        let mut buffer = Vec::new();
+        f.read_to_end(&mut buffer).await?;
+
+        // When we want to go async here,
+        // https://github.com/rust-lang/flate2-rs/pull/214 suggests that we want
+        // to use the `async-compression` crate.
+        let mut gz = GzDecoder::new(&buffer[..]);
+
+        let mut raw_str = String::new();
+        gz.read_to_string(&mut raw_str)?;
+
+        Ok(raw_str)
+    }
+
+    async fn perform_query(&self, _q: &str) -> Result<Value> {
+        // TODO: For this to work, we want to be able to directly invoke the
+        // underpinnings of the web server, which entails porting router.py into
+        // web-server.rs, an act which may involve building it on some of this
+        // infrastructure...
+        Err(ServerError::Unsupported)
+    }
+}
+
+pub fn make_local_server(
+    config_path: &str,
+    tree_name: &str,
+) -> Result<Box<dyn AbstractServer + Send + Sync>> {
+    let mut config = load(config_path, false);
+    let tree_config = match config.trees.remove(&tree_name.to_string()) {
+        Some(t) => t,
+        None => {
+            return Err(ServerError::StickyProblem(ErrorDetails {
+                layer: ErrorLayer::BadInput,
+                message: format!("bad tree name: {}", &tree_name),
+            }))
+        }
+    };
+
+    Ok(Box::new(LocalIndex {
+        // We don't need the blame_map and hg_map (yet)
+        config_paths: tree_config.paths,
+        tree_name: tree_name.to_string(),
+    }))
+}
--- a/tools/src/abstract_server/remote_server.rs
+++ b/tools/src/abstract_server/remote_server.rs
@@ -1,0 +1,105 @@
+use async_trait::async_trait;
+use serde_json::{from_str, Value};
+use url::{ParseError, Url};
+
+use super::server_interface::{AbstractServer, ErrorDetails, ErrorLayer, Result, ServerError};
+
+/// reqwest won't return an error for an unhappy status code itself; someone
+/// would need to call `Response::error_from_status`, so for now we'll generally
+/// assume everything is some kind of transient problem.
+impl From<reqwest::Error> for ServerError {
+    fn from(err: reqwest::Error) -> ServerError {
+        ServerError::TransientProblem(ErrorDetails {
+            layer: ErrorLayer::ServerLayer,
+            message: err.to_string(),
+        })
+    }
+}
+
+impl From<ParseError> for ServerError {
+    fn from(err: ParseError) -> ServerError {
+        ServerError::StickyProblem(ErrorDetails {
+            layer: ErrorLayer::BadInput,
+            message: err.to_string(),
+        })
+    }
+}
+
+#[derive(Debug)]
+struct RemoteServer {
+    server_base_url: Url,
+    tree_base_url: Url,
+    source_base_url: Url,
+    raw_analysis_base_url: Url,
+    search_url: Url,
+}
+
+async fn get(url: Url) -> Result<reqwest::Response> {
+    //println!("Using URL {}", url);
+    let res = reqwest::get(url).await?;
+
+    if !res.status().is_success() {
+        if res.status().is_server_error() {
+            return Err(ServerError::TransientProblem(ErrorDetails {
+                layer: ErrorLayer::ServerLayer,
+                message: format!("Server status of {}", res.status()),
+            }));
+        } else {
+            return Err(ServerError::StickyProblem(ErrorDetails {
+                layer: ErrorLayer::DataLayer,
+                message: format!("Server status of {}", res.status()),
+            }));
+        }
+    }
+
+    Ok(res)
+}
+
+#[async_trait]
+impl AbstractServer for RemoteServer {
+    async fn fetch_raw_analysis(&self, sf_path: &str) -> Result<Vec<Value>> {
+        let url = self.raw_analysis_base_url.join(sf_path)?;
+        let raw_str = get(url).await?.text().await?;
+        raw_str
+            .lines()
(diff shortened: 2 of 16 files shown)


asutherland:
Er, and I suppose a disclaimer is that because of the late stage in which the pipeline cleverness was added, more iteration is required to get "--help" to work cleanly because I used the `_safe` variant of argument parsing which returns an error when `--help` is invoked because that forbids it from synchronously exiting to the command line, which is not something we'd want in web-server.rs.

Examples where I copied and pasted the error message that was output and passed them to `echo -e`:

General help:
```
USAGE:
    searchfox-tool [OPTIONS] <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -o, --output-format <output-format>     [default: concise]  [possible values: Pretty, Concise]
        --server <server>                  URL of the server to query or the path to the root of the index tree if using
                                           local data [env: SEARCHFOX_SERVER=]  [default: https://searchfox.org/]
        --tree <tree>                      The name of the indexed tree to use [env: SEARCHFOX_TREE=]  [default:
                                           mozilla-central]

SUBCOMMANDS:
    filter-analysis      
    help                 Prints this message or the help of the given subcommand(s)
    identifier-lookup    
    query                
    show-html            
```

filter-analysis help:
```
searchfox-tool-filter-analysis 0.1.0

USAGE:
    searchfox-tool filter-analysis [OPTIONS] <file>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -i <identifier>                       Exact identifier match
    -k, --kind <kind>                     
    -r, --record-type <record-type>...     [possible values: Source, Target]
    -s <symbol>                           Exact symbol match

ARGS:
    <file>    Tree-relative analysis file path
```

tools/src/abstract_server/server_interface.rs
```diff
@@ -0,0 +1,96 @@
...
+    DataLayer,
+    /// We're not sure if it was a server issue or a data issue.
+    UnknownLayer,
+}
+
+/// ServerError payload to provide details about what went wrong for
+/// investigation purposes.  In the future, this could wrap the
+/// underlying errors we've seen.
+#[derive(Debug)]
+pub struct ErrorDetails {
```
staktrace, line 25: I really like the way this error stuff is structured. And really nice documentation, thanks!

tools/src/abstract_server/remote_server.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    Ok(res)
+}
+
+#[async_trait]
+impl AbstractServer for RemoteServer {
+    async fn fetch_raw_analysis(&self, sf_path: &str) -> Result<Vec<Value>> {
+        let url = self.raw_analysis_base_url.join(sf_path)?;
+        let raw_str = get(url).await?.text().await?;
+        Ok(raw_str.lines().filter_map(|s| from_str(s).ok()).collect())
```
staktrace, line 63: Ditto here on the ok() dropping errors

tools/src/cmd_pipeline/parser.rs
```diff
@@ -0,0 +1,51 @@
...
+use super::cmd_filter_analysis::FilterAnalysis;
+use super::cmd_show_html::ShowHtml;
+
+arg_enum! {
+    #[derive(Debug, PartialEq)]
+    pub enum OutputFormat {
+        // Pretty-printed JSON.
+        Pretty,
+        // Un-pretty-printed JSON.
+        Concise,
```
staktrace, line 13: Didn't want to go with `Ugly`? :p

tools/src/cmd_pipeline/cmd_show_html.rs
```diff
@@ -0,0 +1,149 @@
...
+            // the spin-off https://github.com/cloudflare/lol-html/issues/78
+            // are implemented, lol_html doesn't explicitly provide a way to
+            // derive the value of an element.
+            //
+            // So we attempt a hack where we use a custom output sink that is
+            // kept aware of where we are in the file.  The good news is that
+            // since lol_html is oriented around minimal memory allocation, we
+            // can generally control when flushes happen.
+
+            let mut writing_line: u32 = 0;
```
staktrace, line 82: I didn't look at this code in detail (need to sleep soon) but it might be nice to extract helper functions with a couple of standalone tests. It looks like the kind of code where things can go wrong easily.

tools/src/cmd_pipeline/parser.rs
```diff
@@ -0,0 +1,51 @@
...
+#[derive(Debug, StructOpt)]
+pub enum Command {
+    Query(Query),
+    FilterAnalysis(FilterAnalysis),
+    ShowHtml(ShowHtml),
+    IdentifierLookup(IdentifierLookup),
+}
+
+#[derive(Debug, StructOpt)]
+pub struct Query {}
```
staktrace, line 48: Is this supposed to be linked to the `SymbolicQueryOpts` structure somehow? It looks to me like this `Query` struct being empty means it's a "TODO: implement this later" kind of thing but then the existence of `SymbolicQueryOpts` seems to imply that it was partially implemented. I don't see anything in the pipeline builder code though, so maybe `SymbolicQueryOpts` is a dangling struct.

staktrace (approved):
This is really cool! TBH I had a hard time following the bug description but it all makes a lot more sense now that I have some code to look at :)

I did a quick pass over the code and it looks pretty reasonable, so I'll r+ now. I'll probably spend some more time over the next days/weeks looking at it in more detail, although I'm not sure how much time I'll have to dedicate to it.

tools/src/abstract_server/local_index.rs
```diff
@@ -0,0 +1,123 @@
...
+        let mut f = File::open(full_path).await?;
+        let mut raw_str = String::new();
+        f.read_to_string(&mut raw_str).await?;
+        */
+
+        let mut f = File::open(full_path).await?;
+        let mut buffer = Vec::new();
+        f.read_to_end(&mut buffer).await?;
+
+        let mut gz = GzDecoder::new(&buffer[..]);
```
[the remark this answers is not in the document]
staktrace: https://github.com/rust-lang/flate2-rs/pull/214 seems to suggest that using the [async_compression](https://docs.rs/async-compression/0.3.8/async_compression/) crate is the way to go here.

asutherland (commented):
Thanks very much for the feedback!

It doesn't sound like there's any opposition to me landing this after changing the noted error handling, but I should note that if it makes sense, I could:
- Create a new branch "structured-landing"
- Land this PR to that branch.
- Plan to have my next PR be to add the "insta" testing.
- Rebase the structured analysis PR onto that branch.  I would then add 2 extra commits:
  1. Changes to the "insta" test expectations for existing behavior.
  2. Addition of new test coverage for structured records.

Then that whole branch could be merged across/landed.

Or I can just land this, then do a new "insta"/testing PR that can land, then just do the extra testing commits for the structured record support.  I don't have any personal preference, just want to provide a possibility that perhaps better reflects the reality that the atomic unit here is really the desire to land structured record support with tests, but not have an overwhelming mega-PR that involves things you already reviewed.

tools/src/abstract_server/local_index.rs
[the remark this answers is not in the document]
staktrace: Yeah a comment here is fine.

tools/src/cmd_pipeline/cmd_show_html.rs
staktrace: Ok, sounds good.

tools/src/cmd_pipeline/parser.rs
staktrace: Ah, I missed that FilterAnalysis used it. Very well then!

staktrace (approved):
Feel free to merge this directly (and not do it on a branch).

tools/src/abstract_server/local_index.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    let mut gz = GzDecoder::new(&buffer[..]);
+
+    let mut raw_str = String::new();
+    gz.read_to_string(&mut raw_str)?;
+
+    // let mut raw_str = String::new();
+    // f.read_to_string(&mut raw_str).await?;
+
+    Ok(raw_str.lines().filter_map(|s| from_str(s).ok()).collect())
```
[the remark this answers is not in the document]
asutherland: Okay, I addressed this here and in remote_server by changing:
```rust
Ok(raw_str.lines().filter_map(|s| from_str(s).ok()).collect())
```
into (after rustfmt)
```rust
    raw_str
        .lines()
        .map(|s| from_str(s).map_err(|e| ServerError::from(e)))
        .collect()
```
after also implementing `impl From<serde_json::Error> for ServerError` in `server_interface.rs` since serde is a common error between all the implementations.

At first I just tried without the `map_err`, but this resulted in an error because `FromIterator` didn't know how to transform the error when going from `Vec<Result<T, E>>` to `Result<Vec<T>, E>` because the E's weren't the same.  I'm not sure if I actually could have implemented a custom FromIterator, but it seemed like a lot of typing, and in general I understand Vec::collect() to generally require help with types, so this seemed reasonably straightforward.

I'm definitely open to hints on more idiomatic options here!

tools/src/abstract_server/local_index.rs
```diff
@@ -0,0 +1,123 @@
...
+
+#[derive(Debug)]
+struct LocalIndex {
+    // We only hold onto the TreeConfigPaths portion of the config because the
+    // git data is not `Sync`.  Specifically, the compiler says:
+    //
+    // "within `TreeConfig`, the trait `Sync` is not implemented for
+    // `*mut libgit2_sys::git_repository`"
+    //
+    // Obviously, it would be nice to be able to do git stuff, but that needs to
```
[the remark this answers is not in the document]
asutherland: I've transformed my hand-waving comments here to basically what you've said here, cross-referencing the `buld-blame.rs` compute threads.  One implication here that I didn't leave in the comment is that presumably this means that all the git stuff will need to be done in straight runs that never awaits (because that allows thread changes and the type signatures enforce this).

tools/src/abstract_server/local_index.rs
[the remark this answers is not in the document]
asutherland: I've added a comment here and in the `read_gzipped_ndjson_from_file` case above.

tools/src/abstract_server/remote_server.rs
asutherland: (addressed as discussed above)

tools/src/abstract_server/server_interface.rs
asutherland: Thanks!

tools/src/cmd_pipeline/parser.rs
asutherland: Ha!

asutherland (commented):
Okay, I think I addressed all the review comments.  I'm gonna squash my changes and land this.  My apologies that I think this will bit-rot your `Cargo.lock` changes for sure and maybe your `Cargo.toml` changes too.

Should there be a robot that is the one that updates changes to Cargo.lock?  Should we not actually be committing it?

staktrace:
Ehh rebuilding Cargo.lock isn't a huge burden, since `cargo update` does the work. But I do think there is value in keeping it checked in, so that the code that runs on the indexer doesn't automatically pick up random updates to crates every time it runs. That would probably result in far more indexer failures.

PR #423, opened 2021-05-22
Bug 1683563 - Tweak diff similarity options

staktrace:
There isn't a new git2-rs release yet with the changes, so for now I'm just using the master branch head revision. Once there's a new release I'll update the Cargo.* files to use that instead.

### DIFF
--- a/tools/Cargo.lock
+++ b/tools/Cargo.lock
@@ -612,9 +612,9 @@ dependencies = [
 
 [[package]]
 name = "git2"
-version = "0.13.17"
+version = "0.13.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


staktrace:
Hm, CI for this failed: https://travis-ci.org/github/staktrace/mozsearch/builds/771969860

I can reproduce that locally, looks like the very latest rustc nightly won't compile git2 properly. Based on what I'm seeing I expect tomorrow's indexing run to fail as well when it tries to build the mozsearch tools. :(

asutherland (approved):
r=asuth for the net result of getting the thing that will build.

As always, thank you for being a model of good open source behaviors (filing high quality upstream bugs, helping provide context for the bugs, helping provide tests for the bugs, fixes, etc.).

PR #424, opened 2021-05-22
Downgrade git2 to 0.13.17

staktrace:
Versions newer that than are affected by https://github.com//rust-lang/rust/issues/85574
which is a regression in rustc nightly. That's being reverted but for now
let's roll back a bit. We were using 0.13.15 until very recently so 0.13.17
should be safe enough.

### DIFF
--- a/tools/Cargo.lock
+++ b/tools/Cargo.lock
@@ -2,9 +2,9 @@
 # It is not intended for manual editing.
 [[package]]
 name = "adler"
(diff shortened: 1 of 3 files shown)


staktrace:
Going to go ahead and merge since indexer runs are going to fail unless this is addressed somehow. It looks like the rustc patch is getting backed out but it might take a day or two for it to propagate.

PR #425, opened 2021-05-23
Bug 1707282 - insta hookup for searchfox-tool based "check" mechanism

asutherland:
- Introduces a test that's an insta test-driver thing that expects to be
  told a CHECK_ROOT via the environment and which we expect to be found at
  CONFIG/TREE/checks.  Beneath that, we expect "inputs" to contain
  searchfox-tool command lines and "snapshots" to be the insta-created
  results of the pipeline.
- Hooks that up by changing `check-index.sh` to use it instead of the check
  scripts.  As the comments in the test driver file above try and make clear,
  this is paired with the above and I may just be abusing the `cargo test`
  mechanism badly, so maybe some stuff should change here.  But maybe it
  changes in a follow-up if we can't figure out a concrete plan beyond
  acknowledging the current approach is very sketchy.
- Tries to convert the existing check script into command-lines.
  - For some of the rust stuff I ended up changing the symbols we were looking
    for because we were producing 3 definitions for the given symbol.  Instead
    I tried to pick symbols with a single definition.
    - I also noticed we don't seem to mark rust source records with a syntax
      that includes a def.  Definitely a neat demonstration of how this new
      style of testing can maybe teach us things.
  - I made up a weird convention that maybe is dumb.
    - The "__json" suffixed pipelines are just the filtered analysis records,
      sometimes filtered to definitions when we otherwise would also see uses.
    - The "__html" suffixed pipelines are exactly the same as the above but
      then with `| show-html` tacked on to get us excerpts.
      - This could definitely be done automatically as a permutation, but the
        rationale was that we may eventually have checks where that's not an
        appropriate transform or something more clever needs to be done.
- Actually runs the checks against both the local index and the server,
  probably.
  - I see two sets of checks being run and I see we're setting the environment
    variables the right way, but I didn't wireshark it to be sure.  However,
    since I set all the defaults to hit https://searchfox.org/ by default,
    there's a good chance things are really working!

What it does not do:
- Does not try to add new checks yet.  But it could make sense to add more,
  especially since the HTML output should include coverage info!
- Does not try and do the thing I proposed where we might have a separate set
  of directories where we basically check in the entirety of the produced
  analysis files, etc.  I was worried that might be spammy and I'm still
  worried that might be spammy, so I'd like to punt on this until such time as
  we overwhelmingly regret not having done that.

What will need to be done before this lands:
- I haven't touched the mozsearch-mozilla checks.  I want to see where we end
  up with this patch first, but they shouldn't be hard to do.

Dependency related changes:
- The RLS crates get updated because we needed the fix for
  https://github.com/rust-lang/rls/issues/1711 where some kind of serde
  broke.
  - This upgrade did result in a minor change to rust-indexer.rs.


asutherland:
This ended up being less of a hassle than I was expecting!

I realize now I need to update/add some docs, but the main things here are:
- `cargo install cargo-insta` in your VM to be able to:
- `make review-test-repo` in `/vagrant` in your VM to be able to experience the cool UI shown at https://insta.rs/docs/cli/ to see differences and approve them.

In terms of review, I'm very much on board with changing pretty much everything about this up in order to reach a consensus on the ergonomics around this.

tools/src/bin/searchfox-tool.rs
```diff
@@ -55,6 +42,10 @@ async fn main() {
                     }
                 }
             }
+        },
+        Err(err) => {
+            println!("Pipeline Error!");
+            println!("{:?}", err);
```
staktrace, line 48: Should this be exiting with an error code?

staktrace (approved):
This looks straightforward enough. Does it make sense to add a `cargo install cargo-insta` to the provisioner scripts?

tools/src/bin/searchfox-tool.rs
asutherland: Yes, that definitely seems appropriate; I'll have it exit with an error code, thanks!

asutherland (commented):
Yes, I should absolutely add it to the provisioner scripts; just wanted to make sure this structuring made sufficient sense.

asutherland:
I'm back to this now to try and quickly land this and hopefully the structured analysis stuff that exists thus far.

> Yes, I should absolutely add it to the provisioner scripts; just wanted to make sure this structuring made sufficient sense.

I've added this to `indexer-provision.sh` just after we install rust like so:
```shell
# cargo-insta makes it possible to use the UI documented at
# https://insta.rs/docs/cli/ to review changes to "check" scripts.  For the test
# repo, this is used by `make review-test-repo`.  It's not expected that this
# will actually be necessary on the production indexer and so this isn't part of
# the update process.
cargo install cargo-insta
```

### Non-test-repo Checks

I propose the following:
- Release indexer runs: "check" script errors will be fatal because broken results are bad.  The current behavior will continue: Web server will not be spawned, indexer will stop the VM to allow manual investigation at a later time.
- Try/non-release indexer runs: "check" script errors will be warnings for non-release indexers.  The web server will be spawned and display the results.
  - This will allow the mozsearch-mozilla "check" scripts to be updated from the developer's local machine using a mozsearch-mozilla checkout which can then be used to generate a pull request.  The goal is the https operation mode can remotely retrieve everything it needs from the server without having to involve ssh or anything.  (This will still enable the local checks to work on the indexer though, at least as long as the `RemoteServer` has general feature parity with the `LocalIndex`.)

Assuming that plan works, I'll write-up some docs for this as a new `docs/testing.md` file.

asutherland:
I made the mozsearch-mozilla changes in https://github.com/mozsearch/mozsearch-mozilla/pull/146 which I will be landing concurrently.

PR #427, opened 2021-05-24
Bug 1712578 - Put TMPDIR on /mnt/index-scratch. a=bustage-fix

asutherland:


asutherland:
Landing immediately for bustage-fix reasons and because it'll be pretty obvious if this breaks the world or does not break the world.

staktrace:
Looks good, thanks!

PR #428, opened 2021-05-25
Bug 1712578 - Use TMPDIR if defined over /tmp in shell scripts. a=bustage

asutherland:
We have some logic that directly references /tmp that assumes it will have
bountiful storage available, but the reality is that it does not.  In
particular, `crossref.sh` was directly sorting the ~1.6G into the /tmp dir
which is very close to the amount of free space on our `/` EBS image.

This is not meant to be the end state fix for this, so I've just corrected
the shell scripts where it's trivial to evaluate an environment variable
and failover.  I have not touched the following other files under /scripts
that have the same problem but involve python/JS:
- output-dir.js
- idl-analyze.py
- idl-analyze.sh (because of the above)
- build-codesearch.py

The simplest solution in terms of avoiding using the wrong tmp dir would
be to just make sure that the /tmp dir is actually backed by the local
SSD, but that would ideally want to happen early in the machine boot
process to avoid /tmp already being in use when we actually get around to
creating the scratch drive.

Maybe that's okay though?  Otherwise we'll potentially want to enforce
some kind of invariant about TMPDIR being set and/or the indexing process
having a "tmp" directory in each index directory (which would make sense
for the purposes of having `mv` operating within the same filesystem).

### DIFF
--- a/scripts/crossref.sh
+++ b/scripts/crossref.sh
@@ -11,14 +11,14 @@ echo Root is $INDEX_ROOT
 
 # Find the files to cross-reference.
 cd $INDEX_ROOT/analysis
-find . -type f | cut -c 3- > /tmp/files
+find . -type f | cut -c 3- > ${TMPDIR:-/tmp}/files
 cd -
 
-$MOZSEARCH_PATH/tools/target/release/crossref $CONFIG_FILE $TREE_NAME /tmp/files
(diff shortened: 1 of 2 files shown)


asutherland:
@staktrace I'm landing this now and going to trigger a config2 indexer, but we definitely will want a more considered/consensus-ified solution than this.

staktrace:
I like the idea of using a tmp dir inside the index dir. If we ever parallelize some of these steps across repos, then using the same `TMPDIR`/`/tmp` folder would not be safe, whereas per-index dirs would be safe.

PR #429, opened 2021-06-17
Bug 1716914 - Stop populating the author field in the blame data

staktrace:


staktrace:
https://kats.searchfox.org/ has this deployed, I verified it all works as expected.

asutherland:
Nice!

Is the expectation that we'll just do this as a slipstream change for now and then next time we regenerate blame it will use up less disk space in the future?

staktrace:
> Is the expectation that we'll just do this as a slipstream change for now and then next time we regenerate blame it will use up less disk space in the future?

Yeah, basically. I have some other changes I'm mulling over that might require a full regeneration so I figured I'd get this in beforehand.

asutherland:
Sweet!  The only blame related thing on my radar has been that (and this has been previously proposed) maybe we should just bake in the blame strip results that are currently asynchronously fetched.  An upside of this is that we should then also be able to do fancy things like "Choose sidebar blame colour based on how recent the change was " https://bugzilla.mozilla.org/show_bug.cgi?id=1545151 while also side-stepping the server performance issues (which I'll still try to address, especially to ensure we get proper parallelism).

tools/src/bin/build-blame.rs
```diff
@@ -278,15 +277,15 @@ fn blame_for_path(
 ) -> Result<String, git2::Error> {
     let linecount = count_lines(&blob);
     // TODO: drop this author field entirely, I don't think it gets consumed by anything
```
jens1o, line 279: This TODO line can be removed, can't it?

staktrace: Oh indeed, yes it can :)

staktrace: https://github.com/mozsearch/mozsearch/pull/431

jens1o: thanks!

PR #430, opened 2021-06-19
Bug 1687016 - Bump cinnabar in provisioning scripts to current tip.

asutherland:
We actually want this for 2 situations:
- Bug 1687016 - hg2git/git2hg hang fixes
- Bug 1716167 - to see if the cinnabar fsck becomes happier


asutherland:
This is the current version of cinnabar I've been using locally this week and it seems fine.  Going to merge this now and re-provision and maybe see if I can quickly re-trigger config1 with the new setup while it's still Saturday.

asutherland:
Re-provisioning completed but timestamps from today's first fsck run indicated the "Saturday" check would fail so I manually edited the already-checked-out git-maintenance script to be unconditional (for this run).

PR #431, opened 2021-06-21
Bug 1716914 - Remove dangling TODO

staktrace:


staktrace:
(Merging since this is trivial.)

PR #432, opened 2021-07-17
Bug 1720037 - Run the update step when provisioning to cache crates.

asutherland:
The following changes are made:
- Provisioning is given an IAM role so that the `aws` command can use it.
  No role = no way to run `aws` without manually providing credentials.
  - It's the same IAM role we use for indexers!
- Indexer and web server AWS provisioning resizes their root partition up
  to 12 gigs from 8.  This depends on the IAM role to work.
- The indexer and web server scripts run their update scripts for side
  effect of caching crates but also validating the installs.
  - This happens in the core non-AWS scripts primarily so that we notice
    any mistakes in changes to the provisioning process sooner when
    rebuilding the VM locally.
- The update scripts and their helper scripts have been updated to handle
  being run multiple times, once at provisioning, and once at spawning.
  - `vagrant up` was very helpful in testing this since it runs both
    indexer and web-server provisioning, which is not quite the same
    thing that happens on the servers (because they're different scripts!)
    but helped uncover the problem.
  - I've added a second invocation of the update scripts as part of the
    provisioning to provide that fidelity during provisioning, noting
    that there is some awkwardness about how the update process is
    explicitly checking out the tip of the repo which makes it hard to
    fully test things until they're landed on the tip.
    - This could be addressed by having provisioning use its own branch
      called "provisioning" or something like that, with changes to
      provisioning first staged there and then only moved to the main
      branch afterwards.  But that assumes some other automation would
      be in place.  That's too much for a manual process right now.

### DIFF
--- a/infrastructure/aws/indexer-provision.sh
+++ b/infrastructure/aws/indexer-provision.sh
@@ -7,6 +7,36 @@ set -o pipefail # Check all commands in a pipeline
 # We want the NVME tools, that's how EBS gets mounted now on "nitro" instances.
 sudo apt-get install -y nvme-cli
 
+date
+
+# Size up our root partition to 12G
+#
+# To this end we need to know the volume id in order to issue an EBS resizing
+# command.  Note that the select constraint here is intended more as a check
+# that our assumption about partition sizes hasn't changed, as when provisioning
+# there should only be this single EBS mount.
+ROOT_VOL_ID=$(sudo nvme list -o json | jq --raw-output '.Devices[] | select(.PhysicalSize < 9000000000) | .SerialNumber | sub("^vol"; "vol-")')
+AWS_REGION=us-west-2
+# The size is in gigs.
+aws ec2 modify-volume --region ${AWS_REGION} --volume-id ${ROOT_VOL_ID} --size 12
+# Re: hardcoded devices: The devices should currently be stable.
+#
+# We use an until loop because it can take some time for the change to
+# propagate to this VM.  The error will look like:
+#   "NOCHANGE: partition 1 is size 16775135. it cannot be grown"
+# And success will look like:
+#   "CHANGED: partition=1 start=2048 old: size=16775135 end=16777183 new: size=25163743 end=25165791"
+#
+# The 5 is arbitrary in both cases.
+sleep 5
+until sudo growpart /dev/nvme0n1 1
+do
+  sleep 5
+done
+sudo resize2fs /dev/nvme0n1p1
+
+date
+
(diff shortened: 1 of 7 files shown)


asutherland:
Landing this now, but your post-landing insight/wisdom is appreciated here.  There are a number of things that could likely be improved/addressed but I mainly just want this to be stable/working first.  In particular, I'm a little concerned about:
- The partition resizing happens after we finish doing things given our current documented invocations of how to provision things.  We might want to make sure we grow the partition as basically our first step.
- The hard-coded use of the default/main branch for the update scripts results in a complicated dance I think, but this is how I've already been provisioning things in general (always provisioning after merging).  I don't know if it would be better for the provisioning script to be doing some kind of interpolation or something like that.

staktrace:
In general this patch seems fine to me, I can't think of anything to improve offhand.

>     * The partition resizing happens after we finish doing things given our current documented invocations of how to provision things.  We might want to make sure we grow the partition as basically our first step.

So this would be done by switching the order of provisioner scripts, I think? Instead of the currently-documented command `infrastructure/aws/trigger-provision.py infrastructure/indexer-provision.sh infrastructure/aws/indexer-provision.sh` it would instead be `infrastructure/aws/trigger-provision.py infrastructure/aws/indexer-provision.sh infrastructure/indexer-provision.sh` (and similarly for the web-server). From a quick look at the scripts I imagine that should work fine out of the box, and seems like a reasonable thing to do in a followup. While doing that we might want to use python3 instead of python2 when running that cloudwatch script (TBH I'm not really sure what that even does). I can file a bug for this.

>     * The hard-coded use of the default/main branch for the update scripts results in a complicated dance I think, but this is how I've already been provisioning things in general (always provisioning after merging).  I don't know if it would be better for the provisioning script to be doing some kind of interpolation or something like that.

I think for now this is fine. We can see how it works and iterate if we find that we're forgetting to reprovision or provisioning pre-merge. Something else I thought of just now: another approach to solving the original problem could be to write a github action or equivalent to create a vendored tarball of mozsearch + dependenent crates on commits. I don't know if we can store such a tarball as a github release or maybe upload it to S3, but that would allow the indexer to download the dependency crates from a more reliable source than crates.io and hopefully with less network failures. I'll think about that and maybe file a bug.

staktrace:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1721037 and https://bugzilla.mozilla.org/show_bug.cgi?id=1721038 and

asutherland:
Thanks!

PR #434, opened 2021-08-29
Bug 1716167 - Add a way to launch an indexer shell

asutherland:
This patch adds a new `trigger_shell.py` command and extracts the duplicated
logic that existed between `trigger_indexer.py` and `trigger_blame_rebuild.py`
up into `trigger_common.py` which now uses argparse as an argument parser.

The primary motivation of the cleanup because optional arguments like `setenv`
seem useful across all indexer types.  This would also be the case for things
like allowing a command-line to override.

As hinted above, we also add arguments for:
- `--setenv` intended to be used like
  `--setenv TRYPUSH_REV=<40-char-rev-hash>`.
  - This specific try use-case is potentially a candidate for a custom arugment
    type perhaps with extra smarts.

### DIFF
--- a/infrastructure/aws/shell-setup.sh
+++ b/infrastructure/aws/shell-setup.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -x # Show commands
+set -eu # Errors/undefined vars are fatal
+set -o pipefail # Check all commands in a pipeline
+
+if [ $# != 6 ]
+then
+    echo "usage: $0 <branch> <channel> <mozsearch-repo-url> <config-repo-url> <config-repo-path> <config-file-name>"
+    exit 1
+fi
+
+SCRIPT_PATH=$(readlink -f "$0")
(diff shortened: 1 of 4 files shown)


asutherland:
I was able to successfully use this to launch a shell that ran the setup tasks so that I could add the esr10 remote, run the cinnabarbar fsck, then tar and upload the revised gecko.tar (which I backed up on s3 first).  You may have seen a lot of exciting force-pushes to the fancy-backend-only branch in this process, but I've now cherry-picked the patch out.

Per previous discussions, I'm going to land this immediately in a post-landing review cycle since I've tested this for both trigger_indexer and trigger_shell (which does leave the re-blaming to be regressed of course) and this should work well enough and I can fix up any review comments as they come and/or I'm happy to quickly review any iterative enhancements/fixes you might have in a different PR, etc.

staktrace (commented):
Nice, thanks!

PR #435, opened 2021-08-30
Bug 1641372 - Python s/consumes/callees/ fixes.

asutherland:
In the review fixes I renamed consumes to callees as suggested but my
renaming was not extensive enough!


asutherland:
This is a follow-up fix to https://github.com/mozsearch/mozsearch/pull/408

PR #437, opened 2021-08-31
Bug 1728294 - Use the nightly+release rust build configuration for te…

staktrace:
…sting

This avoids rebuilding all the rust code in a new configuration,
since it's already built in the nightly+release configuration for
other stuff.

### DIFF
--- a/scripts/check-index.sh
+++ b/scripts/check-index.sh
@@ -25,14 +25,14 @@ then
       SEARCHFOX_SERVER=${CONFIG_FILE} \
       SEARCHFOX_TREE=${TREE_NAME} \
       CHECK_ROOT=${CONFIG_REPO}/${TREE_NAME}/checks \
-      cargo test --manifest-path ${MOZSEARCH_PATH}/tools/Cargo.toml test_check_glob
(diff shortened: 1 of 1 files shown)


staktrace:
I'm running this change on the `kats` channel to verify the tests still pass. Putting up the PR optimistically anyway.

asutherland (approved):
Thanks!  I had assumed this was just the lack of `--release` and was hand-waving that it potentially would be good to have better stacks/etc., but given that I haven't seen any actual crashes through this process and we constrain the test to just be `test_check_glob` and for most errors to be in the non-rust indexing logic, this makes a lot of sense and will make the process much faster.

asutherland:
Relatedly, should we be taking additional actions in the provisioning pass to ensure that nightly is the case by default?

staktrace:
I guess the nightly bit is not actually required since we set nightly to be default here: https://github.com/mozsearch/mozsearch/blob/master/infrastructure/indexer-provision.sh#L97 and actually uninstall stable so nightly should be the only one present. So the recompile I was seeing must have been purely a debug/release difference. Ah well. I guess we can drop all the nightly flags scattered around

PR #438, opened 2021-09-01
Bug 1728376 - Show related symbols in search results with search upsell

asutherland:
Here's a candidate fix for the regression in being able to see overridden
methods.

### DIFF
--- a/clang-plugin/MozsearchIndexer.cpp
+++ b/clang-plugin/MozsearchIndexer.cpp
@@ -1612,7 +1612,8 @@ public:
         wasTemplate = true;
         D = D2->getTemplateInstantiationPattern();
       }
-      Kind = D2->isThisDeclarationADefinition() ? "def" : "decl";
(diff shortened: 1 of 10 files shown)


asutherland:
This is accidentally up on https://searchfox.org/ already because I didn't pay enough attention to the channel tags in ssh.py :(

Definitely motivation for me to create a python script that will correctly pick a spare config1 server that is not the actual production server to rotate to be dev.searchfox.org.  (Unfortunately my asuth.searchfox.org server was too old.)

asutherland:
I'm landing this now for post-review so this can sync up with the landing of the indexer changes in m-c at https://phabricator.services.mozilla.com/D124152 and generally on the basis that:
- We have an increasing test corpus for the back-end!
- The python changes worked locally and for Jeff on the production server modulo the backend bug that this addresses.

I am expecting that we will want another round of enhancements on this, particularly in terms of styling, etc.

PR #445, opened 2021-09-02
Bug 1728858 - Add support for StaticClassBlock

staktrace:


asutherland (approved):
When it rains, it pours!  Thanks so much for addressing all of these issues in rapid succession.  It's especially appreciated as I've tussled with literal rain and rain-related issues as of late!

asutherland:
I merged and re-triggered via lambda (doing the test tab and just clicking test because we don't care about the payload)

PR #447, opened 2021-09-06
Bug 1729178 - Add "query" test coverage and new test cases

asutherland:
(See individual commits for more details.)


asutherland:
One notable non-obvious change in the commits is the addition of an optional `WAIT` argument to the end of `web-server-run.sh`.  I added this because the "query" checks were failing on the web server check because the router.py server hadn't finished starting up yet and would report a 502 error instead.  I verified that things work without WAIT and with it.  `make build-test-repo` now uses WAIT.

PR #448, opened 2021-09-07
Bug 1728000 - Update try build and testing docs.

asutherland:
While writing the testing docs I realized there was a gap for the release
versus non-release variations, so I put an XXX in the document and want
to both address that and then update the docs once it's addressed.  I
think I may have some stashed changes to dig out that may contain the
fix needed.

I'm landing this now because I'm unlikely to have time to directly work
on this during the work week and want this information to be available
in case someone else can dig into any enhancements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ like `eCryptFS` which is a means of encrypting your home directory, things will 
 work.  You will need to move searchfox to a partition that's a normal block device
 (which includes LUKS-style encrypted partitions, etc.)
 
-##### Ubuntu 19.10
+##### Ubuntu
 
 ```shell
 # make sure the apt package database is up-to-date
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -264,62 +201,80 @@ steps in the next section.
...
-which will build an index using the latest mozilla-central version with
-searchfox artifacts.
-
-### Testing basic changes
-
-Note: You can also just do `make build-mozilla-repo` in `/vagrant` to have it
-idempotently do the following for you.
-
+- If you did run a custom m-c try job, the only difference is the addition of a
+  `--setenv TRYPUSH_REV=therevision` to to the command.  So this looks like:
```
staktrace, line 249: Minor nit: it would be good to specify `therevision` is the full 40-char hash. I think using the abbreviated hash won't work because taskcluster doesn't allow indexing artifacts by abbreviated hash.

staktrace (commented):
Nice, thanks!

README.md
asutherland: Good point, thank you!  Addressed in https://github.com/mozsearch/mozsearch/pull/450 hopefully!

PR #450, opened 2021-09-11
Guidance to use the full 40char hash

asutherland:
Address https://github.com/mozsearch/mozsearch/pull/448#pullrequestreview-751374041


staktrace:
Thanks!

PR #451, opened 2021-09-13
Bug 1730294 - channel-tool: load balancer helper

asutherland:
Added channel-tool.py, a tool that adds the ability to:
- "list" known channels, including what pieces are set up, plus display
  the instances associated with them in a pretty UI.
- "inspect" a specific known channel to see its underlying AWS
  settings.
- "add" a new channel.  Under the hood this is called "ensure" because
  I've tried to make it be idempotent, but "ensure" is not a command
  name people would expect
- "remove" an existing channel.  Note that an unfortunate limitation is
  that because of how de-registering an instance from a target group
  takes time, this will error out if an instance is still in the process
  of being de-registered.  At least, this happened to me when using
  "move-server" (below).  Maybe if the instance is terminated, the
  mechanism can handle a more prompt deletion?
- "move-server": move a server between (existing) channels, handling
  target groups and tags.

### DIFF
--- a/docs/aws.md
+++ b/docs/aws.md
@@ -100,7 +100,7 @@ THEEND
 
 virtualenv --python=python3 env
 source env/bin/activate
-pip install boto3 awscli mozilla-aws-cli-mozilla
+pip install boto3 awscli rich mozilla-aws-cli-mozilla
 # Make sure that we have an up-to-date version of certifi for certificate
 # validation.  See comments in build-lambda-indexer-start.sh for more context.
 pip install --upgrade certifi
--- a/infrastructure/aws/channel-tool.py
+++ b/infrastructure/aws/channel-tool.py
@@ -1,0 +1,778 @@
+#!/usr/bin/env python3
+
+# Elastic Load Balancer (ELB) support code to support dynamic use of ELB
+# instances for dev channels so that we don't need to leave them existing all
+# the time, as there is a carrying cost to them.
+#
+# The general desired workflow is:
+# - trigger-web-server.py creates a dev channel on demand if it doesn't already
+#   exist.  This includes doing everything documented at
+#   https://github.com/mozsearch/mozsearch/blob/master/docs/aws.md#creating-additional-development-channels
+#   automatically.
+# - terminate_indexer.py destroys a dev channel by default when terminating a
+#   web-server.
+#
+# Probably fine logistical details:
+# - The ELB gets a random internal DNS name which means that we have to
+#   dynamically update the subdomain even if it already existed.  This means
+#   that it's preferable for us to delete DNS entries when not using a channel.
+#
+# Disclaimers:
+# - I'm really struggling naming-wise in python, using a very inconsistent
+#   mixture of snake_case and camelCase.  Sorry!
+
+
+import argparse
+from datetime import datetime
+import re
+import sys
+
+import boto3
+from rich import box, print
+from rich.console import Console
+from rich.markup import escape
+from rich.table import Table
+from rich.tree import Tree
+from rich.traceback import install
+
+console = Console()
+
+# These are documented in the `HelperResult` doc-block below.
+HANDS_OFF_RELEASE_CHANNELS = 1
+NO_SUCH_CHANNEL = 2
+NO_SUCH_INSTANCE = 3
+WEIRD_CHANNEL = 4
+
+class HelperResult:
+    '''
+    Hacky attempt for the helper to provide suggested error codes and maybe some
+    additional details, even though the helper will print out a tremendous
+    amount of information itself.
+
+    We probably should be using exceptions instead.
+
+    Error codes:
+    - 1: You're trying to mess with a release channel and we don't like that.
+    - 2: You named a channel that doesn't exist and a channel does need to
+         exist.
+    - 3: You named an instance that doesn't exist.
+    - 4: The channel's state is inconsistent and so can't be used for the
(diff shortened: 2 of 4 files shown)


asutherland:
I'm going to land this now since the tool is currently fully standalone, works, and my weekend hacking is at an end.  Integration as proposed in https://bugzilla.mozilla.org/show_bug.cgi?id=1730294 (which was motivated by https://bugzilla.mozilla.org/show_bug.cgi?id=1730293) is not yet here.  I'll be leaving bug 1730294 after landing this.

As I note in the commit message above, the target de-registration may or may not end up being a serious issue for terminate-indexer.py to use this.  I'll have to see what the implications are of termination versus de-registering a still-alive instance.

@staktrace Feel free to use "move-server" against instance `i-0a55e8acd868a9f0d` on the `asuth` channel and to "remove" the `asuth` channel in any testing/experimentation you do.  I'm planning to terminate the server and remove the channel when I'm done.

I was able to create a fresh "fancy" channel with `./channel-tool.py add fancy` and then `./channel-tool.py move-server i-0a55e8acd868a9f0d fancy` and other than needing to wait ~1min for the DNS to show up, https://fancy.searchfox.org/ worked immediately.  `./channel-tool.py list` then showed it in the correct channel, but also correctly listing that it was in 2 target groups for several minutes.  I was then able to move it back to `asuth` via `./channel-tool.py move-server i-0a55e8acd868a9f0d asuth`.  And then `./channel-tool.py remove fancy` worked after target deregistration completed.

asutherland:
Also, I do want to ideally make sure we're on the same page about doing this dynamically.  I think it makes sense for us to create/delete the load balancers on demand, but I'm not intending to be unilaterally making that policy decision!

us-west (oregon pricing) at https://aws.amazon.com/elasticloadbalancing/pricing/ is currently $0.0225 per hour which is ~$16/month (0.0225 * 24 * 30 == 16.2).  (For comparison, our current web-servers are t3.large which has an on-demand rate of $0.0832/hr and a reserved effective rate of $0.052/hr from $38.11 monthly cost per https://aws.amazon.com/ec2/pricing/reserved-instances/pricing/.

asutherland:
Testing note: Gotta run `pip install rich` in your environment for this to work.  I did update the docs for the checkout and included this in the provisioning scripts, but have no yet re-provisioned anything yet.

staktrace:
Wow, this is pretty awesome!

staktrace:
I used this to delete the `kats` channel - it first deleted the lb, but then errored out when deleting the target group, because:
```
ResourceInUseException: An error occurred (ResourceInUse) when calling the DeleteTargetGroup operation: Target group
'arn:aws:elasticloadbalancing:us-west-2:653057761566:targetgroup/kats-target/f1847a04b63d357d' is currently in use by a listener or a rule
```

But I think that was just because the deletion of the lb hadn't propagated yet. I waited a bit and re-ran the command and it managed to then delete the tg and dns stuff.

staktrace:
The initial part of the output from the first `channel-tool remove` command is this:

```
$ ./infrastructure/aws/channel-tool.py remove kats
Located channel load balancer ARN: arn:aws:elasticloadbalancing:us-west-2:653057761566:loadbalancer/app/kats-lb/3a8cc65236d52cb7
Deleted load balancer (and listeners)!
{
    'ResponseMetadata': {
        'RequestId': 'a8aec742-e5f8-4594-8c2d-e2d6b32f952d',
        'HTTPStatusCode': 200,
        'HTTPHeaders': {'x-amzn-requestid': 'a8aec742-e5f8-4594-8c2d-e2d6b32f952d', 'content-type': 'text/xml', 'content-length': '262', 'date': 'Tue, 14 Sep 2021 00:47:08 GMT'},
        'RetryAttempts': 0
    }
}
Located channel target group ARN: arn:aws:elasticloadbalancing:us-west-2:653057761566:targetgroup/kats-target/f1847a04b63d357d
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
```

And as far as I can tell from [the docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elbv2.html#ElasticLoadBalancingv2.Client.delete_load_balancer) there's no indication that this operation is async so I don't know if there's anything we can do differently here, except maybe poll to wait until the lb is actually gone.

asutherland:
Interesting!  Had you done any `move-server` operations that could have meant the target group was still in the process of de-registering an instance?  (This would show up in the "list" command.)  I had assumed that was the cause before and written some comments to that effect, but it makes sense that the spin-down of the load balancer could take extra time given the other docs discussing how it uses a piece of an "m" instance somehow.

Either way, it does sound like actively polling for either the load balancer to go away (if there are direct indications of its continued existence) or retrying the target group removal is appropriate, as that's a much better UX than repeatedly invoking the command given how I've erred on having the commands be very spammy in terms of output.  (Like showing the locals on the rich tracebacks results in potentially a bunch of scrolling, but being able to have that info in a log in the future I think will be invaluable if something goes wrong, as it basically snapshots the entirety of what's going on.)

I'll plan to add some kind of polling for that in the next follow-up.  (That said, I won't be touching this before Saturday, so if you want to hack on any aspect of this and land changes/etc., feel free)

staktrace:
> Interesting! Had you done any move-server operations that could have meant the target group was still in the process of de-registering an instance?

Nope, I just did a `list` and then a `remove`. And yeah I kind of like the rich tracebacks - although verbose it is good to have that info rather than too little.

PR #462, opened 2021-11-08
Bug 1727453 - Migrate to serde_json

asutherland:
This is a migration to serde_json from rustc_serialize and undertakes
some related cleanups/refactorings when doing so.  It can be handy to
read / consult the following sources of information for context:
- https://docs.serde.rs/serde_json/ at the top-level has a great intro
  to `json!` and related helpers.
- https://serde.rs/ while notionally about serde proper has many
  specifics relating to serde_json.  The "attributes" and "examples"
  documentation is exceedingly relevant.

The following things are going on in this patch:
- Mechanical translations from the rustc_serialize wrappers to just
  using `json!`.  In some cases I've also gotten rid of intermediary
  maps because `json!` has really great ergonomics.
  - There are probably places we could use `json!` more.  In some
    cases I punted because I didn't see a way to conditionally omit
    some key/value pairs trivially, but presumably those could just
    be optionally mutated onto the result after.
- Migration to using serde's Serialize and Deserialize directly, for
  everything.
  - Initially I was going to be more conservative, potentially only
    using Serialize, but this ended up seeming foolish and also more
    confusing when implementing (de)serialization helpers.
  - Previously we were manually converting from rustc_serialize's
    parsed JSON value rep, and then we implemented `fmt::Display`
    traits for output.  serde's automation shines at everything we
    were doing here through annotations and helpers.
  - Got rid of additional representation structs that could instead
    be represented by a single struct.  In particular, crossref was
    using reference-counted types that required a lot of labor in the
    code.
  - Searchfox's use of `"source": 1` and `"target": 1` in the JSON to
    convey the type of record that was being used gets a little bit of
    a hack.
    - serde has a nice documentation page on enum representations at
      https://serde.rs/enum-representations.html and one might note
      that what searchfox has been doing doesn't really fit into any
      (other than "untagged").  Adjacently tagged sorta seems like it
      might work if we flattened the adjacent payloads, but our use of
      the `read_` functions as filters that themselves expect to see
      the tags made that a little weird.  If we always used an enum
      type instead of the distinct types, this could be moot, but that
      also potentially runs into `WithLocation` issues.  (I maintained
      existing `WithLocation` behaviour through use of `flatten`,
      although maybe that should be rethought at some point.)
      - I did consider changing the reps we use, but that would be a
        logistical hassle to update all the m-c and c-c branches.
    - We use a single-valued enum for each type which theoretically
      doesn't use storage space but does need to be listed when
      instantiating the struct.
- Switched to using the `Ustr` https://github.com/anderslanglands/ustr
  automagically reference-counted string type directly in our types
  where appropriate.
  - I did a moderately extensive survey and settled on Ustr primarily
    because it already had serde integration and seemed mature in
    terms of continued updates as well as having benchmarks.
  - The primary caveat to ustr is that all interned strings are
    permanently interned and never freed, so long-lived processes that
    potentially see a lot of strings need to be sure not to use Ustr
    for strings that will never be seen again.  To this end I tried to
    be responsible in my choices and at the transition layer, but it's
    conceivable we could see ramifications.
  - It does feel like this has really helped simplify the crossref.rs
    logic and so while I'm not wedded to Ustr, I'd like to stick with
    something with similar ergonomics.
- `peekLines`, a multi-line string, in the crossref output has been
  changed back to its source data of `peekRange` because we never ended
  up using `peekLines` in production and I think it likely makes sense
  to switch to using lol_html to instead filter the statically
  generated HTML instead.  Should that not be sufficiently performant,
  we'd switch to having direct source file offsets with any necessary
  state to put the tokenizer in the right state.
  - "fancy" branch prototyping seemed to show that the lack of syntax
    highlighting was a real problem.
  - Part of why this never got used was I was trying to convert us to
    using offsets, but ended up in a hyper.rs-upgrade nightmare
    situation back during the original searchfox work week when :billm
    handed off searchfox.
- `AnalysisStructured` got less hacky!  Previously we only extracted a
  few fields and re-emitted the rest as a JSON string `payload` which
  we would then glue together again when formatting it for output.  Now
  we use serde's ability to `flatten` all the extra values that don't
  have direct struct fields into its value rep.
  - I didn't feel great about the original `payload` mechanism, but it
    seemed least bad versus the heavy-seeming rustc_serialize Value
    reps.  (Heavy both in terms of memory usage and in terms of hassle
    to work with them.)
  - Given the conversion to using the more structured serde rep, it no
    longer made any sense to try and keep this up and likely would have
    been a total nightmare.
  - Many of `AnalysisStructured`'s fields are now mapped directly into
    rust structs.  Serde makes this easy and this is consistent with my
    conclusions from the "pretty" branch that for performance we need
    to be doing any fancy stuff in the back-end where it can be
    (partially) pre-computed and/or cached (with cached values possibly
    begetting pre-cached values in future indexer runs, making them
    pre-computed).  The JS single-page-app thing was not fast enough to
    be considered Searchfox brand searchfoxing!
- I added allow-dead-code as appropriate.

### DIFF
--- a/tests/tests/checks/inputs/merge__big_cpp
+++ b/tests/tests/checks/inputs/merge__big_cpp
@@ -1,1 +1,1 @@
-merge-analyses big_cpp.cpp -p foo-plat
+merge-analyses big_cpp.cpp big_cpp.cpp -p foo-plat -p bar-plat
--- a/tests/tests/checks/inputs/merge__big_cpp
+++ b/tests/tests/checks/inputs/merge__big_cpp
@@ -1,0 +1,1 @@
+merge-analyses big_cpp.cpp -p foo-plat
--- a/tests/tests/checks/snapshots/check_glob@merge__big_cpp.snap
+++ b/tests/tests/checks/snapshots/check_glob@merge__big_cpp.snap
@@ -1,0 +1,3653 @@
+---
+source: tests/test_check_insta.rs
+expression: "&json_results"
+
+---
+[
+  {
+    "loc": "00001:0",
+    "target": 1,
+    "kind": "def",
+    "pretty": "big_cpp.cpp",
+    "sym": "FILE_big_cpp@2Ecpp"
+  },
+  {
+    "loc": "00020:9-23",
+    "target": 1,
+    "kind": "use",
+    "pretty": "big_header.h",
+    "sym": "FILE_big_header@2Eh"
+  },
+  {
+    "loc": "00021:9-50",
+    "target": 1,
+    "kind": "use",
+    "pretty": "subdir/header@with,many^strange~chars.h",
+    "sym": "FILE_subdir/header@40with@2Cmany@5Estrange@7Echars@2Eh"
+  },
+  {
+    "loc": "00022:9-23",
+    "target": 1,
+    "kind": "use",
+    "pretty": "atom_magic.h",
+    "sym": "FILE_atom_magic@2Eh"
+  },
+  {
+    "loc": "00024:6-19",
+    "target": 1,
+    "kind": "def",
+    "pretty": "GlobalContext",
+    "sym": "T_GlobalContext",
+    "peekRange": "24-24"
+  },
+  {
+    "loc": "00027:14-32",
+    "target": 1,
+    "kind": "def",
+    "pretty": "GlobalContext::decideBooleanTrait",
+    "sym": "_ZN13GlobalContext18decideBooleanTraitEv",
+    "context": "GlobalContext",
+    "contextsym": "T_GlobalContext",
+    "peekRange": "27-27"
+  },
+  {
+    "loc": "00028:15-19",
+    "target": 1,
+    "kind": "use",
+    "pretty": "rand",
+    "sym": "rand",
+    "context": "GlobalContext::decideBooleanTrait",
+    "contextsym": "_ZN13GlobalContext18decideBooleanTraitEv"
+  },
+  {
+    "loc": "00034:24-32",
+    "target": 1,
+    "kind": "use",
+    "pretty": "RAND_MAX",
+    "sym": "M_85a652bdd3e78849"
+  },
+  {
+    "loc": "00063:14-47",
+    "target": 1,
+    "kind": "def",
+    "pretty": "GlobalContext::decideEnigmaticAnimalBooleanTrait",
+    "sym": "_ZN13GlobalContext33decideEnigmaticAnimalBooleanTraitEv",
+    "context": "GlobalContext",
+    "contextsym": "T_GlobalContext",
+    "peekRange": "63-63"
+  },
+  {
+    "loc": "00064:11-29",
+    "target": 1,
+    "kind": "use",
+    "pretty": "GlobalContext::decideBooleanTrait",
+    "sym": "_ZN13GlobalContext18decideBooleanTraitEv",
+    "context": "GlobalContext::decideEnigmaticAnimalBooleanTrait",
+    "contextsym": "_ZN13GlobalContext33decideEnigmaticAnimalBooleanTraitEv"
+  },
+  {
+    "loc": "00067:14-35",
+    "target": 1,
+    "kind": "def",
+    "pretty": "GlobalContext::decideCatBooleanTrait",
+    "sym": "_ZN13GlobalContext21decideCatBooleanTraitEv",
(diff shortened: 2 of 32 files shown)


asutherland:
Comment 0 is currently the commit message.  Limited extra context that doesn't go in there:
- I haven't done a try indexing against mozilla-central yet, I'll do that sometime this week.  In-repo tests do pass (with minor snapshot updates because of field reordering and s/peekLines/peekRange/.)
- I'm planning to land these this weekend/Wellness Friday as long as the try indexing seems like it doesn't break things.  Per our modern convention (and now with better tests), I'm not treating these reviews as blocking as long as it seems to not break things.  That said, I very much value any/all feedback and all the better if it comes before landing so I can improve things, but I'm also happy for post-landing feedback too.  (Also, to be clear, if there are issues of concern that make anyone think this shouldn't be landed without changes and subsequent review, I'm also on board with that.)
- :mccr8 I added you here in particular because you originally implemented the original reference-counting in crossref.rs and likely have some valuable thoughts about Ustr or other candidates.
  - One important note is that I think it's now a non-goal to ensure that people can index mozilla-central locally in their VM.  Which is mainly to say that if my changes to reference counting wildly regress memory usage on crossref for mozilla-central, that's probably a problem, but the bar is no longer the vagrant VM, but the m5d.2xlarge's 32 GiB, so slight regressions are probably fine.

staktrace (approved):
Wow, this is a pretty big cleanup! I didn't expect the changes to be so involved, thanks for doing this!

emilio (approved):
Weird that ustr requires parking-lot, but I guess that's alright.

This is indeed an awesome cleanup :)

asutherland:
As somewhat expected owing to the lack of test coverage, merge-analysis failed on the config1 indexer run on https://searchfox.org/mozilla-central/source/__GENERATED__/__RUST_BUILD_SCRIPT__/neqo-crypto/nspr_error.rs.  I feel like I've previously tested merge-analysis by running a file against itself to good effect but my immediate searches don't show evidence of something I can copy and paste for this.

The idea is straightforward though, so I'll see about adding basic coverage to the Makefile and a follow-up for more extensive coverage.  It looks like the failure was on an identical-across-platforms file (though not the first one of these), so the merge-against-self scenario should be sufficient, and the "build some stuff with different defines" would go in the follow-up.

I will potentially do a small amount of refactoring of merge-analyses.rs in order to make it possible to be used by the test_insta_glob/searchfox-tool test infrastructure as it seems like it should be straightforward-ish and certainly much more pleasant than doing shell scripting stuff.

### Error info

Relevant excerpts from the email I got:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("can only flatten structs and maps (got a sequence)", line: 0, column: 0)', src/bin/merge-analyses.rs:155:43
```

with the only interesting stack frame:
```
  17:     0x559aab582a8f - merge_analyses::main::h3af2fb05cdb08423
                               at /home/ubuntu/mozsearch/tools/src/bin/merge-analyses.rs:155:22
```

```
Generated file ./__RUST_BUILD_SCRIPT__/neqo-crypto/nspr_error.rs was identical across platforms where it was created
...
parallel: This job failed:
/home/ubuntu/config/shared/collapse-generated-files.sh ./__RUST_BUILD_SCRIPT__/neqo-crypto/nspr_error.rs
```

asutherland:
Okay, I've now added coverage; the problem was as discussed in the https://github.com/mozsearch/mozsearch/pull/462/commits/a90decf9eda72602151cc9cd6ac3a53f29759404 commit message (excerpted below).  I'll be running this against m-c and expect it to succeed and then I expect to land it.

A nuance I'd missed was that there were cases where we have
`WithLocation<Vec<T>>` as an aggregate where the JSON serialization
then produces a separate JSON object for each item in `Vec<T>`.  This
doesn't work with a bidirectional JSON mapping and so I've changed the
merger logic to manually unroll this scenario.

asutherland:
The dev.searchfox.org run completed successfully with all indications of merge logic running successfully.  It seemed to correct all intended warnings, but we have 2 new warnings that I believe are probably due to the lol_html dependency and I'll look into bumping things if possible:

```

warning: the following packages contain code that will be rejected by a future version of Rust: cssparser v0.25.9, selectors v0.21.0
warning: the following packages contain code that will be rejected by a future version of Rust: cssparser v0.25.9, selectors v0.21.0
```

PR #463, opened 2021-11-19
Minor CSS clean-ups.

emilio:
This simplifies quite a bit the implementation of the code coverage styling, and removes some other unused CSS variables and so on.


asutherland (approved):
Thanks for the clean-ups!  Hopefully this makes things go faster too!  The performance hit was noticeable before with a11y on, I think.

PR #464, opened 2021-11-19
Dark mode

emilio:
This is on top of #463.

Simpler, hopefully more palatable implementation of dark mode than #415.

The main trick that allows this to be so tiny is that we rely on the
browser's form controls (using color-scheme: dark).

The other main trick is including the copy SVG icons in the source
directly, so that we don't have to have multiple versions of them and
they can just use `fill: currentColor`.

### DIFF
--- a/scripts/output.js
+++ b/scripts/output.js
@@ -65,7 +65,8 @@ function generate(content, opt)
   let output = `<!DOCTYPE html>
 <html lang="en-US">
 <head>
-  <meta charset="utf-8" />
+  <meta charset="utf-8">
+  <meta name="color-scheme" content="light dark">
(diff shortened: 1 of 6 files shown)


emilio:
Btw, I didn't make an opt-out out of this. If people complain we can fairly trivially add one or remove the feature until a proper switch exists, but I think I'd rather avoid it if possible.

emilio:
cc @mystor

emilio:
I triggered a dev run with all these changes so it should be up at https://dev.searchfox.org/ in a bit.

asutherland (approved):
Woo!  Thank you!

PR #466, opened 2021-11-22
Bug 1742145 - Dark mode: code coverage colors.

emilio:


asutherland (approved):
Thanks for the quick follow-up, the clean-up, and the updated/revised comments!

PR #467, opened 2021-11-22
Dark mode: Fix expando arrow color.

nhnt11:


static/css/mozsearch.css
```diff
@@ -75,9 +78,12 @@ table {
 @media (prefers-color-scheme: dark) {
   :root {
     --page-background: rgb(28, 27, 34);
-    --page-foreground: rgb(251, 251, 254);
+    --page-foreground: rgb(210, 211, 212);
```
emilio, line 81: As per the bug comment, these come from our in-content pages so I'd rather keep foreground and background as-is unless there's a good reason not to do that.

static/css/mozsearch.css
```diff
@@ -127,6 +133,9 @@ body {
 h4 {
   margin: 0 0 0.5rem 0;
 }
+a {
+  color: 
+}
```
emilio, line 138: Please remove this?

static/css/mozsearch.css
```diff
@@ -430,7 +439,7 @@ tr.after-context-line + tr.before-context-line {
 
 .expando {
   cursor: pointer;
-  color: black;
+  color: var(--page-foreground);
```
emilio, line 442: Can this just be currentColor or inherit instead?

static/css/mozsearch.css
```diff
@@ -497,15 +506,15 @@ tr.after-context-line + tr.before-context-line {
 .bubble a:visited,
 .result-head a:link,
 .result-head a:visited {
-  color: #0095dd;
+  color: var(--link-color);
 }
 .breadcrumbs a:hover,
 .result-head a:hover {
-  color: #00539f;
+  color: var(--link-hover-color);
```
emilio, line 513: https://github.com/mozsearch/mozsearch/pull/468 has some additional clean-ups here, might be worth incorporating, but happy to do them in a follow-up instead.

emilio (changes requested):
Thanks!

static/css/mozsearch.css
nhnt11: Hmm, ok, that's fair. FWIW GitHub also has a lighter foreground color value similar to Bugzilla. I personally find it more pleasant but I get the design system consistency aspect.

static/css/mozsearch.css
nhnt11: Apparently not. Inherit/currentColor gets the color from a .left-column element which also holds line numbers and is colored #aaa.

static/css/mozsearch.css
nhnt11: done

static/css/mozsearch.css
nhnt11: If you're working on this actively please feel free to even reject this and e.g. only land the expando change. I might not check back here for a while once I jump into meetings.

static/css/mozsearch.css
nhnt11: reverted

static/css/mozsearch.css
emilio: Ok, I'll land that and rebase your branch to land the expando change. That way we probably get it on the next indexer run.

PR #468, opened 2021-11-22
Bug 1742365 - Dark mode: better link colors.

emilio:
Also simplify link styling while at it. We don't have links in bubbles
and :link+:visited is the same as :any-link.

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -37,6 +37,9 @@ table {
   --button-hover-background: white;
   --button-active-background: #d5d5d5;
 
+  --link-color: #0095dd;
+  --link-hover-color: #00539f;
+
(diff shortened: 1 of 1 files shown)


emilio:
Merging this to rebase #467 on top. @asutherland I think this should be really straight-forward but happy to address issues in follow-ups if you comment :)

asutherland:
> Merging this to rebase #467 on top. @asutherland I think this should be really straight-forward but happy to address issues in follow-ups if you comment :)

Yeah, these changes make sense and this is very much consistent with the approach :kats and I have been taking of post-landing review being fine for most things.  I think the only exception in this area would be for things that would impact a11y; and even in that case I think it's a question of someone verifying NVDA's behavior remains unchanged and/or that the a11y tree looks the same or is believed to be the same.  (Ideally we can further expand the insta-based checks to provide explicit coverage for this somehow.)

PR #469, opened 2021-11-24
Bug 1742820 - Dark mode: Use darker colors for highlights.

emilio:
So that they contrast better with comments and so.


emilio:
image

PR #470, opened 2021-12-02
Bug 1744153 - stuck line box-shadow wants padding

asutherland:
The box-shadow for stuck lines makes it hard to read the directly linked lines.

This patch adds an additional line of scroll padding so that the box-shadow can't possibly
interfere with the line.

The one caveat to this fix is that when directly linking to the first level of nesting, like
`namespace mozilla {`, we will add an extra line of padding.  This is because we are
choosing to favor the case where there's 2 stuck lines (like a nested namespace) and
avoid adding another selector or crazy math.  If only calc supported a
[signum function](https://en.wikipedia.org/wiki/Sign_function)!

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -698,14 +698,21 @@ span[data-symbols].hovered {
 */
 .goto {
   /* This needs to be large enough so that the "position: sticky"
-   stuck lines don't obscure the line we're trying to display. */
-  scroll-margin-top: calc((var(--nesting-level, -1) + 1) * 12px * 1.3);
+     stuck lines don't obscure the line we're trying to display.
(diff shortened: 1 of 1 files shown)


asutherland:
Feel free to supersede this with a better approach to https://bugzilla.mozilla.org/show_bug.cgi?id=1744153!

emilio (approved):
Seems reasonable! Alternatively we could just put the shadow color behind a variable and make it different in light vs. dark, your call. Either way works for me.

asutherland:
Let's try this for now, as I do think the increased demarcation between the stuck lines and the content is probably helpful and I don't think anyone has shown any concern about the increased clarity.

PR #471, opened 2021-12-15
context-menu: Improve giant context menus / context menus towards the bottom of the page.

emilio:
This patch prevents context menus from overflowing the viewport by
making them scrollable, and also positioning them towards the top if you
click down enough the page.

This could be improved in the future, but seems like a trivial
improvement to me.

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -494,10 +494,16 @@ tr.after-context-line + tr.before-context-line {
   padding: 0;
   border: 1px solid var(--page-border-color);
   border-radius: 6px;
-  border-top-left-radius: 0;
   list-style: none;
   z-index: 102;
   overflow: auto;
+  transition: opacity .2s ease;
+}
+.context-menu:not(.bottom) {
+  border-top-left-radius: 0;
+}
+.context-menu.bottom {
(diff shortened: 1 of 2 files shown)


asutherland:
xref https://bugzilla.mozilla.org/show_bug.cgi?id=1727557

asutherland (changes requested):
This seems like a major regression for people's muscle memory by doing this on the bottom 5th of the viewport as the mouse now has to be moved a non-trivial distance to reach the top menu item that is only a pixel away when you're on the top 80% of the viewport.  It seems reasonable if it's only when we're at the bottom of the actual document itself and there's nowhere for the user to scroll, though as noted in https://bugzilla.mozilla.org/show_bug.cgi?id=1727557 I think we could probably just add padding to the bottom of the document to help improve that situation too.

emilio:
@asutherland I implemented the flipping as discussed, can you take a look? Thanks! :)

asutherland (approved):
I'm **flipping** out over how great these changes are! ;)

I put this on dev.searchfox.org and this works great.  I also very much appreciate the elegant logic about when to constrain versus flip!  Thank you!

PR #472, opened 2021-12-16
context-menu: Make context menu scrollable.

emilio:
This is the part of #471 which should not mess with people's
muscle-memory :-)


asutherland (approved):
Also per discussion in the channel I'm r+ on just having the mirroring based on dynamic detection of the menu otherwise going off the screen.

emilio:
Yep, this is useful regardless, for context menus that are very big, so let's get this landed (I'll try to implement the mirroring today or tomorrow).

PR #473, opened 2021-12-16
Bug 1746071 - Add dark mode backgrounds for diff view.

emilio:
To choose the colors, I used the light mode colors, saturated them, then
made them semi-transparent until they looked the same with 0.17 alpha
(somewhat arbitrary), and applied them to the dark background.

image


emilio:
Merging as it's hard to look worse than we do now, but any tweak or improvement to the colors welcome. cc @mystor @asutherland

mystor:
Looks much nicer than what we had before, thanks!

PR #478, opened 2021-12-28
Bug 1747289 - Step 2 of automated provisioning.

asutherland:
- Triggering VMs still don't use the new AMIs, but...
- We now probably wait for the AMI snapshot process to take effect.
- We now probably rotate tags correctly.
- We now probably terminate the VM afterwards (with a grace period).
- I've set the lol_html version to a specific revision now that was the
  one picked when I used the branch specifier.  I did this because I
  noticed the Cargo.lock changed and that's not really a benefit for
  us; I hadn't appreciated how easily the latching would update.


asutherland:
About to run this after landing.

PR #485, opened 2022-01-09
Additional doc fixes

staktrace:
Follow-up to the documentation updates from bug 1747289

### DIFF
--- a/docs/aws.md
+++ b/docs/aws.md
@@ -347,7 +347,7 @@ our shell scripts:
   second time itself; we no longer support re-provisioning the dev VM manually.
   (Instead, the VM should be destroyed and rebuilt.)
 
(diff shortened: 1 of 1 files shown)


asutherland (approved):
Oh wow, I hadn't realize this step existed!  I was assuming they would go way when the AMI was de-registered!  Thanks for documenting this!  (And my apologies if this was previously documented elsewhere and my eyes glazed over, missing it...)

edit: A few minutes after reading this I think a memory of having to do this came back... clearly at some point I thought I had mastered the steps required but in fact was only remembering pieces of what to do!

asutherland:
I just deregistered the older 20.04 images so we're down to 4 AMIs and performed the steps and they worked great.  Thanks again!

PR #486, opened 2022-02-17
Bug 1749176 - Add missing test snapshots.

asutherland:
A current limitation of the test infrastructure is that it's very easy to accidentally
only add a single new test snapshot at a time because `make build-test-repo` stops at
the first failure.

Here are the new missing test snapshot values.


asutherland:
These were missing from https://github.com/mozsearch/mozsearch/pull/483

PR #488, opened 2022-03-07
Bug 1755807 - Add XPIDL Test Coverage and Fix XPIDL analysis.

asutherland:
## Test Coverage

Add XPIDL data to the test corpus that's partially extracted from
mozilla-central and can be updated via the new
`refresh-mozilla-central-stuff` script from a running config1 web
server.

The rationale behind this approach is that the XPIDL generation process
is non-trivial and likely to break if directly embedded in this repo,
so we just copy over the byproducts that we care about.  These will
also form the basis for our production m-c checks, which should ideally
make it easier to keep up-to-date because if one of them breaks, they
both should break (at least after running
`refresh-mozilla-central-stuff`.

## The Problem and Fix

Our source records for IDL files did not include "syntax" keys/values,
and these are necessary for the serde-based serialization to decide
that a source record is valid.

Our target records for IDL files were also missing "pretty" fields
which didn't really matter but is necessary for the `filter-analysis`
searchfox-tool command to be able to do an `-i` identifier match on
things, so it now does that too.

This provides for semantically highlighting the symbol in source files
and providing a "search" option.  It also provides a very useless "go
to definition" option which is the C++ definition which no one would
ever want.  But I think that's largely a longstanding issue and I have
made various exciting proposals in bug 1727789 that I think likely is
how this will be addressed, although maybe we can hack up a specialized
fix for XPIDL for this case.


scripts/idl-analyze.py
```diff
@@ -272,6 +286,9 @@ def handle_interface(analysis, iface):
         print('Syntax error in IDL', fname, file=sys.stderr)
         raise e
         sys.exit(1)
+    print('XPIDL: Parsed', fname, 'into', len(r.productions), 'productions', file=sys.stderr)
     for p in r.productions:
```
emilio, line 290: Did you want to keep these?

emilio (commented):
Looks good, thanks!

scripts/idl-analyze.py
asutherland: Yeah, the lack of output from the `idl-analyze.py` step was a hindrance in setting this up and has also been a hindrance in previous variations of XPIDL (I think due to merge-analyses making the non-identical files per-platform) and IPDL failures (which is an ongoing problem of rust/python parser inconsistencies, although a different analyzer entirely).  Not a major hindrance, but being able to grep/search for a file of interest in the index log and get (or informatively not get!) a hit that explains what happened in the pipeline I hope will make analyzing failures even easier for people who might ramp up in the future on searchfox.

asutherland: Whoops, forgot the example I wanted to paste.  Here's what `make build-tests-repo` provides:

```
+ cat /home/vagrant/index/tests/idl-files      
+ parallel /vagrant/scripts/idl-analyze.py /home/vagrant/index/tests '/home/vagrant/index/tests/files/{}' '>' '/home/vagrant/index/tests/analysis/{}'                                         
XPIDL: No C++ analysis data found for /home/vagrant/index/tests/files/xpidl/nsISupports.idl                                                                                                   
XPIDL: No C++ analysis data found for /home/vagrant/index/tests/files/xpidl/nsrootidl.idl                                                                                                     
XPIDL: Parsed /home/vagrant/index/tests/files/xpidl/xpctest_params.idl into 5 productions                                                                                                     
+ echo 0 
```

PR #494, opened 2022-03-21
Bug 1749232 - Implement "traverse" command with graph support infra

asutherland:
Note that all new functionality in this patch is not exposed to the
web and because RemoteServer::crossref_lookup is a no-op, the
functionality can only be used against a local index.

This implements:
- A basic graph abstraction in `symbol_graph.rs`.  petgraph is fairly
  low level and is not geared towards storing heavy-weight metadata
  with nodes or edges so we add some decidedly non-clever wrappers
  but which should largely be type-safe.
  - This includes serialization of the graph to JSON in a rep that
    tries to sort everything so that the test output and human
    inspection will be largely sane.  I did not optimize for
    performance/memory here; I used BTreeMap and BTreeSet and called
    it a day.
- A core "traverse" command which only implements "callees" right now.
- A single check that provides coverage of taking 2 input symbols'
  crossref data (OuterCat::meet is overloaded) and produces a graph
  that only has edges involving callable nodes.  I inspected the
  result and have verified it's correct.

Dependency-wise this:
- Adds petgraph without "serde-1" enabled because we don't actually
  need it because we're using a sufficiently custom rep that the
  petgraph native output is not useful and we now already know how
  to render our rep to reasonably sane JSON.

### DIFF
--- a/tests/tests/checks/inputs/traverse_callees__big_cpp__outercat_meet__json
+++ b/tests/tests/checks/inputs/traverse_callees__big_cpp__outercat_meet__json
@@ -1,0 +1,1 @@
+search-identifiers outerNS::OuterCat::meet | crossref-lookup | traverse
--- a/tests/tests/checks/snapshots/check_glob@traverse_callees__big_cpp__outercat_meet__json.snap
+++ b/tests/tests/checks/snapshots/check_glob@traverse_callees__big_cpp__outercat_meet__json.snap
@@ -1,0 +1,506 @@
+---
+source: tests/test_check_insta.rs
+expression: sgc.to_json()
+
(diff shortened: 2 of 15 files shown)


asutherland:
This is incremental progress towards graphing.  I'll be landing patches incrementally as the overall goal of graphing on https://bugzilla.mozilla.org/show_bug.cgi?id=1749232 is iterated towards.  Any potential changes to existing UX will be discussed on the chat.mozilla.org `#searchfox` channel and/or explicit bugs first, but this will largely be built as an optional/parallel capability first that we have people alpha and beta test on an opt-in basis, etc.

PR #497, opened 2022-03-28
sticky: Accurately detect stuck position.

emilio:
This fixes the issue Olli reported on Matrix.

### DIFF
--- a/static/js/code-highlighter.js
+++ b/static/js/code-highlighter.js
@@ -288,7 +288,7 @@ var Sticky = new (class Sticky {
     const jitter = 6;
 
     function extractLineNumberFromElem(elem) {
-      if (!elem.classList.contains("line-number")) {
+      if (!elem || !elem.classList.contains("line-number")) {
         return null;
       }
       let num = parseInt(elem.dataset.lineNumber, 10);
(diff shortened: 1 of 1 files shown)


asutherland (approved):
Eureka!  Paying attention to the displacement does seem like the right thing to do!  If you could update the 2 sets of block comments above that reference line number inference in some way, that would be great, but it's also likely you're the only person who's going to touch this code anytime soon, so I'm not particularly concerned.

I also like the idea of using intersection observer you mentioned on CMO.  For something like https://searchfox.org/mozilla-central/source/dom/indexedDB/ActorsParent.cpp `document.querySelectorAll(".nesting-container").length` is 861; is that something where it'd be okay to add observers to all of the nesting-containers, or would we want to do some kind of adaptive mechanism where we dynamically instrument containers?

asutherland:
I did get an exception on the insanely large https://searchfox.org/mozilla-central/source/__GENERATED__/security/rlbox/rlbox.wasm.c but I'm not sure if this is something that already happened:
```
Uncaught TypeError: can't access property "classList", elem is null
    extractLineNumberFromElem https://searchfox.org/static/js/code-highlighter.js:291
    walkLinesAndGetLines https://searchfox.org/static/js/code-highlighter.js:319
    handleScroll https://searchfox.org/static/js/code-highlighter.js:371
    Sticky https://searchfox.org/static/js/code-highlighter.js:248
21 code-highlighter.js:291:11
```

emilio:
> I did get an exception on the insanely large https://searchfox.org/mozilla-central/source/__GENERATED__/security/rlbox/rlbox.wasm.c but I'm not sure if this is something that already happened:
> 
> ```
> Uncaught TypeError: can't access property "classList", elem is null
>     extractLineNumberFromElem https://searchfox.org/static/js/code-highlighter.js:291
>     walkLinesAndGetLines https://searchfox.org/static/js/code-highlighter.js:319
>     handleScroll https://searchfox.org/static/js/code-highlighter.js:371
>     Sticky https://searchfox.org/static/js/code-highlighter.js:248
> 21 code-highlighter.js:291:11
> ```

Yeah that seems pre-existing, but fixed anyways.

asutherland:
I just updated the active web server to the master branch and because we haven't addressed the thing about the web-server checking out the latest revision rather than what the indexer was using, this will also be fixed when the current indexer finishes.

Thanks!

PR #499, opened 2022-04-13
Bug 1763005 - generate markdown explanations of test pipelines

asutherland:
This provides the backbone of the pipeline explanation mechanism, but
the templates in use right now just output the subscriber::fmt JSON
output for each logged event right now.  A few additional steps are
required to make this useful.

This patch can and will generate explanations parallel to the inputs
and checks sub-trees, but I haven't checked them in yet because they
are bad and include timestamps which means they would be a churn
nightmare.

Those next expected steps are:
- Add more tracing events, in particular for the server API calls.
- Have the events have some kind of semantic tag that helps the
  template figure out how to present things.
- Make the template use that semantic tag!
- Make sure the explanation markdown files are stable across
  multiple runs (no timestamps!).

In terms of what's been done in the patch:
- We add tracing as a dependency and do some bare bones event generation.
  - I added a bunch of Debug derives for the pipeline structures so that
    tracing's field mechanism could dump them to the logs in a readable
    fashion.  Another alternative we might still want to pursue is
    instead deriving Serialize so they can be encoded as JSON instead,
    but I'm not sure that helps comprehension in any meaningful way and
    it's a non-goal to allow populating those structures via Deserialize
    at this time.
  - I added a bunch of Serialize derives for the core pipline values so
    that they can easily be dumped as JSON.
    - In the case of SymbolGraphCollection, I hacked a trivial struct
      serializer that just reuses our existing to_json helpers.  This is
      not maximally efficient, but it doesn't need to be.  I did think
      about re-writing the to_json logic to just be custom serialization
      implementations, but that seemed like it would be much less
      readable and also potentially less performant for the actual hot
      paths and we would never serialize to anything but JSON anyways.
- We copy the tracing project's test MockWriter/MakeMockWriter as a means
  of letting us retrieve the contents of the JSON log output from an
  in-memory buffer without touching disk.  This is useful for the tests
  and will also be useful for the pipeline-server, ideally.
- For our tests, we use the dispatcher mechanism to try and set our
  subscriber to be active for the current test using a guard, which
  ideally should provide a means of us being able to have the subscriber
  follow task dispatches from our test tasks to other threads without
  having us scooping up logging from other uninvolved threads.
  - This approach comes from the advice at
    https://github.com/tokio-rs/tracing/issues/1517
  - It's not clear that I fully understand the ramifications of the
    threading model aspects of this, but it is clear that the guard
    approach is preferable to the FnOnce approach of subscriber::with_default
    https://docs.rs/tracing/0.1.32/tracing/subscriber/fn.with_default.html
    if only to avoid massive indentation problems and hassles related to
    our use of async functions!
  - In general I do expect parallelism from the pipline jobs to result in
    explicit new task schedulings, which seems like we can probably make
    the dispatcher approach work if necessary with glue logic.  I'm not
    expecting us to have any other kinds of pre-existing threads we'd be
    trying to communicate with, or if we did, it seems like a reasonable
    API / logging boundary.
  - A general motivation was that I understand cargo may try and run tests
    in parallel by default and I didn't want to build things in such a way
    that would aggressively preclude this.  I also was planning / hoping
    for this to work for the pipeline-server being able to easily log only
    for its own request.
  - Our general control and data-flow can all also just nest under a
    single root span or with some other distinct request identifier, so
    that's also a way we could pull the data back out in production.  I
    definitely have a lot of learning to do about `tracing` and am very
    excited to do so as it seems like an amazing, well-thought-out lib!
- We add liquid templating with a base template.
- We add a custom filter so we can render the liquid values as JSON;
  this is something that is notionally part of the language standard lib
  but is not yet in the rust implementation and it was not immediately
  obvious to me how to fashion an upstream-able fix, so I started with
  just getting it working at all.
- We add include_dir as a dep so that liquid's mechanism that use a VFS
  sorta setup like `render` and `layout` can do that and we don't have
  to do any legwork beyond just adding new files to `tools/templates`.
  - This isn't fully tested/used yet.
- We generate separate explanations for "local" and "remote" which
  makes sense because we expect the server ops to differ when we start
  logging those, but was also mainly just a hack because right now the
  "remote" checks run last and will frequently bail out because of
  no-ops (which cause the test to be skipped partway through) and so
  their explanations are fundamentally boring!

### DIFF
--- a/tools/Cargo.lock
+++ b/tools/Cargo.lock
@@ -47,6 +47,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anymap2"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d301b3b94cb4b2f23d7917810addbbaff90738e0ca2be692bd027e70d7e0330c"
+
+[[package]]
 name = "ascii-canvas"
 version = "3.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -227,7 +233,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9213f7cd7c27e95c2b57c49f0e69b1ea65b27138da84a170133fd21b07659c00"
 dependencies = [
  "num",
- "time",
+ "time 0.1.43",
 ]
 
 [[package]]
@@ -387,6 +393,12 @@ dependencies = [
  "redox_users",
  "winapi 0.3.9",
 ]
+
+[[package]]
+name = "doc-comment"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fea41bba32d969b513997752735605054bc0dfa92b4c56bf1189f2e174be7a10"
 
 [[package]]
 name = "dot-generator"
@@ -794,7 +806,7 @@ dependencies = [
  "log 0.3.9",
  "mime 0.2.6",
  "num_cpus",
- "time",
+ "time 0.1.43",
  "traitobject",
  "typeable",
  "unicase",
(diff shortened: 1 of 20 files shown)


asutherland:
This works locally and should not change anything outside of test byproducts.  I'm going to land this and trigger re-provisioning because this adds new cargo deps and we like to re-provision to make sure we're less likely to break because of crates.io having non-200 responses sometimes.

PR #505, opened 2022-05-01
WIP pipeline server, WIP "query" command, working `search-text` command

asutherland:
This landing include a bunch of progress towards the pipeline server that's still in progress, plus a working implementation of livegrep/codesearch logic via a new `search-text` command.

Because the search-text command includes a new submodule as described at https://bugzilla.mozilla.org/show_bug.cgi?id=1762992#c1 and a bunch of our provisioning logic hardcodes the "master" branch for provisioning, I need to land this to validate `vagrant destroy; vagrant up` locally with the required fidelity.  I will also trigger the automated provisioning process (which we know now detects errors after some fixes, woo!) because re-provisioning is absolutely required for tonic-build/prost-build in order for the search-text command to work.

The search-text changes also include changes to how we launch codesearch as discussed at https://bugzilla.mozilla.org/show_bug.cgi?id=1762992#c2 through https://bugzilla.mozilla.org/show_bug.cgi?id=1762992#c4 which was the actual plan arrived at.  This should not regress router.py behavior but immediately following successful provisioning I will kick off a config1 run against the "dev" channel to sanity-check that I didn't break things ahead of the UTC22 run tonight.

PR #506, opened 2022-05-02
The indexer now needs livegrep gRPC built too.

asutherland:
Now that our indexer checks run with the livegrep server running as
started by codesearch.py, we need to replicate the steps to build the
python livegrep gRPC bindings.  This was missed in local testing
because the VM runs both provisioner scripts.

This is a fixup to https://github.com/mozsearch/mozsearch/pull/505

### DIFF
--- a/infrastructure/indexer-provision.sh
+++ b/infrastructure/indexer-provision.sh
@@ -127,6 +127,18 @@ if [ ! -d livegrep ]; then
   rm -rf .cache/bazel
 fi
 
+# Install gRPC python libs and generate the python modules to communicate with the codesearch server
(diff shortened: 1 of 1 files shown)


asutherland:
This ran into the usual reciprocal edge case where the VM running both scripts means that I needed to have the logic be idempotent in the scripts for VM purposes either by nuking directories, suppressing errors about existing directories, or by not doing things if the directory already worked.  I'm addressing this and testing this locally by just folding the gRPC python building inside the "only do this if the livegrep directory doesn't already exist" logic.  This is running now and I'll push the change up and land it once I validate it.

PR #507, opened 2022-05-03
Make livegrep grpc build idempotent for VM.

asutherland:
Since now both provisioning scripts will try to install the python
gRPC, it needs to live in the directory-guarded check so we don't
try and do it twice.  This has been tested locally and ideally won't
break the server provisioning!


asutherland:
xref correcting the server-specific correction on https://github.com/mozsearch/mozsearch/pull/506, comment there

PR #508, opened 2022-05-07
Better default link styling.

emilio:
This makes links behave appropriately by default, and overrides where
needed, rather than what we were doing which is a bit footgunny.


asutherland (approved):
Thanks!

PR #509, opened 2022-05-13
Explicitly install protoc to decrease prost-build failures.

asutherland:
We've been experiencing intermittent "update" cargo build failures
which seem due to prost-build wanting to build `protoc` because we
don't already have it installed and something non-deterministic is
happening there.

It is recommended to already have `protoc` available and it seems
like that this does bypass the build process for protobuf, which will
ideally sidestep the build failures and will definitely improve build
time.  I'm not 100% clear on whether this means we can drop the cmake
dep, but we'll keep it for now.


asutherland:
Local VM re-provisioning and testing passed, plus automated EC2 AMI provisioning succeeded, so merging this and we'll see how it goes.

PR #511, opened 2022-05-18
Bug 1769936 - Add Copy as Markdown shortcuts.

arai-a:
This adds "Copy as Markdown" section and 3 items, per [bug 1769936](https://bugzilla.mozilla.org/show_bug.cgi?id=1769936) comments,
between "Revision control" section and "Other Tools" section.

markdown

Here's the example text for each:

"Filename Link"
```
[Promise.cpp](https://searchfox.org/mozilla-central/source/js/src/builtin/Promise.cpp#443,448,451-453)
```

"Symbol Link"
```
[PromiseDebugInfo::id](https://searchfox.org/mozilla-central/source/js/src/builtin/Promise.cpp#443,448,451-453)
```

"Code Block"
````
https://searchfox.org/mozilla-central/source/js/src/builtin/Promise.cpp#443,448,451-453
```cpp
  static uint64_t id(PromiseObject* promise) {
...
    } else if (idVal.isObject()) {
...
      if (idVal.isUndefined()) {
        idVal.setDouble(++gIDGenerator);
        debugInfo->setFixedSlot(Slot_Id, idVal);
```
````

"Filename Link" and "Code Block" are enabled when there's selected line(s).
"Symbol Link" is enabled when `DocumentTitler` detected title, but I haven't tested this locally given the information isn't available.

### DIFF
--- a/static/css/icons.css
+++ b/static/css/icons.css
@@ -51,6 +51,11 @@
   vertical-align: middle;
 }
 
+.panel .copy.indicator {
+  display: inline-block;
+  pointer: unset;
+}
+
 .panel .copy:hover {
   background-color: var(--button-hover-background);
 }
--- a/static/css/icons.css
+++ b/static/css/icons.css
@@ -53,7 +53,7 @@
 
 .panel .copy.indicator {
   display: inline-block;
-  pointer: unset;
+  cursor: unset;
 }
 
 .panel .copy:hover {
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -173,7 +173,7 @@ a:hover {
 }
 
 /* Some links look like list items */
-.panel section a,
+.panel section .item,
 .context-menu a,
 .folder-content a {
   text-decoration: none;
@@ -181,7 +181,7 @@ a:hover {
 }
 
 /* .folder-content is handled with the whole row getting hover styles, so it doesn't need this. */
-.panel section a:is(:hover, :focus),
+.panel section .item:is(:hover, :focus),
 .context-menu a:is(:hover, :focus) {
   background-color: var(--list-hover-background);
   color: var(--list-hover-color);
@@ -619,9 +619,14 @@ tr.after-context-line + tr.before-context-line {
 .panel a.field {
   background-position: 2px 10px;
 }
-.panel section a {
+.panel section .item {
   display: inline-block;
   width: 100%;
+  cursor: pointer;
+}
+.panel section .item.disabled {
+  opacity: 0.5;
+  cursor: not-allowed;
 }
 
 /* Info Boxes */
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -630,7 +630,7 @@ tr.after-context-line + tr.before-context-line {
   border: unset;
   text-align: unset;
 }
-.panel section button.item[disabled] {
+.panel section .item:disabled {
   opacity: 0.5;
   cursor: not-allowed;
 }
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -624,7 +624,13 @@ tr.after-context-line + tr.before-context-line {
   width: 100%;
   cursor: pointer;
 }
-.panel section .item.disabled {
+.panel section button.item {
+  font: unset;
+  appearance: unset;
+  border: unset;
+  text-align: unset;
+}
+.panel section button.item[disabled] {
   opacity: 0.5;
   cursor: not-allowed;
 }
--- a/static/js/code-highlighter.js
+++ b/static/js/code-highlighter.js
@@ -18,6 +18,7 @@ var DocumentTitler = new (class DocumentTitler {
 
     this.stickyTitle = null;
     this.selectionTitle = null;
+    Panel.onSelectionTitleChanged(this.selectionTitle);
   }
 
   updateTitle() {
@@ -229,6 +230,7 @@ var DocumentTitler = new (class DocumentTitler {
       this.selectionTitle =
         this._findBestPrettySymbolInSourceLineElem(sourceLine);
     }
+    Panel.onSelectionTitleChanged(this.selectionTitle);
 
     this.updateTitle();
   }
@@ -406,6 +408,7 @@ var Highlight = new (class Highlight {
     }
     this.lastSelectedLine = null;
     this.selectedLines = new Set();
+    Panel.onSelectedLineChanged(this.selectedLines);
     this.updateFromHash();
     window.addEventListener("hashchange", () => {
       this.updateFromHash();
@@ -417,11 +420,13 @@ var Highlight = new (class Highlight {
     // NOTE: The order here is intentional so that we throw above if the line
     // is not in the document.
     this.selectedLines.add(line);
+    Panel.onSelectedLineChanged(this.selectedLines);
     this.lastSelectedLine = line;
   }
 
   removeSelectedLine(line) {
     this.selectedLines.delete(line);
+    Panel.onSelectedLineChanged(this.selectedLines);
     if (this.lastSelectedLine == line) {
       this.lastSelectedLine = null;
     }
--- a/static/js/code-highlighter.js
+++ b/static/js/code-highlighter.js
@@ -18,7 +18,7 @@ var DocumentTitler = new (class DocumentTitler {
 
     this.stickyTitle = null;
     this.selectionTitle = null;
-    Panel.onSelectedSymbolChanged(null);
+    this.selectedSymbol = null;
   }
 
   updateTitle() {
@@ -222,6 +222,7 @@ var DocumentTitler = new (class DocumentTitler {
    */
   processLineSelection(lastSelectedLine) {
     this.selectionTitle = null;
+    this.selectedSymbol = null;
     if (lastSelectedLine) {
       const selectedLine = document.getElementById(`line-${lastSelectedLine}`);
       const nestingContainer = selectedLine?.closest(".nesting-container");
@@ -229,12 +230,12 @@ var DocumentTitler = new (class DocumentTitler {
       const sourceLine = nestingLine?.querySelector(".source-line");
       const bestPretty = this._findBestPrettySymbolInSourceLineElem(sourceLine);
       this.selectionTitle = bestPretty.short;
-      Panel.onSelectedSymbolChanged(bestPretty.long);
-    } else {
-      Panel.onSelectedSymbolChanged(null);
+      this.selectedSymbol = bestPretty.long;
     }
 
     this.updateTitle();
+
+    Panel.onSelectedSymbolChanged();
   }
 })();
 
@@ -410,7 +411,6 @@ var Highlight = new (class Highlight {
     }
     this.lastSelectedLine = null;
     this.selectedLines = new Set();
-    Panel.onSelectedLineChanged(this.selectedLines);
     this.updateFromHash();
     window.addEventListener("hashchange", () => {
       this.updateFromHash();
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

**Comment by asutherland** [comment]
This is super duper awesome!!  I'm going to trigger an indexer run so this should show up at https://dev.searchfox.org/ for mozilla-central and the other [config1.json](https://github.com/mozsearch/mozsearch-mozilla/blob/master/config1.json) trees in about 3 hours, unless it fails... in which case the automation I think will email you.

> "Symbol Link" is enabled when `DocumentTitler` detected title, but I haven't tested this locally given the information isn't available.

Assuming you did `make build-test-repo` in the VM from the `/vagrant` directory, then you should have symbolic information available.  For example after checking out your branch locally, with the URL http://localhost:16995/tests/source/big_cpp.cpp#208-218 I was able to use the "symbol link" from your patch to get [outerNS::Couch](http://localhost:16995/tests/source/big_cpp.cpp#208-218) (raw: `[outerNS::Couch](http://localhost:16995/tests/source/big_cpp.cpp#208-218)`).

**Comment by asutherland** [comment]
The indexing run completed and is up at https://dev.searchfox.org/ for testing.

**Comment by arai-a** [comment]
> Assuming you did `make build-test-repo` in the VM from the `/vagrant` directory, then you should have symbolic information available.

Thank you for pointing that out :)
I was using `build-searchfox-repo`.  and yes, it works on `build-test-repo`.

Then, I realized that `DocumentTitler` shortens the namespace part.
That works nicely for document title, but it's not much necessary for symbol link.
So, I'll modify it to notify with not-shortened name.

File: tools/src/output.rs [hunk 1]
```diff
@@ -372,11 +372,21 @@ pub fn generate_panel(
                     } else {
                         String::new()
                     };
+                    let tag = if !item.link.is_empty() {
+                        "a"
+                    } else {
+                        "span"
```
emilio, line 378: Seems like it should probably be a button rather than a span? I guess no big deal if it makes styling too annoying but..

static/js/panel.js
```diff
@@ -126,4 +181,131 @@ var Panel = new (class Panel {
...
+  copyMarkdown(type) {
+    const { node, getText } = this.markdown[type];
+    if (node.classList.contains("disabled")) {
+      return;
+    }
+
+    const copy = node.querySelector(".copy");
+    const url = this.permalinkNode ? this.permalinkNode.href : document.location.href;
+    const filename = new URL(url).pathname.match(/\/([^\/]+)$/)[1];
+    const text = getText(url, filename);;
```
emilio, line 210: We have `select_formatting` in `tools/src/languages.rs`. Seems like that could be reused by adding a `markdown_slug: &'static str` to `LanguageSpec`, and then adding the markdown-spec to the `<div id="file">` in `format_code` (maybe as a data-attribute or something), wdyt?

static/js/panel.js
```diff
@@ -56,6 +85,20 @@ var Panel = new (class Panel {
       });
     }
 
+    for (const [name, { node }] of Object.entries(this.markdown)) {
+      if (node) {
```
emilio, line 89: nit: probably `if (!node) { continue; }` then deindent?

static/js/code-highlighter.js
```diff
@@ -417,11 +422,13 @@ var Highlight = new (class Highlight {
     // NOTE: The order here is intentional so that we throw above if the line
     // is not in the document.
     this.selectedLines.add(line);
+    Panel.onSelectedLineChanged(this.selectedLines);
```
emilio, line 425: (In the future we can do something more generic I guess, but this is ok for now)

It'd be great to put this at the end of the method (once we've updated `lastSelectedLine` here and below), to ensure nothing funny happens in between. Not that it should but...

static/js/panel.js
```diff
@@ -126,4 +181,131 @@ var Panel = new (class Panel {
...
+      } else {
+        node.classList.add("disabled");
+        node.setAttribute("aria-disabled", "true");
+        node.querySelector(".copy").disabled = true;
+      }
+    }
+  }
+
+  onSelectedLineChanged(selectedLines) {
+    this.selectedLines = [...selectedLines].sort((a, b) => a - b);
```
emilio, line 303: It'd be great to avoid this if not copying, is there any reason you need `this.selectedLines` rather than just poking at `Highlight.selectedLines` when needed?

static/js/panel.js
```diff
@@ -110,6 +153,18 @@ var Panel = new (class Panel {
...
+          return this.markdown.filename.node;
+          break;
+        case "s":
+        case "S":
+          return this.markdown.symbol.node;
+          break;
+        case "c":
+        case "C":
+          return this.markdown.block.node;
+          break;
```
emilio, line 167: nit: no break after return

emilio (changes requested):
Yeah, thanks. I think my biggest concern is having duplicated the whole "extension to language" information. Instead, we should be able to reuse what we have without too much effort, wdyt?

static/js/panel.js
emilio: Can't you use `self.Highlight && Highlight.selectedLines.length > 0`?

arai-a: if `isEnabled` returns false until `Highlight` global variable is set,
then buttons aren't enabled when loading URL with line number selected.

emilio: But same happens now right, `Highlight` is what notifies of the selected lines, maybe something like, instead of notifying from the constructor, just notify after `Highlight` definition?

arai-a: Currently the buttons are enabled when loading URL with line numbers selected, because the necessary information is copied from the `Highlight` instance to `Panel` instance, without requiring `Highlight` global variable.

But yeah, if adding extra method call after `Highlight` global variable definition is acceptable, that works for me.

static/js/panel.js
```diff
@@ -126,4 +172,69 @@ var Panel = new (class Panel {
...
+  formatSelectedLines() {
+    const texts = [];
+    let lastLine = -1;
+    for (const line of [...Highlight.selectedLines].sort((a, b) => a - b)) {
+      if (lastLine !== -1 && lastLine != line - 1) {
+        texts.push("...");
+      }
+
+      const lineElem = document.getElementById(`line-${line}`).querySelector(".source-line");
+      texts.push(lineElem.textContent.replace(/\n/g, ""));
```
emilio, line 214: Maybe just `.trim()` rather than `.replace(..)`?

static/js/panel.js
```diff
@@ -126,4 +172,69 @@ var Panel = new (class Panel {
...
+      lastLine = line;
+    }
+    return texts;
+  }
+
+  updateMarkdownState() {
+    for (const [_, { node, isEnabled }] of Object.entries(this.markdown)) {
+      if (isEnabled()) {
+        node.disabled = false;
+        node.removeAttribute("aria-disabled");
```
emilio, line 225: This could be written:

```
node.disabled = !isEnabled();
node.toggleAttribute("aria-disabled", node.disabled);
```

But this looks good as well.

tools/src/languages.rs
```diff
@@ -566,6 +567,16 @@ lazy_static! {
...
+    };
+
+    static ref HTML_SPEC : LanguageSpec = LanguageSpec {
+        reserved_words: make_reserved(&*RESERVED_WORDS_JS),
+        c_style_comments: true,
+        backtick_strings: true,
+        regexp_literals: true,
+        markdown_slug: "html",
         .. LanguageSpec::default()
     };
```
emilio, line 582: nit: you can do something like this instead:

```diff
diff --git a/tools/src/languages.rs b/tools/src/languages.rs
index 096de84..9d258e5 100644
--- a/tools/src/languages.rs
+++ b/tools/src/languages.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 use std::path::Path;
 
-#[derive(Debug, Default)]
+#[derive(Debug, Clone, Default)]
 pub struct LanguageSpec {
     pub reserved_words: HashMap<String, String>,
     pub hash_comment: bool,
@@ -572,12 +572,8 @@ lazy_static! {
     };
 
     static ref HTML_SPEC : LanguageSpec = LanguageSpec {
-        reserved_words: make_reserved(&*RESERVED_WORDS_JS),
-        c_style_comments: true,
-        backtick_strings: true,
-        regexp_literals: true,
         markdown_slug: "html",
-        .. LanguageSpec::default()
+        .. JS_SPEC.clone()
     };
 
     static ref CPP_SPEC : LanguageSpec = LanguageSpec {
```

static/js/panel.js
```diff
@@ -126,4 +172,69 @@ var Panel = new (class Panel {
...
+  }
+
+  copyMarkdown(type) {
+    const { node, getText } = this.markdown[type];
+    if (node.disabled) {
+      return;
+    }
+
+    const copy = node.querySelector(".copy");
+    const url = this.permalinkNode ? this.permalinkNode.href : document.location.href;
```
emilio, line 199: nit: `this.permalinkNode?.href || document.location.href` if you want.

static/js/panel.js
```diff
@@ -6,9 +6,42 @@ var Panel = new (class Panel {
...
-    this.rawNode = this.findLink("Raw");
+    this.permalinkNode = this.findItem("Permalink");
+    this.logNode = this.findItem("Log");
+    this.rawNode = this.findItem("Raw");
+
+    this.markdown = {
+      "filename": {
+        node: this.findItem("Filename Link"),
+        isEnabled: () => {
+          return Highlight?.selectedLines.size > 0;
```
emilio, line 17: nit: This one can probably be unconditionally enabled? I could see it being useful without any selected lines.

static/css/icons.css
```diff
@@ -51,6 +51,11 @@
   vertical-align: middle;
 }
 
+.panel .copy.indicator {
+  display: inline-block;
+  pointer: unset;
```
emilio, line 56: You probably mean `cursor: unset` right?

static/css/mozsearch.css
```diff
@@ -619,9 +619,20 @@ tr.after-context-line + tr.before-context-line {
...
   width: 100%;
+  cursor: pointer;
+}
+.panel section button.item {
+  font: unset;
+  appearance: unset;
+  border: unset;
+  text-align: unset;
+}
+.panel section button.item[disabled] {
```
emilio, line 633: nit: you can use `.panel section .item:disabled` probably, it's a bit more to the point.

emilio (approved):
Thanks! Looks good to me with the following nits.

static/js/panel.js
arai-a: `trim` removes WhiteSpace and LineTerminator, so, that doesn't preserve the leading and trailing whitespace.
https://tc39.es/ecma262/#sec-trimstring

It would be a rare case that there's trailing whitespace, so `trimEnd` might work, but I think `replace` for LineTerminator (that's added [here](https://github.com/mozsearch/mozsearch/blob/e464d4af245a3dbbacef87628ba92ef1f3cb3468/tools/src/format.rs#L568)) is safer.
but I realized `g` flag shouldn't be necessary.

static/js/panel.js
arai-a: The ARIA spec seem to expect the value being "true", "false", or unset
https://w3c.github.io/aria/#enumeration-example

`toggleAttribute` sets it to empty string, so I guess `setAttribute`+`removeAttribute` is safer.

asutherland:
Woo!  I know this is going to make a huge quality-of-life difference in working for searchfox for a lot of people!  Thanks so much, @arai-a!

One extra thing you might want to consider doing is sending a mail to `dev-platform@mozilla.org` to let people know about the change.  It looks like we can expect the first run with this change to go active at about ~18:10 UTC.  I've also re-triggered dev.searchfox.org against trunk and that should become active between 2h10m and 3h from now.

PR #512, opened 2022-05-19
Ignore tests/tests/checks/explanations/ directory generated by test

arai-a:
This directory is created by test
https://github.com/mozsearch/mozsearch/blob/117e70d95008c48f83f7c16eb856488135265baf/tools/tests/test_check_insta.rs#L241-L242

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ repos
 /index
 /config
 addons
+/tests/tests/checks/explanations


asutherland:
Thanks and sorry about that.  The intent is to eventually commit these when they become useful; I should have added the ignore line when I decided they weren't ready to be committed.

PR #513, opened 2022-05-20
Bug 1770501 - Show Copy as Markdown also on not-latest page.

arai-a:
Given the panel for non-latest page is generated in `format_path` in `tools/src/format.rs`, not in `main` in `tools/src/bin/output-file.rs`, the "Copy as Markdown" panel section needs to be added to both of them.
I've factored out the PanelSection creation into `create_markdown_panel_section` and let both of them call it.

### DIFF
--- a/static/js/panel.js
+++ b/static/js/panel.js
@@ -191,7 +191,7 @@ var Panel = new (class Panel {
 
   copyMarkdown(type) {
     const { node, getText } = this.markdown[type];
-    if (node.disabled) {
+    if (!node || node.disabled) {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


asutherland:
Currently, I don't think "Symbol Link" functionality can ever work for not-latest pages.

arai-a:
Given the panel content is already different for not-latest page, hiding it makes sense.
will update patch shortly.

asutherland:
Thanks!  This seems straightforward and sounds like it's been locally tested, so I'm going to merge and re-trigger config1.

PR #517, opened 2022-05-26
js: Re-run prettier on static/js.

emilio:


emilio:
(Merging as it's trivial and should make other JS patches a bit easier to review)

PR #519, opened 2022-05-27
Operating core "query" implementation with pipeline-server.rs exposure, no UI beyond JSON

asutherland:
This stack:
- Provides the MVP classic searchfox search via the new "query" mechanism which translates query-parser style syntax into a pipeline graph consisting of 3 parallel searches (file search, semantic search via crossref lookup and results expansion via crossref-expand, fulltext search via livegrep/codesearch), a junction in "compile-results" that takes any of those, plus lookup of lines from the generated HTML files with user-configurable context including overlap compensation.
  - Parallelism is leveraged by spawning a tokio task for each parallel pipeline.  Currently all our memory map lookups are treated as non-blocking but that may need to be revisited.
- Exposes "query" internally for the searchfox-tool for testing.
- Makes the pipeline-server a thing that seems to work.  If you run the VM and add a keyword bookmark for `http://localhost:16995/tests/query/default?q=%s` you can issue queries and get JSON back that looks convincingly like reasonable search results, although probably with some bugs here and there.
  - Specifically I've noticed a "pretty" symbol that was still mangled, but that may be a pre-existing data bug that router.py may side-step because it likes to overzealously demangle everything whereas I've tried to leverage that we now should have structured "meta" info for most C++ types.
  - Note that I'd previously talked about hardening the abstract servers against misuse, but the query mechanism doesn't actually provide a way for web requests to manipulate the pipeline in a way to generate paths; only filter them.  And there's no way to arbitrarily use other pipeline commands or provide them with arbitrary arguments.  Everything is just fill-in-the-blanks.

This landing won't actually provide any useful functionality to anyone yet, but it's pretty close.  The key next step is to bind the search results from the above into a UI along the lines of the existing "search" UI.  The only deviation right now is the introduction of automatic faceting of qual-kind group results so that we can potentially provide some checkboxes that allow turning off sets of related symbols for overrides, or turning off groups of paths.  This data is provided as a sort-of sidecar that can also just be ignored.

My tentative plan is to use the liquid templating support I've already added for the test "explanation" mechanism (which is still a WIP to be useful) so that we can translate the results directly into semantic HTML rather than continuing to use our ad hoc JS HTML templating in `populateResults` in `search.js` that we then poke into innerHTML.  That said, note that I explicitly chose liquid because there's also a JS version of it too, so we could also potentially use the same templating logic on both the client and server if desired, etc. etc.

### DIFF
--- a/docs/searchfox-tool-cookbook.md
+++ b/docs/searchfox-tool-cookbook.md
@@ -49,6 +49,13 @@ searchfox-tool 'query --help'
 --tree=mozilla-central search-identifiers mozilla::GetContentWin32kLockdownEnabled | crossref-lookup | traverse --edge=uses --max-depth=9 | graph --format=svg'
 ```
 
+```
+~/mozsearch/tools/target/release/searchfox-tool '--server=/home/ubuntu/index/config.json
+--tree=mozilla-central search-identifiers GetLiveWin32kLockdownState | crossref-lookup | traverse --edge=uses --max-depth=9 | graph --format=svg'
+```
+
+
+
 ### Diffing Query Results
 
 While investigating aspects of queries that hit limits because of non-intuitive
--- a/infrastructure/web-server-run.sh
+++ b/infrastructure/web-server-run.sh
@@ -22,6 +22,7 @@ STATUS_FILE="${SERVER_ROOT}/docroot/status.txt"
 pkill codesearch || true
 pkill -f router/router.py || true
 pkill -f tools/target/release/web-server || true
+pkill -f tools/target/release/pipeline-server || true
 
 sleep 1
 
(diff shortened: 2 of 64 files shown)


asutherland:
I'm going to land this now and trigger a re-provision (for memmap2, made mainly because ripgrep uses that because memmap is abandoned) to get some stability baking time on this over the less-busy weekend time-range with US holiday attached and because I'm also planning to iterate rapidly on getting the pipeline server to do useful things over the weekend.

asutherland:
Uh, and just for clarity, the branch name of "graphviz-serve" ended up being a misnomer.  Right now `query_core.toml` that defines the query mappings doesn't know how to translate queries into a graph pipeline, but that can be easily added and would result in a JSON representation of the underlying PipelineValues representation of whatever output rep resulted (JSON, dot, SVG, etc.)

asutherland:
Well, I hit merge like 2 minutes before the new web-server for today forked, so one can already visit https://searchfox.org/mozilla-central/query/default?q=workerprivate and behold the JSON results!  What?  You say you want 4 lines of context on the semantic results too and you demand the results still be provided in an unusable JSON format?  Behold https://searchfox.org/mozilla-central/query/default?q=C:4%20workerprivate !

PR #522, opened 2022-05-31
Bug 1771805 - Upgrade to 22.04 from 20.04.

asutherland:
Likely appropriate steps for people to do locally:
- On linux, make sure you have a recent version of `vagrant-libvirt`.  Things
  work for me with `vagrant-libvirt` 0.7.0 on Ubuntu 22.04 as a host.
- Run `vagrant destroy`
- Run `vagrant up --no-destroy-on-error`.  If you experience any errors,
  check the logs which you can access via `vagrant ssh`.
- It will likely be necessary to manually `make clean` in `clang-plugin`
  before running `make build-test-repo` in `/vagrant` in the VM.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,6 +1,6 @@
 Vagrant.configure("2") do |config|
-  config.vm.box = "generic/ubuntu2004"
-  config.vm.box_version = "3.2.6"
+  config.vm.box = "generic/ubuntu2204"
+  config.vm.box_version = "4.0.0"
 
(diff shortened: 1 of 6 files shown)


asutherland:
This works locally in the VM and it's necessary to land this to test the online provisioning process.

Right now there's a config1 indexer in-flight, which means it will be triggering a web-server in about 43 minutes and so I'm going to trigger indexer provisioning first and verify that before triggering the web-server provisioning.  I'll then be doing some "dev" channel pushes which should help validate things.

PR #527, opened 2022-06-07
panel: Fix deindent with leading line with no whitespace.

arai-a:
This fixes the case that the first line has no whitespace.
the current code ignores that line and always deindent removing the first line's characters


arai-a:
@emilio, can you review?

emilio (approved):

PR #528, opened 2022-06-11
Bug 1773767 - Update the IPDL parser and add limited test coverage.

asutherland:
Because there's a lot of moving parts for the mozilla-central IPC
implementation, we just copy analysis files from a web-server that
has the C++ analysis data we want for the source files we want to show.

This provides ipdl-analyze.rs with the analysis files it needs to
scrape to establish the C++ symbol mapping.  And this is then enough
for the test repository to give us a full fidelity experience of
viewing the .ipdl file and sufficient fidelity when clicking on those
associated Send/Recv methods as they relate to the IPDL file.  That's
the end of our fidelity, although if we pulled down other source files
and their associated analysis files, we could get a little more.

The goal is to allow rapid iteration on the IPDL mapping and analysis
record generation without having to go through a full mozilla-central
lookup.

TODO: Still need to add the test coverage here, but I want a try run.


asutherland:
@amccreight There isn't a ton going on here that's all that interesting and I probably should split the stuff copied from mozilla-central into its own commit for clarity, but if you're interested in looking at this, please feel free!  Those files live under mc-generated and mc-analysis in general, though.

I've started a try run of this on "dev" so we can see if anything breaks and to maybe see how the UX works.  There are going to be some limitations on the "jumps" "Go to definition" because we treat the IDL record as definition record, but that will be fixable when moving from ANALYSIS_DATA to storing symbol metadata in the outputted files instead.

asutherland:
With the most up-to-date changes from @amccreight's IPDL parser updates we got a successful run against mozilla-central, so I've landed this now so that we can get a few runs in to avoid breakage on Monday.  I did update the "source" record generation after the dev run (and verified locally) so that we provide basically 3 search options when viewing the IPDL sources:
- Just the C++ send symbol
- Just the C++ recv symbol
- A combination of the synthetic IPDL symbol (which only ever occurs in the IPDL source and exists to have a logically distinct symbol to hold the structured analysis information independent of the C++ symbols), the C++ send symbol, and the C++ recv symbol.

Right now that's also the order in which things are presented.  It probably would be better if the unified symbol came first, but I personally probably won't try and address that until we move from using ANALYSIS_DATA in the medium term, as at that point we can have the context menu code be able to specialize things with a full understanding of the symbols at play and can have "go to def" variations that work which are currently impacted by the limitations of the jumps heuristics.

PR #529, opened 2022-06-20
Improved build times, land "query" graphing support code

asutherland:
See specific commits, but this generally is a landing of the diagram functionality that's been on the "dev" server for over a week and that started producing useful stuff on Wednesday.  This also includes `make build-test-repo` iteration speed improvements.

The main notable thing here that will want review attention in the future is that the non-diagram "query" support that is intended to be an enhanced version of the classic "search" results does include changes to the icon CSS rules so that we don't need to use a look-up table to map languages to icons and instead can rely on the CSS rules to do the matching with fallback for us.  I believe the added rules don't break existing behavior, and I'll endeavor to draw attention to these rules as we now start to get to the point where there will be UX related behavior that wants feedback and review!  (The things landing are still mainly MVP/proof-of-concept.)


asutherland:
Note that all provisioning succeeded (both AWS types and local VM re-provisioning), but it seems like rust has added some kind of new lifetime checks that are somehow non-deterministic; I think I made them happy, but if they are in fact non-deterministic, there's a chance there will be breakage.

PR #543, opened 2022-07-17
Bug 1779939 - pin git-cinnabar at 0.5.8 for now

asutherland:
This is a breakage mitigation and we should un-pin when feasible.


asutherland:
Re-provisioning with this change succeeded and the config1 index triggered using this image is now running okay.

PR #545, opened 2022-07-18
Bug 1779899 - pre-cache all m-c indices on release1

asutherland:
I will immediately provision both of these.


asutherland:
https://github.com/mozsearch/mozsearch/pull/546 landed as a fixup to this where the format string was passing "channel" twice after my change (despite channel not be used).  Now that we're solidly on python3 we can generally just move to `f'blah'` formatted strings that resemble JS template literals to avoid all failure modes related to this.

PR #568, opened 2022-09-12
Process mjs files as modules

gijsk:
Use the .mjs extension to opt in to module parsing.


asutherland (approved):
Ideally we can add some "checks" for JS in the future, but by inspection this looks pretty good and it's a pretty big regression to not have this and I don't know of anyone else looking to contribute in this area, so let's land this for now and I'll make sure we have an appropriate follow-up bug or something.

aminomancer:
Thanks!

PR #569, opened 2022-09-13
Revert "Process mjs files as modules"

asutherland:
Reverts mozsearch/mozsearch#568

js-analzye was not prepared for the additional new AST nodes it found in there:
```
/home/ubuntu/mozsearch/scripts/js-analyze.js:326:12 uncaught exception: Unexpected statement: ExportDeclaration
```

### DIFF
--- a/scripts/js-analyze.js
+++ b/scripts/js-analyze.js
@@ -292,8 +292,7 @@ let Analyzer = {
   parse(text, filename, line) {
     let ast;
     try {
-      let target = filename.endsWith(".mjs") ? "module" : "script";
-      ast = Reflect.parse(text, {loc: true, source: filename, line, target});
+      ast = Reflect.parse(text, {loc: true, source: filename, line});
 
       let parsedLines = text.split('\n');
 


emilio:
cc @gijsk

Probably the nodes to handle are these: https://searchfox.org/mozilla-central/rev/8986d2d483c472e27b78acb0106215d7c523e7fa/js/src/jsast.tbl#70-78

gijsk:
@emilio yes, but I don't have an easy way of setting up searchfox locally and testing any changes. Someone else on the front-end team tried and ran into:

- vagrant failing to configure the VM
- the build somehow requiring building a custom jemalloc for one of the rust packages, which then requires symlinks to work on the host filesystem, which they don't on Windows because... well, symlinks + Windows
- followed by: `FATAL: bazel crashed due to an internal error.`

and then they gave up (for which I can't exactly blame them), and I am now also kind of loathe to try on Windows (and don't really want to clutter up my mac with virtualbox, vagrant, and everything else, when I do all my other virtualization on Windows).

It would be useful if for the JS analyzer bits and the web bits (stylesheets / front-endy bits), there was a way of hacking on them without requiring building all the other stuff, a VM, etc. Like artifact builds for Firefox, or something.

asutherland:
> @emilio yes, but I don't have an easy way of setting up searchfox locally and testing any changes. Someone else on the front-end team tried and ran into:
> 
>     * vagrant failing to configure the VM

@gijsk Is there any chance you can get us the output of how vagrant failed to configure things?  I try and actively keep an eye on https://chat.mozilla.org/#/room/#searchfox:mozilla.org and filed searchfox bugs, but I missed it if the difficulty was reported by either of these channels.  I'm also very happy to respond to private messages on matrix (or small group chats) if that makes anyone more comfortable!  (I know if I were starting out on searchfox and/or was new at Mozilla, I'd probably prefer it myself.)

That said, it sounds like the Windows use case is common enough and vagrant headaches as well that we should probably try and find an easier on ramp for people, so maybe it's better to just start from there.  I think the options are:

1. Docker (which bhearsum recommended)
   - For people who don't have a VM solution on their machine already, "Docker Desktop" is available to order from the Service Desk "Order Stuff" portal at https://mozilla-hub.atlassian.net/servicedesk/customer/portal/6 (and which hates direct links).
   - For people who already have a VM solution on their machine so they can run linux for pernosco, docker should be able to work within the VM I think?  This ideally would avoid complications related to credentials because the VM that's running docker would probably already be set up for that.
2. VMware: Try and provide a 1-click-ish mechanism for people to get a searchfox VM that they can directly launch from "VMware Workstation Pro" as available to order from the Service Desk "Order Stuff" portal at https://mozilla-hub.atlassian.net/servicedesk/customer/portal/6
3. WSL2: Try and provide minimal setup steps for using WSL2.  https://learn.microsoft.com/en-us/windows/wsl/install documents installing WSL which installs Ubuntu by default.  https://learn.microsoft.com/en-us/windows/wsl/use-custom-distro#importing-centos-example documents importing a WSL OS from a tarball, AIUI.

My inclination would be to start with docker and go from there since it could be beneficial for the existing contributor-base as well which also helps ensure it doesn't bit-rot.

>     * the build somehow requiring building a custom jemalloc for one of the rust packages, which then requires symlinks to work on the host filesystem, which they don't on Windows because... well, symlinks + Windows
> 
>     * followed by: `FATAL: bazel crashed due to an internal error.`

Unfortunately it's not surprising for this to happen if trying to build/install outside of the container/VM, especially on Windows if outside WSL(2).  I definitely understand trying this after being confronted by opaque vagrant errors, but yeah, it definitely will be a painful process that ends in failure even if we removed the jemalloc and lld deps.  (Although it could make sense to remove them if necessary for use of rust-analyzer by VS code running outside a VM/container.  lld makes a huge incremental build time difference right now is why we're using it but has annoying complications.)
 
> It would be useful if for the JS analyzer bits and the web bits (stylesheets / front-endy bits), there was a way of hacking on them without requiring building all the other stuff, a VM, etc. Like artifact builds for Firefox, or something.

I think the equivalent for this would be CI/try runs, which we do have, but are generally very un-ergonomic / unpleasant to use compared to a local VM.  Something like https://github.com/features/codespaces could be an option but it's not clear there's management support for the level of additional costs that would require unless that's already something all teams have access to for firefox dev purposes.

asutherland:
I've landed initial docker support in https://github.com/mozsearch/mozsearch/pull/571 that works on my Ubuntu 22.04 box with the "docker.io" package installed.  I'll try running that under WSL2 Ubuntu on Windows tomorrow to see how it goes and try and overhaul the docs for how people can use it.

Right now the usage is exactly like the vagrant usage where we assume you have some outer posix-y dev environment which is capable of running the docker CLI, and then the docker image is just a sandbox, but you do all your development outside the docker image including git stuff and editing stuff.  I think this may rule out the "Docker Desktop" approach for now.

asutherland:
> > vagrant failing to configure the VM
> 
> The cause of this was that I was using VMWare as the Vagrant provider and the vagrantfile here only has support for Virtualbox and libvirt. I copied the VBox block into a VMWare one and then it seemed to work. It would have been nicer to find a way to write the config in a more general way, instead of needing blocks for every possible provider, but I couldn't see a way that the syntax would allow me to do that.

Yeah, unfortunately the bind mechanism does seem to need to be per-provider.  If you could provide a pull request to add the VMware provider support you have locally, that would be great just in case anyone tries to use this in the future, although I think it makes sense to try and move the vagrant stuff to like "we don't recommend it, but it's an option!"

> > the build somehow requiring building a custom jemalloc for one of the rust packages, which then requires symlinks to work on the host filesystem, which they don't on Windows because... well, symlinks + Windows
> 
> Pretty much, yeah. Some Rust crate somewhere had a dependency on jemalloc-sys, which in turn ran the C build for jemalloc, which then tried to make a symlink inside of the VM's shared folder (it seemed like this was at the end of that build, so I'm guessing it was the usual `libname.so` -> `libname.so.x.y.z` link), and that command predictably failed on an NTFS host. I don't know if there's any good way around this one other than either to tell people to enable symlinks on their Windows machines (which, I mean, there's a reason that setting is off by default and also not exposed in any GUI), or otherwise to somehow have that phase of build run in some temporary location within the guest filesystem, but this command is so many layers deep I'm not sure if that's possible.

Aha!  Okay, I misinterpreted what Gijs had said to be that you'd hit problems with vagrant and then had tried building/running things outside of the VM entirely.  (Most of the problems I've faced with vagrant have been in the ruby glue logic freaking out such that the VM never even gets to our provisioning scripts or in getting the bind working, so that's what I assumed happened, which left only trying to directly do stuff outside the VM.)

I agree no one developing on windows should need to enable symlinks for their host filesystem.  Along these lines, I think it should also be possible to use VS code natively in windows with Rust-Analyzer with full fidelity rather than having to try and run VS code from inside a VM or WSL2 or something convoluted like that.  So I think this probably precludes any cleverness about doing parts of the build inside the VM FS, which is good because I feel like that would be a major source of confusion for at least me.

So I'll tentatively plan to remove the jemalloc dependency and the lld build optimization for now.  I added jemalloc for the memory introspection in https://bugzilla.mozilla.org/show_bug.cgi?id=1464297#c14 and https://github.com/mozsearch/mozsearch/pull/275.  A recent set of numbers is:

```
Config: 174526816 bytes                                                                                                                                                                                           
  Tree: comm-central = 4216000 bytes                                                                                                                                                                              
    blame_map = 4215632 bytes                                                                                                                                                                                     
    hg_map    = 0 bytes                                                                                                                                                                                           
  Tree: mozilla-central = 167630964 bytes                                                                                                                                                                         
    blame_map = 76570192 bytes                                                                                                                                                                                    
    hg_map    = 90939372 bytes                                                                                                                                                                                    
  Tree: mozilla-mobile = 240 bytes                                                                                                                                                                                
    blame_map = 0 bytes                                                                                                                                                                                           
    hg_map    = 0 bytes                                                                                                                                                                                           
  Tree: nss = 2677508 bytes                                                                                                                                                                                       
    blame_map = 1148728 bytes                                                                                                                                                                                     
    hg_map    = 1528492 bytes     
```

Which tells us our memory usage for m-c is now at 159.9 MiB, up from 145.9 MiB 31 months ago which is like 0.47 MiB a month which means we don't expect it to become a problem and we can lose the mechanism without much concern.  Especially since I don't think we've checked the numbers between now and then.  If we were to want something similar in the future I think I'd probably just use a memory profiler or do basic wall-clock checks of our VM size before and after populating the given data structure.

> > followed by: `FATAL: bazel crashed due to an internal error.`
> 
> This happened while I was trying to manually resume the provisioning procedure after it had failed midway through due to some web request failing for reasons I didn't understand. I don't have further details on that because, to be honest, I was out of patience at this point, so as soon as I saw the words "bazel" and "internal error" I just closed the terminal. I was not outside of the VM at any point. But yeah, this probably isn't actionable.

If you have any vague recollection of what the web request failure might have been, like if it was apt stuff or the provision scripts wgetting/curling, that would be useful.  Also if you remember how stable the internet connection you were using was?  (I know I've had major trouble provisioning on all-hands wifi (both hotel and conference) where things just fail partway through because of general network reliability issues, but haven't experienced problems outside of that.)

I should say the bazel build process always gives some weird errors partway through about being unable to fetch something from sourceforge but it keeps going, so that probably wasn't what you saw.  If you end up trying any of this out again and see the bazel failure again, I'd definitely appreciate any log excerpts you can provide.

> As far as where to go from here, I think I agree that Docker is probably a fine choice. It _might_ require some setup to get running on Windows, but for anyone that has WSL2, you're already good. I don't think specific integration with VMWare is necessary; Docker can run alongside it just fine. Now personally, I certainly don't _like_ having to use Docker instead of things running "natively" on Windows, but I'm also aware I'm probably the only person in the "open source" world who didn't give up on that decades ago. So I think it's probably the best option here.

Would WSL2 count as running things natively?  In general it does seem desirable for Mozilla projects to not try and dissuade devs from running Windows as their native OS, so I could see trying to support searchfox on WSL2 without involving Docker since it should provide pretty good production fidelity.  (Especially because the docker licensing situation is frankly a bit confusing/stressful.)  But supporting searchfox running directly with like `C:\foo` paths for example seems like a much bigger ask because searchfox will never run in production on windows because it costs ~80% more for indexers and ~44% for web servers.

asutherland:
I've landed https://github.com/mozsearch/mozsearch/pull/572 removing the jemalloc deps which also proved problematic on WSL2 for me last night where I checked out searchfox into WSL2 Ubuntu 22.04.  I was more than a little surprised when VS code was like "hey, I can mount the contents of WSL filesystem easily!" and I pushed the button and then it indeed did that easily.

Some docs on WSL and VS code are (noting that in general MS seems to have done a fantastic job at providing docs for enabling WSL and making sure they rank highly on search engines):
- https://code.visualstudio.com/docs/remote/wsl-tutorial which also covers enabling WSL as necessary.
- https://code.visualstudio.com/docs/remote/wsl

One thing to note re: WSL2 is that the current "Ubuntu" VM version appears to be 20.04 but the version I'd recommend is 22.04 and that's [available in the Microsoft/Windows Store](https://apps.microsoft.com/store/detail/ubuntu-22041-lts/9PN20MSR04DW?hl=en-us&gl=us) but not currently available to `wsl --list --online`, although it presumably will be added soon and perhaps "Ubuntu" will be updated to Ubuntu 22.04 then too.  (Noting that it's always possible to `sudo do-release-upgrade`.)

asutherland:
I landed https://github.com/mozsearch/mozsearch/pull/573 to eliminate the lld dep too.  Going to try and finish getting searchfox building and running on Windows under WSL2 (with searchfox checked out inside the WSL2 VM, which again, I think is explicitly not what @mhowell-mozilla was doing, but might be the most sane thing to recommend to people in order to avoid requiring use of VMware or Docker Desktop).

asutherland:
With those patches landed and some additional dependencies installed into the WSL Ubuntu 22.04 distro, VS Code connected to WSL and its rust-analyzer were able to build everything.  Then, if dockerd is manually started[1] inside that WSL instance (once iptables is reconfigured to use iptables-legacy via `sudo update-alternatives --config iptables`), I was able to use the `./builder-docker.sh` and `./run-docker.sh` scripts successfully, with `make build test-repo` in `/vagrant` passing inside the docker container.

I created a script to help detect when dockerd isn't running and to start it.  It probably also makes sense to translate my notes into a helper shell script that can be run under sudo to help simplify the WSL setup process.

1:
```
sudo dockerd -H unix:///var/run/docker.sock > /dev/null 2>&1 &
disown
```

PR #570, opened 2022-09-19
Rust support for SCIP / rust-analyzer.

emilio:


asutherland:
I've assigned https://bugzilla.mozilla.org/show_bug.cgi?id=1761287 to you and added an attachment link to this PR.

Aside: What's here already is fine for landing already as long as `make build-test-repo` passes in your VM if you want to do smaller landings, etc.

emilio:
@asutherland Ok, so I took another look today, and I found out why I got stuck.

I got stuck because I couldn't figure out how to find the right local reference a definition was pointing to. I think this is a rust analyzer bug tho, because the local symbols are per file rather than global. I sent https://github.com/rust-lang/rust-analyzer/pull/13456. Assuming I'm right and that's accepted, then this should be in a decent enough state.

tools/src/bin/rust-indexer.rs
```diff
@@ -809,3 +824,219 @@ fn main() {
...
+                }
+            }
+        };
+
+        for occurrence in &doc.occurrences {
+            let loc = scip_range_to_searchfox_location(&occurrence.range);
+            let symbol = lookup_symbol(&occurrence.symbol);
+            {
+                let global = sanitize_symbol(&symbol.symbol);
+                let pretty = pretty_symbol(&symbol.symbol);
```
asutherland, line 958: For [source symbols](https://github.com/mozsearch/mozsearch/blob/master/docs/analysis.md#sources) we expect pretty to have a space-delimited human-readable description before the actual pretty symbol name, for example "constructor SYMBOL".

The [existing code does this](https://github.com/mozsearch/mozsearch/blob/b7c0f39354fe3ec74a74d3ad4937147ab29c213c/tools/src/bin/rust-indexer.rs#L325-L341) so it would probably be a regression.  Looking at the rust analysis check outputs ([1](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/snapshots/analysis/rust/simple.rs/check_glob%40simple_Loader_new__json.snap) [2](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/snapshots/analysis/rust/generated.rs/check_glob%40GeneratedType__json.snap) [3](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/snapshots/analysis/rust/test_rust_dependency/check_glob%40MyType__json.snap)) I only see target records.

It turns out the reason we don't have source records in there is that we don't populate the "syntax" field [which is used to determine the "kind" of source records by the `filter-analysis` command](https://github.com/mozsearch/mozsearch/blob/b7c0f39354fe3ec74a74d3ad4937147ab29c213c/tools/src/cmd_pipeline/cmd_filter_analysis.rs#L63-L83).

tools/src/bin/rust-indexer.rs
```diff
@@ -809,3 +824,219 @@ fn main() {
...
+
+                let source_data = WithLocation {
+                    data: AnalysisSource {
+                        source: SourceTag::Source,
+                        syntax: vec![],
```
asutherland, line 963: As per the above it turns out that our "check" infrastructure needs us to populate the syntax field.  [In the C++ indexer we explicitly put the canonical target kind in there](https://searchfox.org/mozilla-central/rev/bfa4dfd40b8fe6f1a6504e86b5ad2c5fd402bfc0/build/clang-plugin/mozsearch-plugin/MozsearchIndexer.cpp#1417-1420).

Note that if fixing this (which I think is probably a good idea), we do have production mozilla-central checks that will be impacted ([ex](https://github.com/mozsearch/mozsearch-mozilla/blob/master/mozilla-central/checks/snapshots/analysis/rust/check_glob%40build__webrender_shader__json.snap)).  The test docs at https://github.com/mozsearch/mozsearch/blob/master/docs/testing-checks.md should provide some useful context on this.

It's probably possible to fix the existing path and update the checks separate from this patch which would then not need to change the checks, but I don't have a strong opinion other than thinking we probably want the improved coverage and it's probably good for you to experience this workflow as it's one of the areas I think I've been basically doing my own thing and it would be great to be able to have others have more experience here and to be able to provide feedback/etc.

Note that the WIP test explanations do exist and capture `tracing` logging, so like right now I have at `REPO/tests/tests/checks/explanations/analysis/rust/simple.rs/simple_Loader_new__json-local.md`, the following file contents:

```json
{
  "level": "TRACE",
  "target": "tools::cmd_pipeline::builder",
  "fields": {
    "cmd": "FilterAnalysis(FilterAnalysis { file: \"simple.rs\", record_type: None, kind: Some(\"def\"), query_opts: SymbolicQueryOpts { symbol: Some(\"simple::<Loader>::new\"), identifier: None } })"
  },
  "timestamp": "2022-10-12T15:49:26.742788Z",
  "spans": [
    {
      "name": "build_pipeline",
      "arg_str": "filter-analysis simple.rs -k def -s simple::<Loader>::new\n"
    }
  ]
}
```


```json
{
  "timestamp": "2022-10-12T15:49:26.743350Z",
  "spans": [
    {
      "name": "run_pipeline_step",
      "cmd": "FilterAnalysisCommand { args: FilterAnalysis { file: \"simple.rs\", record_type: None, kind: Some(\"def\"), query_opts: SymbolicQueryOpts { symbol: Some(\"simple::<Loader>::new\"), identifier: None } } }"
    }
  ],
  "level": "TRACE",
  "fields": {
    "output_json": "{\n  \"JsonRecords\": {\n    \"by_file\": [\n      {\n        \"file\": \"simple.rs\",\n        \"records\": [\n          {\n            \"loc\": \"00037:11-14\",\n            \"target\": 1,\n            \"kind\": \"def\",\n            \"pretty\": \"simple::<Loader>::new\",\n            \"sym\": \"simple::<Loader>::new\"\n          }\n        ]\n      }\n    ]\n  }\n}"
  },
  "target": "tools::cmd_pipeline::interface"
}
```

These contents are `.gitignore`d right now because I want to improve [the template that logs the explanations](https://github.com/mozsearch/mozsearch/blob/master/tools/templates/pipeline_explainer.liquid) and also because the [tracing "cleverness" in test_check_insta](https://github.com/mozsearch/mozsearch/blob/b7c0f39354fe3ec74a74d3ad4937147ab29c213c/tools/tests/test_check_insta.rs#L129-L138) only is able to report for simple pipelines that stay on the testing thread but loses track of things when they go multi-threaded.  https://bugzilla.mozilla.org/show_bug.cgi?id=1763005 generally tracks the goal and remaining issues there.

So if you want more insight into what the checks are doing, that's a place you can find it and if you add more tracing logging, you can find that there too.  For these simple test cases that are basically just a single pipeline step, there isn't much utility there, but like for the HTML extraction it gets more interesting because you can see the intermediate values of the pipeline, for example for the HTML version of the above ([input](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/inputs/analysis/rust/simple.rs/simple_Loader_new__html)):

```json
{
  "spans": [
    {
      "arg_str": "filter-analysis simple.rs -k def -s simple::<Loader>::new | show-html\n",
      "name": "build_pipeline"
    }
  ],
  "target": "tools::cmd_pipeline::builder",
  "fields": {
    "cmd": "FilterAnalysis(FilterAnalysis { file: \"simple.rs\", record_type: None, kind: Some(\"def\"), query_opts: SymbolicQueryOpts { symbol: Some(\"simple::<Loader>::new\"), identifier: None } })"
  },
  "level": "TRACE",
  "timestamp": "2022-10-12T15:49:26.735229Z"
}
```


```json
{
  "target": "tools::cmd_pipeline::builder",
  "level": "TRACE",
  "timestamp": "2022-10-12T15:49:26.735335Z",
  "fields": {
    "cmd": "ShowHtml(ShowHtml)"
  },
  "spans": [
    {
      "name": "build_pipeline",
      "arg_str": "filter-analysis simple.rs -k def -s simple::<Loader>::new | show-html\n"
    }
  ]
}
```


```json
{
  "level": "TRACE",
  "fields": {
    "output_json": "{\n  \"JsonRecords\": {\n    \"by_file\": [\n      {\n        \"file\": \"simple.rs\",\n        \"records\": [\n          {\n            \"loc\": \"00037:11-14\",\n            \"target\": 1,\n            \"kind\": \"def\",\n            \"pretty\": \"simple::<Loader>::new\",\n            \"sym\": \"simple::<Loader>::new\"\n          }\n        ]\n      }\n    ]\n  }\n}"
  },
  "timestamp": "2022-10-12T15:49:26.736039Z",
  "target": "tools::cmd_pipeline::interface",
  "spans": [
    {
      "cmd": "FilterAnalysisCommand { args: FilterAnalysis { file: \"simple.rs\", record_type: None, kind: Some(\"def\"), query_opts: SymbolicQueryOpts { symbol: Some(\"simple::<Loader>::new\"), identifier: None } } }",
      "name": "run_pipeline_step"
    }
  ]
}
```


```json
{
  "target": "tools::cmd_pipeline::interface",
  "level": "TRACE",
  "fields": {
    "output_json": "{\n  \"HtmlExcerpts\": {\n    \"by_file\": [\n      {\n        \"file\": \"simple.rs\",\n        \"excerpts\": [\n          \"<div role=\\\"row\\\" id=\\\"line-37\\\" class=\\\"source-line-with-number\\\">\\n  <div role=\\\"cell\\\"><div class=\\\"cov-strip cov-no-data\\\"></div></div>\\n  <div role=\\\"cell\\\"><div class=\\\"blame-strip\\\"></div></div>\\n  <div role=\\\"cell\\\" class=\\\"line-number\\\" data-line-number=\\\"37\\\"></div>\\n  <code role=\\\"cell\\\" class=\\\"source-line\\\">    <span class=\\\"syn_reserved\\\" >pub</span> <span class=\\\"syn_reserved\\\" >fn</span> <span data-symbols=\\\"simple::<Loader>::new\\\" data-i=\\\"25\\\" >new</span>(<span data-symbols=\\\"simple::<Loader>::new::deps_dir\\\" data-i=\\\"26\\\" >deps_dir</span>: PathBuf) -> <span class=\\\"syn_reserved\\\" >Self</span> {\\n</code>\\n</div>\\n\"\n        ]\n      }\n    ]\n  }\n}"
  },
  "timestamp": "2022-10-12T15:49:26.737078Z",
  "spans": [
    {
      "name": "run_pipeline_step",
      "cmd": "ShowHtmlCommand { args: ShowHtml }"
    }
  ]
}
```

asutherland (approved):
This is super fantastic!  Thank you for digging into this!  I'm feeling really optimistic about SCIP after seeing this, as this all seems quite excellently straightforward and like something we can potentially generalize to support scip-python in future iterations[1]!

For those following along, this will only immediately change how we process the "tests" repo and there is additional work to convert mozilla-central to generate SCIP data.  Currently that logic is located [around save-analysis flags](https://searchfox.org/mozilla-central/search?q=save-analysis&path=).

1: Given that we have mozilla-beta and mozilla-release and mozilla-esr102 still generating save-analysis for a while, it might be reasonable to simply copy-and-paste when generalizing so we can leave rust-indexer as-is until everything is transitioned over to SCIP.

tools/src/bin/rust-indexer.rs
emilio: Given this won't be used in m-c just yet I'll merge this for now and work on extra improvements for this and syntax in follow-ups.

emilio:
My rust-analyzer fix got merged and I confirmed stuff looks much better now. Will merge this as-is to keep working on it in-tree as time allows.

tools/src/bin/rust-indexer.rs
asutherland: Makes sense!  Feel free to land any follow-ups without additional review unless you particularly are looking for feedback.

PR #571, opened 2022-09-20
Bug 1612525 - Initial docker support

asutherland:
On Ubuntu, after installing the package "docker.io", it should be possible to run:
- `./build-docker.sh` to build a "searchfox" image
- `./run-docker.sh` to attach the image

Once in the VM, it should be possible to `cd /vagrant` and then `make build-test-repo` and then access the searchfox web UI via http://localhost:16995/ like could be done with the vagrant VM mechanism.


asutherland:
@emilio It would be great if you could check this out.  I'm going to land this now since it's largely additive and I haven't changed any of the docs yet.  I expect to do additional iteration passes for docs and probable refinements once I test this under WSL2 on windows.

PR #572, opened 2022-09-21
Bug 1612525 - Remove jemalloc dependency to reduce cross-platform issues.

asutherland:
This is basically a manual revert of
https://github.com/mozsearch/mozsearch/pull/275 which added a jemalloc dependency in order to be able to allow us to try and find the recursive memory cost of any data structure with the given derive.

This is being removed in the context of trying to make it easier for new contributors to searchfox to get up and running and in particular those who might be using Windows.  It seems like our use of jemalloc (at least the revisions used here) was breaking even when built under WSL2 on Windows (per my local repro) in addition to in a vagrant VMware VM due to limitations of the host NTFS file system.

As I discuss in the bug there, we haven't particularly needed these metrics since their introduction and especially in cases where we just want some one-off numbers, there are developer tools that are well suited to one-off investigations.


asutherland:
Note that the commit message unfortunately references "the bug there" where I meant https://github.com/mozsearch/mozsearch/pull/569 which I had intended to reference, but did not.

PR #575, opened 2022-09-25
Bug 1789329 - Add explicit mjs parsing, fallback module parsing, tests.

asutherland:
For JS parsing:
- If the file ends in ".mjs" we treat it as a module only.
- If the file ends in ".js" we try parsing it as a script and fall back to also trying it as a module if it doesn't parse as a script because imports/exports are illegal.

For the analysis:
- We process ExportDeclaration nodes by recursively processing the declaration it wraps.
- We otherwise ignore everything having to do with modules because our SCIP support should let us do clever things soon so it's a bit of a waste to try and do something smart since it will all just get thrown away.

I add some very basic definition checks to ensure that we've parsed the modules and the export declarations.  I don't add any other pre-existing JS checks at this time.  This is largely because we haven't really seen regressions and also because we are 100% going to have to change the symbol mapping when we move to SCIP.

### DIFF
--- a/scripts/idl-analyze.py
+++ b/scripts/idl-analyze.py
@@ -318,13 +318,7 @@ for l in lines:
     linebreaks.append(cur)
 
 if analysis:
-    # compatibility before and after bug 1633156
-    import inspect
-    initMethod = [obj for (name, obj) in inspect.getmembers(xpidl.IDLParser) if name == '__init__'][0]
-    if 'outputdir' in inspect.getargspec(initMethod).args:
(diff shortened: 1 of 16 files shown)


asutherland:
This builds on https://github.com/mozsearch/mozsearch/pull/568 which was reverted in https://github.com/mozsearch/mozsearch/pull/569 because we needed to handle the import/export statements.  I've also changed the logging to be captured by our warning mechanism and slip-stream fixed a deprecation warning I saw in idl-analyze.py while checking things.

PR #584, opened 2022-10-01
Bug 1783761 - Overhaul the per-file info pipeline.

asutherland:
This does not yet replace JS templating of output directories, but is a large atomic set of changes to overhaul how we do per-file info data aggregation.

At a high level:
- `derive-per-file-info.rs` is now gone; crossref.rs now handles the ingestion process, with the bulk of the logic existing in `repo_data_ingestion.rs`.
- There is now a `config_defaults` directory which we fall back to when looking for per-config files.  The idea is to avoid having to duplicate the same config file in every directory or to add a ton of symlinks (which can complicate windows support).
   - config.rs has been moved for clarity and some of its helpers have been moved to be impls for the classes rather than helpers that take a config.
   - The config.json files now specify their CONFIG_REPO root, which may feel a little weird, but was a simple way to use the existing data-flow.  I'll land mozsearch-mozilla changes concurrently with this, but it would be great if we could make this more clean.
- `per-file-info.toml` is our new config file that allows us to:
   - Define `path_kind` heuristics and rules so we no longer have to have the logic be hardcoded in router.py and potentially the front-end JS.  This is not yet hooked up to router.py, but that will be a logical near-term step.
   - Define text file ingestion like `.eslintignore` so we can tag/un-tag files based on their presence in such a list.  In our initial landing here, we tag files with "eslint-ignored" if the tree has an `.eslintignore` in the root.  we don't bother to understand them at any other level of the tree, but it's a thing we could do in the future.  This was just an MVP.
   - Define JSON file ingestion to replace what was previously hand-rolled data ingestion in `derive-per-file-info.rs`.
      - Practically speaking I generalized the existing logic and because of the many ways this data can be nested, we actually have 4 different `nesting` algorithms for the 4 different JSON files we ingest, and we'll probably have to add more.  Things are somewhat more sane now though because we just use a hashmap by path for lookup.  Previously we tried to build a nested hierarchy that imitated one of our input files and that was a mess.
      - These JSON ingestion files allow us to explicitly set the path_kind (or other arbitrary metadata) based on the JSON files. This means that our path_kinds for tests can be based on them existing in a manifest rather than based on a filename heuristic!
- Our test info box rendering is moved from being hardcoded in `output-file.rs` to being defined as a liquid template.  Honestly, the HTML formatting in rust wasn't horrible for our strongly typed data models, but a big change here is moving to having a generic, extensible pipeline for the JSON ingestion, and it is a major boilerplate nightmare to deal with un-typed JSON in the rust code. Also our existing HTML rendering felt like it was really at its scaling limits and could not reasonably allow, for example, webkit to customize stuff without interfering with mozilla-central.
- A related change to that is that we also use liquid templating for the WPT dashboard link.  This is potentially a baby step to moving some other panels too, but it's not a high priority.
- I added a `liquid-templating-cheatsheet.md` file since, while liquid as a templating language is pretty sane and the `liquid-rust` crate is amazing (thank you liquid-rust team, you rock!!), there are still some rough edges.  (And I think we can probably help improve the rust crate if we want to improve the situation.  It's really well engineered.)
- I have standardized our use of Ustr at least in crossref so that any file path, file path segment, symbol, or pretty identifier string is expected to be a Ustr.  Part of this logic is being careful to not try and intern a string into a Ustr until we're sure that it actually corresponds to something that should be interned. For example, if we get symbols from a query, we try and look them up from our crossref database first and then we only intern the string if it matched.
   - This is done primarily the benefit of graph processing and our future work in that space where there are potentially quite significant performance and resource wins from being able to do fast equality tests and have pre-computed hashes.
- cmd_search_files and the local index remote server and its backend have been changed to operate off of the concise info database which we eagerly load into memory.  The search-files results will now contain all the concise info we have on a file, but we still have a separate detailed info for larger hunks of data that only need to be available when rendering the file or at other times when it makes sense to load it as needed.
- The tentative plan has also been to include the file information in the crossref database under the `FILE_` symbol for consistency, but that was mainly something I thought was reasonable when I wasn't sure whether we should still have the concise-per-file-info.json file or not and load it at the start of the server.  But in the end it seemed clear that our file searching patterns could benefit from pre-computation/indexing of any relevant queries, and just loading a flat file from disk is going to be faster to feed that than doing N binary searches in the crossref, etc. etc.
   - It could still make sense to store the information in the crossref info for the files, but tha can be implemented as needed.


asutherland:
I have ideally the last dev.searchfox.org test run running for this right now.  There's already a pretty good run up but the bugzilla component mapping was disjoint because it was ingested with a leading "/" on the paths and that should now be addressed in the new run.  I'm planning to land this later tonight Eastern or Mid-Day Eastern tomorrow if all goes well.  This should then be followed up by output-dir.js being obsoleted and its other JS friends.

asutherland:
Got a clean run on dev.searchfox.org.  Going to land now.

The last issues had been:
1. NSS has some build byproducts that we don't expect, like I think files are being dynamically generated in the source tree instead of under the objdir.  This was causing failures/warnings, but now they're just info.  Arguably we should fix this up for our NSS indexing build, but I think it's also harmless and maybe not a bad edge case to have in place.
2. The coverage data ingestion was trying to evaluate the directories in a file context which was resulting in fatal errors when generating the detailed per-file info which (as it's been under `derive-per-file-info.rs` too) is its own directory tree, and you can't have a file and a directory with the same name, of course.  I've added additional data to our synthetic coverage metadata in the "tests" repo to get coverage of this.  I also moved a bunch of things from unwrap() to more explicit error handling[1].

As examples of what the contents of `concise-per-file-info.json` look like:
```json
  "dom/indexedDB/ActorsParent.cpp": {
    "path_kind": "Normal",
    "file_size": 705680,
    "bugzilla_component": [
      "Core",
      "Storage: IndexedDB"
    ],
    "subsystem": "Core/Storage/IndexedDB",
    "tags": [],
    "description": null,
    "info": {}
  },
```

```json
  "dom/imptests/testharnessreport.js": {
    "path_kind": "Test files",
    "file_size": 10646,
    "bugzilla_component": [
      "Core",
      "DOM: Core & HTML"
    ],
    "subsystem": "Core/DOM/Core HTML",
    "tags": [],
    "description": "Dictionary mapping a test URL to either the string \"all\", which means that\n all tests in this file are expected to fail, or a dictionary mapping test\n names to either
 the boolean |true|, or the string \"debug\". The former\n means that this test is expected to fail in all builds, and the latter\n that it is only expected to fail in debug builds.\n   ",
    "info": {}
  },
```

```json
  "dom/html/test/test_window_open_close.html": {
    "path_kind": "Test files",
    "file_size": 1366,
    "bugzilla_component": [
      "Core",
      "DOM: Core & HTML"
    ],
    "subsystem": "Core/DOM/Core HTML",
    "tags": [],
    "description": null,
    "info": {
      "test": {
        "failure_count": null,
        "skip_if": "(toolkit == \"android\" && debug && !is_fennec) || (os == \"linux\") || (os == \"win\" && debug && bits == 64)"
      }
    }
  },
```




1: For the processing pipeline, there are a bunch of `expect()`-based fail-fast modes for data ingestion configuration problems because we want those to fail ASAP and ideally directly in the "tests" repo rather than manifesting as missing/corrupt data.

PR #587, opened 2022-10-03
Bug 1783761 - Switch directory listings to rust from JS

asutherland:
This leaves the JS machinery in place to generate directory listings (although I removed some dead JS code), commenting out their invocation in favor of the new rust rendering.

Behaviorally, we see the following changes:
- Our sort changes slightly.  The JS code was using localeCompare and although I am very exctied that the "icu" crate has reached 1.0, I ended up choosing https://crates.io/crates/lexical-sort and its `natural_lexical_cmp` method for sorting.  I hope this will generally be an improvement, but some things do change, like we're now sorting in a case-insensitive fashion and we'll try and sort numbers naturally instead of the normal ASCII robotic way.
  - This was necessitated by the direct codepoint comparison sorting "__GENERATED__" between UpperCaseFiles and lowerCaseFiles which is super-weird.
  - Note that there actually was a very hard-coded ordering hack for __GENERATED__ and files that started with "." in output-dir.js and we are not trying to propagate them because I think the natural sort order seems sufficiently sane.
- We stop trying to use the source images themselves as their own icons for files that look like images.  This was accomplished by pointing at the hg server and having the browser fetch the full-size image.  The hg server is already slow/overloaded as-is, and for large images this seems undesirable, so I've removed it.
  - Note that we were also pointing at the "tip" revision and so this would not immediately translate to us supporting directory listings for specific revisions.
  - I'm not opposed to adding it back, but it should be ready for handling historical revisions.  (Noting that we will need to update the existing template for this too; maybe we should add a liquid filter to help with the link revision logic.)

In terms of infrastructure:
- We now generate concise and detailed information for directories, not just files.  However, the search-files command only returns files by default.  Also, although we write out detailed info for the directories, we currently will never read it back in because we don't have any need for that right now.  But, for example, it might be useful for directory image thumbnailing or some other case where we want to have rich data that exists exclusively for directory listings.


asutherland:
I've pushed a try run to the "dev" channel and if there are no obvious regressions I missed in local testing when doing A/B between searchfox.org and dev.searchfox.org for m-c on various dir trees, I'll land this.  (As noted in the commit message, some directory listing orderings will absolutely change.  I think feedback is likely most beneficial if people notice and find the difference big enough to note in the channel, whereas if I bring it up in the channel before anyone has a chance to notice it's likely to bias their experience enough that they will absolutely notice and maybe find the change more jarring.)

asutherland:
Whoops!  Apparently there are 334k+ files in mozilla-central and so a limit of 32k will tend to truncate the file listings a bit!

asutherland:
All 3 problems have been addressed and I've added checks for the source file info boxes for the fake WPT files, as well as verified that we now have the right data in the existing checks that cover the fake WPT files.

1. There's now a `path_prefix` config option.
2. jmaher implemented https://phabricator.services.mozilla.com/D158527 on newly filed https://bugzilla.mozilla.org/show_bug.cgi?id=1793486 to add a `_filename` field to all the WPT data.  I've made this `_filename` field required (we skip WPT entries if they don't exist) in the interest of avoiding a half-working situation.  This means that we won't have the data for anything but m-c until that fix makes its way there.  Since we generally only care about m-c for WPTs, there's no need to expedite this.
3. None of the file ingestion mechanisms will create files anymore.  Only the file-list ingestions will do this.  (And the textfile ".eslintignore" mechanism only ever worked on the known file list since it's a list of globs, not a list of concrete paths.)

One behavioral change within the ingestion mechanism is that the "object" mechanism where we evaluate a bunch of nodes to produce an object instead of just evaluating nodes to produce a single key/value pair now behaves the same as the node evaluation in that null values will not be written.

This is done primarily because for the WPT records that we're ingesting there are only keys when there is a value, and it seemed undesirable to be normalizing these into existence as keys with null values under our "info" expando hierarchy.  It uses up memory and decreases the signal/noise content of the concise object info, etc.  Arguably if our mechanism had a better concept of undefined versus null, this could avoid create ambiguity for future data we ingest where the existence of a key with a a null value might have significance.  Unfortunately, serde_json doesn't have undefined because JSON doesn't have undefined so that would be a bit more involved.

asutherland:
Okay, that gave us a clean run modulo a minor templating change to get us green test info boxes because of the change to suppress null keys.

Having the dot-files sort at the top definitely feels wrong at the root of the tree because there's just so many dot files, but I am finding it interesting/useful for cases like https://searchfox.org/mozilla-central/source/dom/indexedDB/test where there's an `.eslintrc.js` in the directory.  It's a very notable presence and seeing it at the top is nice.

That said, it wouldn't surprise me if people notice this change and there's popular acclaim to restore it to its previous custom collation behavior or to do what github does which is put the folders first and then sort the files with dot-files first.  A patch will be gladly accepted for that!  There's also some potential to let people have a custom synchronous JS sort on page load based on a setting from the very soon settings page.

PR #588, opened 2022-10-04
Bug 1783761 - Move remaining JS templating to rust.

asutherland:
This switches the search template and help.html templating to be done from rust instead of from `output-template.js` and `output-help.js`.

I have left the JS templating code in place for now, but commented out in `output.sh` and will file a follow-up bug to remove them and `output-dir.js` after it's clear we don't need to switch back and any A/B checking by interested parties is completed.


asutherland:
Whoops, I put the new footer outside the scrolling area which is amusing but not desirable.  This did not impact the source listings which is how this was not more obvious.  One more landing coming soon and I've terminated the utc22 indexer to avoid anyone seeing that.

PR #590, opened 2022-11-03
search: Escape a little bit more.

emilio:
This fixes Mozilla bug 1798496.

### DIFF
--- a/static/js/search.js
+++ b/static/js/search.js
@@ -338,6 +338,10 @@ function populateResults(data, full, jumpToSingle) {
     return html;
   }
 
+  function escape(s) {
+    return s.replace(/&/gm, "&amp;").replace(/</gm, "&lt;");
+  }
+
   function renderSingleSearchResult(pathkind, qkind, file, line, isContext) {
     var [start, end] = line.bounds || [0, 0];
     var before = line.line.slice(0, start).replace(/^\s+/, "");
@@ -357,10 +361,6 @@ function populateResults(data, full, jumpToSingle) {
       "</a></td>";
     html += "<td><a href='" + makeURL(file.path) + "#" + line.lno + "'>";
 
-    function escape(s) {
-      return s.replace(/&/gm, "&amp;").replace(/</gm, "&lt;");
-    }
-
     html += "<code>";
     html += escape(before);
     html += "<b>" + escape(middle) + "</b>";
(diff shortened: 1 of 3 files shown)


emilio:
So I added something to the test repo but I'm not sure we have an easy way to test this code-path, is that right @asutherland?

asutherland:
> So I added something to the test repo but I'm not sure we have an easy way to test this code-path, is that right @asutherland?

Yeah, unfortunately we don't have anything that simulates a browser running the client-side JS code right now.  I've also avoided it a little bit just because it potentially brings in a bunch of new machinery that tends to be more on the brittle and slow side[1] if we go with a full fidelity browser, or we end up with a lower fidelity solution that potentially isn't realistic.  For something like escaping, we could adopt eslint rules/plugins like what we use in m-c already, although that might require some changes to how the code works.  Alternately, there are template literal interpolation mechanics that would automatically escape all content unless it's explicitly blessed to be known safe HTML, etc.  That said, I'm not opposed to us adding a browser mechanism, although I fear a lot of the more plug-and-play options use chromium.

That said, one change I have made with how I approached the "query" endpoint was that it uses "liquid" templates to render the "query" results, and we can evaluate the [templates](https://github.com/mozsearch/mozsearch/tree/master/tools/templates) on our server in rust, or optionally use the LiquidJS runtime to serve JSON to the client and have it render in the browser, like we do right now for "search".  Right now the test cases for that are only using the JSON rep (ex: [input](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/inputs/web/query/execution/simple/overs.cpp/query__doublePure__context4__json) and [expected output](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/snapshots/web/query/execution/simple/overs.cpp/check_glob%40query__doublePure__context4__json.snap)) but my plan is to add HTML checks too.  I've added the ability to use selectors in some of the HTML outputting searchfox-tool pipeline steps (ex: [check input for test info box HTML output](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/inputs/web/files/test-info-boxes/wpt__test_ima_disabled_wpt__html) and [output](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/snapshots/web/files/test-info-boxes/check_glob%40wpt__test_ima_disabled_wpt__html.snap)), so we can have the check be targeted to a representative excerpt of the templated output, for example, a specific line result, rather than having like a massive un-readable HTML diff whenever we change things.

1: Mainly there is overhead to browser startup.

asutherland (approved):

PR #591, opened 2022-11-15
Bug 1797153 - patch traitobject to use destructure_traitobject

asutherland:
We're using the mechanism documented at
https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html to reference destructure_traitobject when its Cargo.toml still claimed to be traitobject.

### DIFF
--- a/tools/Cargo.lock
+++ b/tools/Cargo.lock
@@ -3045,8 +3045,7 @@ dependencies = [
 [[package]]
 name = "traitobject"
 version = "0.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


asutherland:
@emilio I'm going to land this now because this seems to locally fix things for me and we actually end up doing exactly what https://github.com/saltyrtc/saltyrtc-client-rs/tree/e8689830dba1b0346c87f3dd31d2dbbe62a9c159#dependency-patching suggests but you may have better approaches, noting that as I say in https://bugzilla.mozilla.org/show_bug.cgi?id=1797153 comment 0 I think when we tried to upgrade hyper.rs before we ran into changes to the process model, so I don't think fixing by upgrading hyper is necessarily going to work, but I also didn't look if there are very unlikely dot releases.

PR #593, opened 2022-11-22
Format `.m` files the same way `.mm` files are formatted

padenot:


padenot (commented):
This happens in a vendored dependency in `mozilla-central`: https://searchfox.org/mozilla-central/source/third_party/libwebrtc/sdk/objc/components/capturer/RTCCameraVideoCapturer.m. Apparently it's a objective-c file that doesn't contain C++ idioms (more like objective-C++ then).

emilio (approved):

PR #595, opened 2022-11-29
Bug 1801960 - Widen the context-menu word regexp to include #.

asutherland:
As :saschanaz pointed out, the context-menu's "search for the substring" option is excluding the leading "#" for private symbols still, which is weird.

This fix naively just widens the regex.  I imagine there could be regressions for this and it could make sense for this JS to instead understand the difference between:
- The user clicking on a symbol for which we can know the exact symbol string to use.
- The user clicking on random text which is not marked up as a symbol and for which we should use a regex heuristic.
- The user selecting a region of text and then clicking on it with the intent to potentially provide our context menu to search for that string, but where we will currently erase the user's own selection.

### DIFF
--- a/static/js/context-menu.js
+++ b/static/js/context-menu.js
@@ -259,7 +259,13 @@ function getTargetWord() {
 
   function isWordChar(character) {
     // TODO: this could be more non-ascii friendly.
-    return /[A-Z0-9_]/i.test(character);
+    //
+    // Notable Changes:
+    // - We have added "#" to deal with JS private symbols.  This will widen
+    //   C preprocessor directives to include the leading #, which makes sense.
+    //   This will also impact use of the "stringizing operator" for macros,
+    //   where it won't be what we want.
+    return /[#A-Z0-9_]/i.test(character);
   }
 
   if (offset < string.length && !isWordChar(string[offset])) {


static/js/context-menu.js
```diff
@@ -259,7 +259,13 @@ function getTargetWord() {
...
   function isWordChar(character) {
     // TODO: this could be more non-ascii friendly.
-    return /[A-Z0-9_]/i.test(character);
+    //
+    // Notable Changes:
+    // - We have added "#" to deal with JS private symbols.  This will widen
+    //   C preprocessor directives to include the leading #, which makes sense.
+    //   This will also impact use of the "stringizing operator" for macros,
+    //   where it won't be what we want.
+    return /[#A-Z0-9_]/i.test(character);
```
emilio, line 268: This is probably fine. We could perhaps add some extra logic to consider `#` only at the beginning of the word, but that's probably not worth it, I don't think macro-pasting operators are common enough to warrant that.

emilio (approved):

static/js/context-menu.js
asutherland: Agreed that specializing for the start of the word could make sense.  There's probably a lot more we could do for the text selection here.  (Ideally by adding tests.  This specific logic is thankfully a case where we could unit test in node.js/deno without too much trouble.)

One interesting possibility with [the proposed enhancement to use tree-sitter as a tokenizer](https://bugzilla.mozilla.org/show_bug.cgi?id=1796870) would be that it presumably would have a much better idea of what are language-appropriate tokens, so in many cases we might be able to just rely on the syntax highlighting spans more.

PR #599, opened 2023-01-10
Set JS icon for mjs files

nchevobbe:
I didn't actually test this, but I think this should work


asutherland:
Yeah, this should fix this for the "search" endpoint.  And it looks like [I already made the CSS change for the "query" endpoint when adding icon support to the "query" rendering](https://github.com/mozsearch/mozsearch/blob/4be27ac6b6d726b11098beea338661366acad853/static/css/icons.css#L204-L206) so we should be good.  Thanks!

PR #600, opened 2023-01-13
Bug 1809987 - Update CSS icon font and replace use of inline SVG.

asutherland:


asutherland:
I've landed this eagerly in order to try and get eyes/opinions on this for Friday which seems like the best-case trade-off between potential disruption and getting feedback.  Please see [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1809987) and the comments in the commit for any context.

asutherland:
There was a minor regression in the context menu due to it surprisingly using icons and the context-menu unfortunately being something we don't have test coverage for because it's a runtime JS thing.  I rolled back to the spare and killed the buggy web-server, although I also could have patched the `context-menu.js` logic to mitigate but think it's better to defer anyone seeing any UX changes until tomorrow's utc10 run when I'll be awake again.  I've already landed the minor fix as part of https://github.com/mozsearch/mozsearch/pull/601 (which is more extensive than just the `context-menu.js` changes and does require a full run).

PR #603, opened 2023-01-17
Move from ANALYSIS_DATA to SYM_INFO

asutherland:


asutherland:
This is up on dev.searchfox.org and I believe I've validated that feature parity is maintained.  (In particular, I made a bunch of fixes/cleanups in response to the more extensive set of permutations that the real mozilla-central gives us.)

The main change in behavior is that for IDL symbols we are more thorough in how we handle the "binding slots" and so we end up with a bunch more context menu options because we are able to see and offer both "go to def" and "go to decl".  Note that "Go to decl" is also new and, somewhat ironically as I just announce the new settings functionality, I'm not providing a setting to disable "go to decl" at this time.  I'm certainly not opposed to adding the setting if people find it jarring.  Also, we actually do spell out "declaration", and I'd be fine with shortening that, etc.

I'm going to re-trigger config1 immediately as I did terminate the utc22 indexer knowing I'd get to this.  I will manually validate config1 after it comes online before I go to bed.

PR #612, opened 2023-02-09
[WIP] Bug 1799579 - Handle OverloadExpr in MozSearchIndexer

HighCommander4:
Posting a WIP patch for [bug 1799579](https://bugzilla.mozilla.org/show_bug.cgi?id=1799579).

The patch partially fixes the bug, in that the call to `Project` is now clickable and offers the menu item "Search for function `Foo::Project`", but it does not yet offer the menu item "Go to definition of `Foo::Project`", which we'd like as well.

I will continue investigating why it doesn't offer that second menu item, but in the meantime I wanted to post what I have.

---

Also, a note on why I didn't use the `AutoTemplateContext` machinery:

While `OverloadExpr` is (usually, and in this testcase in particular) dependent, it's a bit different from `CXXDependentScopeMemberExpr` in that candidate declarations being referenced are available, and it's only the selection among the candidates that's deferred until instantiation (whereas, with `CXXDependentScopeMemberExpr`, candidate declarations are not available prior to instantiation because the scope in which name lookup would have to be performed to find the candidates is itself not available).

As a result, we can provide heuristic results for `OverloadExpr` without looking at instantiations (namely, just provide all the candidates), and I think this is useful enough for a first pass.

I think there's an opportunity to _improve_ the results by using the `AutoTemplateContext` machinery to discover which candidates are actually used by instantiations and potentially narrow the results that way, but I'm proposing to defer that to a future improvement.

(I should definitely expand the testcase to include a scenario with multiple overloads to illustrate the behaviour, I will do that in an upcoming revision of this patch.)

### DIFF
--- a/clang-plugin/MozsearchIndexer.cpp
+++ b/clang-plugin/MozsearchIndexer.cpp
@@ -1980,6 +1980,26 @@ public:
     return true;
   }
 
+  bool VisitOverloadExpr(OverloadExpr *E) {
+    SourceLocation Loc = E->getExprLoc();
+    normalizeLocation(&Loc);
+    if (!isInterestingLocation(Loc)) {
+      return true;
+    }
+
+    for (auto *Candidate : E->decls()) {
+      if (TemplateDecl *TD = dyn_cast<TemplateDecl>(Candidate)) {
+        Candidate = TD->getTemplatedDecl();
+      }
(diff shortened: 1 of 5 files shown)


HighCommander4:
(Apologies, posted under wrong account. Correct PR at https://github.com/mozsearch/mozsearch/pull/613.)

PR #615, opened 2023-03-11
Update ipdl_parser to latest

staktrace:
This picks up a new (forked) version of lalrpop
to get rid of the build warnings.


emilio (approved):
Yeah I kinda didn't want to use s fork but given there's no progress on lalrpop to fix the warning is rather annoying... Let's do this, thanks!

PR #626, opened 2023-04-21
rust: Prefer scip index if available.

emilio:
With the latest mozilla-config and scip changes, this gives me a working mozilla-central index.

There are some regressions that I noticed that might be worth addressing first:

 * No stdlib symbols: If you have `use std::io::Write;`, clicking `Write` won't find all references to it. Not super-surprising given we have a bunch of extra code to deal with the rust stdlib in save analysis, but maybe we can do something about it.

 * No module definitions: If you have `mod foo;`, clicking `foo` won't show the "Go to definition" button to get you to the file.

 * #[no_mangle] functions keep the rust-module-namespaced symbol names. I'm not sure how easy to address is this without upstream fixes, we'll see.

Other than that, it seems to be working, and maybe we can even enable it by default if those regressions get worked on or if we consider they're not such a huge deal.

### DIFF
--- a/scripts/rust-analyze.sh
+++ b/scripts/rust-analyze.sh
@@ -34,12 +34,11 @@ GENERATED_SRC=$4
 SF_ANALYSIS_OUT=$5
 
 if [ -d "$RUST_ANALYSIS_IN" ]; then
-  INPUTS="$(find $RUST_ANALYSIS_IN -type d -name save-analysis)"
-  SCIP_FLAGS=""
+  INPUTS="$(find $RUST_ANALYSIS_IN -type f -name rust.scip)"
+  SCIP_FLAGS="--scip --scip-prefix $RUST_ANALYSIS_IN"
   if [ "x$INPUTS" = "x" ]; then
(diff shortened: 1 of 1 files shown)


emilio:
cc @asutherland

asutherland:
Woo!

Judging from the symbols, it looks like this is what's on dev.searchfox.org right now, and the results seem pretty great!  And given that I guess save-analysis just fell over and you had to turn it off, it does seem like this makes sense to switch to explicitly switch to favoring rust-analyzer for consistency for beta and release.  (At least, I assume that's the implication?  That with save-analysis gone on nightly we'll definitely start using SCIP, and so it's the older channels where there's a choice to make, at least for those that had the save-analysis turned on recently enough.)

The loss of going to module definitions I think may end up being noticeable but since you can still go to explicit imports I don't think it's that big a deal since normally one would use the more explicit definition anyways, and probably would find the symbol at its point-of-use not at the import.

emilio:
Yeah, let's do this :)

PR #628, opened 2023-04-25
Bug 1828223 - Implement heuristic resolution of CXXDependentScopeMemberExpr

theres-waldo:
Posting a draft fix for bug 1828223 for feedback.

I would appreciate feedback on two points in particular:

 1. Currently, in `IndexConsumer::VisitCXXDependentScopeMemberExpr()`, I both (1) produce a heuristic result (that's what's new in this patch), **and** (2) record the location for possibly producing concrete results when visiting instantiations (if we have any) during the `AnalyzeDependent` phase (what the code was doing before).
 
    I'm wondering, should we add some logic to **only** produce the heuristic result if we didn't end up producing concrete results from the `AnalyzeDependent` phase?

    I feel like I could go either way on this. On the one hand, the indexer might only know about a subset of the instantiations actually occurring in the project, and the concrete results might therefore miss something that the heuristic results catch. On the other hand, if e.g. we're working with an overload set, it's possible that all instantiations end up using only a subset of the overloads, and the heuristic results would add noise by providing all overloads. In other words, I don't think either approach is _categorically_ more accurate than the other.

 2. The `HeuristicResolver` facility added in this patch is a slimmed-down version of clangd's (currently much more featureful) facility [of the same name](https://searchfox.org/llvm/rev/be9c91843bab5bb46574c27836bfcd9ad6fc9ef5/clang-tools-extra/clangd/HeuristicResolver.h#45).

    My current approach is to port over pieces of this facility to Searchfox as we need them to make specific testcases work.

    However, one could imagine a different approach of just... copying the class from clangd wholesale, as a piece of code which has been battle-tested in a different project? (It doesn't have any dependencies on anything else in clangd, only public clang APIs.)

    Let me know if you have a preference here, I'm happy to switch approaches.

    (Granted, even better would be to upstream `HeuristicResolver` to `libClangTooling` or somesuch and expose it as a public clang API, and have both clangd and Searchfox consume it from there. I'd like to explore this, but I'd prefer not to block our progress on Searchfox enhancements on this LLVM change, because LLVM patches can be slow to be reviewed (and then we'd need to adopt the new LLVM version in which it appears and so on.))

### DIFF
--- a/clang-plugin/MozsearchIndexer.cpp
+++ b/clang-plugin/MozsearchIndexer.cpp
@@ -207,6 +207,49 @@ public:
                       const MacroDefinition &Md) override;
 };
 
+// Performs heuristic resolution of entities in template code.
+class HeuristicResolver {
+public:
+  // Try to heuristically resolve certain types of expressions to one or more
+  // likely-referenced declarations.
+  static std::vector<const NamedDecl *>
+  resolveMemberExpr(const CXXDependentScopeMemberExpr *ME) {
+    const Type *BaseType = ME->getBaseType().getTypePtrOrNull();
+    return resolveDependentMember(BaseType, ME->getMember());
+  }
+
+private:
+  // Given a dependent class type and a member name, heuristically resolve the
+  // name to one or more declarations. The current heuristic is simply to look
+  // up the name in the primary template. This is a heuristic because the
+  // template could potentially have specializations that declare different
+  // members. Multiple declarations could be returned if the name is overloaded
+  // (e.g. an overloaded method in the primary template).
+  static std::vector<const NamedDecl *>
+  resolveDependentMember(const Type *T, DeclarationName Name) {
+    if (!T)
+      return {};
+
+    const auto *TST = T->getAs<TemplateSpecializationType>();
+    if (!TST) {
+      return {};
+    }
+
+    const ClassTemplateDecl *TD = dyn_cast_or_null<ClassTemplateDecl>(
+        TST->getTemplateName().getAsTemplateDecl());
+    if (!TD)
+      return {};
+
+    CXXRecordDecl *RD = TD->getTemplatedDecl();
+    if (!RD->hasDefinition())
+      return {};
+
+    RD = RD->getDefinition();
+    return RD->lookupDependentName(Name,
+                                   [](const NamedDecl *) { return true; });
+  }
+};
+
 class IndexConsumer : public ASTConsumer,
                       public RecursiveASTVisitor<IndexConsumer>,
                       public DiagnosticConsumer {
@@ -2009,6 +2052,17 @@ public:
       return true;
     }
 
+    // If possible, provide a heuristic result without instantiation.
+    for (const NamedDecl *D : HeuristicResolver::resolveMemberExpr(E)) {
+      if (const FunctionDecl *F = dyn_cast<FunctionDecl>(D)) {
+        std::string Mangled = getMangledName(CurMangleContext, F);
+        visitIdentifier("use", "function", getQualifiedName(F), Loc, Mangled,
+                        F->getType(), getContext(Loc));
+      }
+    }
+
+    // Alsoo record this location so that if we have instantiations, we can
+    // gather more accurate results from them.
     if (TemplateStack) {
       TemplateStack->visitDependent(Loc);
     }
--- a/tests/tests/checks/inputs/analysis/cpp/bug1781178.cpp/Point_IsThereOne
+++ b/tests/tests/checks/inputs/analysis/cpp/bug1781178.cpp/Point_IsThereOne
@@ -1,0 +1,1 @@
+filter-analysis bug1781178.cpp -i Point::IsThereOne
--- a/tests/tests/checks/snapshots/analysis/cpp/bug1781178.cpp/check_glob@Foo_Project.snap
+++ b/tests/tests/checks/snapshots/analysis/cpp/bug1781178.cpp/check_glob@Foo_Project.snap
@@ -4,45 +4,45 @@ expression: "&json_results"
 ---
 [
   {
-    "loc": "00006:7-14",
+    "loc": "00008:7-14",
     "source": 1,
-    "nestingRange": "6:25-6:26",
+    "nestingRange": "8:25-8:26",
     "syntax": "def,function",
     "type": "void (Point<F>)",
     "pretty": "function Foo::Project",
     "sym": "_ZN3Foo7ProjectE5PointITL0__E"
   },
   {
-    "loc": "00006:7-14",
+    "loc": "00008:7-14",
     "target": 1,
     "kind": "def",
     "pretty": "Foo::Project",
     "sym": "_ZN3Foo7ProjectE5PointITL0__E",
     "context": "Foo",
     "contextsym": "T_Foo",
-    "peekRange": "6-6"
+    "peekRange": "8-8"
   },
   {
-    "loc": "00009:7-14",
+    "loc": "00011:7-14",
     "source": 1,
-    "nestingRange": "9:35-9:36",
+    "nestingRange": "11:35-11:36",
     "syntax": "def,function",
     "type": "void (Point<F>, Point<F>)",
     "pretty": "function Foo::Project",
     "sym": "_ZN3Foo7ProjectE5PointITL0__ES2_"
   },
   {
-    "loc": "00009:7-14",
+    "loc": "00011:7-14",
     "target": 1,
     "kind": "def",
     "pretty": "Foo::Project",
     "sym": "_ZN3Foo7ProjectE5PointITL0__ES2_",
     "context": "Foo",
     "contextsym": "T_Foo",
-    "peekRange": "9-9"
+    "peekRange": "11-11"
   },
   {
-    "loc": "00013:4-11",
+    "loc": "00015:4-11",
     "source": 1,
     "syntax": "use,function",
     "type": "void (Point<F>)",
@@ -50,7 +50,7 @@ expression: "&json_results"
     "sym": "_ZN3Foo7ProjectE5PointITL0__E"
   },
   {
-    "loc": "00013:4-11",
+    "loc": "00015:4-11",
     "source": 1,
     "syntax": "use,function",
     "type": "void (Point<F>, Point<F>)",
@@ -58,7 +58,7 @@ expression: "&json_results"
     "sym": "_ZN3Foo7ProjectE5PointITL0__ES2_"
   },
   {
-    "loc": "00013:4-11",
+    "loc": "00015:4-11",
     "target": 1,
     "kind": "use",
     "pretty": "Foo::Project",
@@ -67,7 +67,7 @@ expression: "&json_results"
     "contextsym": "_ZN3Foo3BarEv"
   },
   {
-    "loc": "00013:4-11",
+    "loc": "00015:4-11",
     "target": 1,
     "kind": "use",
     "pretty": "Foo::Project",
--- a/tests/tests/checks/snapshots/analysis/cpp/bug1781178.cpp/check_glob@Point_IsThereOne.snap
+++ b/tests/tests/checks/snapshots/analysis/cpp/bug1781178.cpp/check_glob@Point_IsThereOne.snap
@@ -1,0 +1,24 @@
+---
+source: tests/test_check_insta.rs
+expression: "&json_results"
+---
+[
+  {
+    "loc": "00002:7-17",
+    "source": 1,
+    "syntax": "decl,function",
+    "type": "_Bool (void)",
+    "pretty": "function Point::IsThereOne",
+    "sym": "_ZN5Point10IsThereOneEv"
+  },
+  {
+    "loc": "00002:7-17",
+    "target": 1,
+    "kind": "decl",
+    "pretty": "Point::IsThereOne",
+    "sym": "_ZN5Point10IsThereOneEv",
+    "context": "Point",
+    "contextsym": "T_Point",
+    "peekRange": "2-2"
+  }
+]
--- a/tests/tests/checks/snapshots/analysis/cpp/bug1781178.cpp/check_glob@Point_IsThereOne.snap
+++ b/tests/tests/checks/snapshots/analysis/cpp/bug1781178.cpp/check_glob@Point_IsThereOne.snap
@@ -20,5 +20,22 @@ expression: "&json_results"
     "context": "Point",
     "contextsym": "T_Point",
     "peekRange": "2-2"
+  },
+  {
+    "loc": "00021:4-14",
+    "source": 1,
+    "syntax": "use,function",
+    "type": "_Bool (void)",
+    "pretty": "function Point::IsThereOne",
+    "sym": "_ZN5Point10IsThereOneEv"
+  },
+  {
+    "loc": "00021:4-14",
+    "target": 1,
+    "kind": "use",
+    "pretty": "Point::IsThereOne",
+    "sym": "_ZN5Point10IsThereOneEv",
+    "context": "TemplateFunc",
+    "contextsym": "_Z12TemplateFuncv"
   }
 ]
--- a/tests/tests/checks/snapshots/web/dirs/check_glob@root_listing__html.snap
+++ b/tests/tests/checks/snapshots/web/dirs/check_glob@root_listing__html.snap
@@ -191,7 +191,7 @@ expression: "&fb.contents"
         <tr>
           <td><a href="/tests/source/bug1781178.cpp" class="mimetype-fixed-container mimetype-icon-cpp">bug1781178.cpp</a></td>
           <td class="description"><a href="/tests/source/bug1781178.cpp" title=""></td>
-          <td><a href="/tests/source/bug1781178.cpp">245</a></td>
+          <td><a href="/tests/source/bug1781178.cpp">346</a></td>
         </tr>
 
         <tr>
--- a/tests/tests/files/bug1781178.cpp
+++ b/tests/tests/files/bug1781178.cpp
@@ -1,4 +1,6 @@
-template <typename> struct Point {};
+template <typename> struct Point {
+  bool IsThereOne();
+};
 
 template <typename>
 struct Foo {
@@ -13,3 +15,8 @@ struct Foo {
     Project(p);
   }
 };
+
+template <typename T> void TemplateFunc() {
+  Point<T> p;
+  p.IsThereOne();
+}


asutherland:
Just as a timeline update, I was at a team meet-up last week and should be able to get to this this week, although it might be another day or two before the jet lag is out of my system.

asutherland:
## Investigation of Changes using gfx/2d/Matrix.h

Thanks for doing the dev run, this makes it much easier to dig into the impacts of the changes!  Given your links to https://searchfox.org/mozilla-central/source/gfx/2d/Matrix.h from https://bugzilla.mozilla.org/show_bug.cgi?id=1781178 I'm using that as a basis for comparison by doing (roughly):
```shell
$ curl https://searchfox.org/mozilla-central/raw-analysis/gfx/2d/Matrix.h > trunk-Matrix.h
$ curl https://dev.searchfox.org/mozilla-central/raw-analysis/gfx/2d/Matrix.h > dev-Matrix.h
$ diff -U3 trunk-Matrix.h dev-Matrix.h         
```

There are a lot of changes, so I'm just going to broadly categorize them.  I should also note the files aren't particularly sorted as-is, so it may be useful to pass them through `sort` as well. 

### New Uses show up!

Woooo!  This is really fantastic!

### Redundant target records with varying contextsym values, example of IsIdentity

On trunk, [this use of IsIdentity](https://searchfox.org/mozilla-central/rev/e6cb503ac22402421186e7488d4250cc1c5fecab/gfx/2d/Matrix.h#78-80) is not linked:
```cpp
if (aMatrix.IsIdentity()) {
  return aStream << "[ I ]";
}
```

We now get the use with the following new target records:
```json
{"loc":"00078:16-26","target":1,"kind":"use","pretty":"mozilla::gfx::BaseMatrix::IsIdentity","sym":"_ZNK7mozilla3gfx10BaseMatrix10IsIdentityEv","context":"mozilla::gfx::operator<<","contextsym":"_ZN7mozilla3gfxlsERNSt6__ndk113basic_ostreamIcNS1_11char_traitsIcEEEERKNS0_10BaseMatrixIT_EE"}
{"loc":"00078:16-26","target":1,"kind":"use","pretty":"mozilla::gfx::BaseMatrix::IsIdentity","sym":"_ZNK7mozilla3gfx10BaseMatrix10IsIdentityEv","context":"mozilla::gfx::operator<<","contextsym":"_ZN7mozilla3gfxlsERSoRKNS0_10BaseMatrixIT_EE"}
{"loc":"00078:16-26","target":1,"kind":"use","pretty":"mozilla::gfx::BaseMatrix::IsIdentity","sym":"_ZNK7mozilla3gfx10BaseMatrix10IsIdentityEv","context":"mozilla::gfx::operator<<","contextsym":"_ZN7mozilla3gfxlsERNSt3__113basic_ostreamIcNS1_11char_traitsIcEEEERKNS0_10BaseMatrixIT_EE"}
```

and the following new source record:
```json
{"loc":"00078:16-26","source":1,"syntax":"function,use","pretty":"function mozilla::gfx::BaseMatrix::IsIdentity","sym":"_ZNK7mozilla3gfx10BaseMatrix10IsIdentityEv","type":"_Bool (void) const"}
```

The difference between the target records is their contextsym, they otherwise are the same.  Specifically, we have:
- `_ZN7mozilla3gfxlsERNSt6__ndk113basic_ostreamIcNS1_11char_traitsIcEEEERKNS0_10BaseMatrixIT_EE`
- `_ZN7mozilla3gfxlsERSoRKNS0_10BaseMatrixIT_EE`
- `_ZN7mozilla3gfxlsERNSt3__113basic_ostreamIcNS1_11char_traitsIcEEEERKNS0_10BaseMatrixIT_EE`

No normalization happens here because [target records are de-duplicated on their overall hash only](https://github.com/mozsearch/mozsearch/blob/3d93252576a6ba59c2724d95f55819f26095f618/tools/src/file_format/merger.rs#L47-L48).  This raises the question of whether the crossref database somehow consolidates these.  We do not expect to see (and in fact do not see) redundant search results because `router.py` explicitly only allows a single result per line; this is primarily intended to suppress text search results from being displayed when we already showed semantic results, but also could be doing something here.

If we do a sorch lookup by adding a bash function like the following:
```bash
dev-sorch-raw () {
  QUERY=$1
  curl -H 'Accept: application/json' "https://dev.searchfox.org/mozilla-central/sorch?q=${QUERY}&path=" | jq '.'
}
```

and then do `dev-sorch-raw mozilla::gfx::BaseMatrix::IsIdentity` we do in fact see the redundant lines in our uses path hit list:
```json
            {
              "path": "gfx/2d/Matrix.h",
              "path_kind": "Normal",
              "lines": [
                {
                  "lno": 78,
                  "bounds": [
                    12,
                    22
                  ],
                  "line": "if (aMatrix.IsIdentity()) {",
                  "context": "mozilla::gfx::operator<<",
                  "contextsym": "_ZN7mozilla3gfxlsERNSt6__ndk113basic_ostreamIcNS1_11char_traitsIcEEEERKNS0_10BaseMatrixIT_EE"
                },
                {
                  "lno": 78,
                  "bounds": [
                    12,
                    22
                  ],
                  "line": "if (aMatrix.IsIdentity()) {",
                  "context": "mozilla::gfx::operator<<",
                  "contextsym": "_ZN7mozilla3gfxlsERSoRKNS0_10BaseMatrixIT_EE"
                },
                {
                  "lno": 78,
                  "bounds": [
                    12,
                    22
                  ],
                  "line": "if (aMatrix.IsIdentity()) {",
                  "context": "mozilla::gfx::operator<<",
                  "contextsym": "_ZN7mozilla3gfxlsERNSt3__113basic_ostreamIcNS1_11char_traitsIcEEEERKNS0_10BaseMatrixIT_EE"
                }
              ]
            },
```

Because the contextsym is used for magic diagrams, it's probably appropriate to enhance the data-model and merge logic to support the contextsym being a list of symbols which are merged with a similar set of rules to source record merging.  In particular, it probably makes sense to merge only iff the loc, kind, pretty, sym, and context are all the same.

### We get a bunch of noise in structured records because of "anonymous struct at PATH" and includes via source dir and objdir public header symlinks

For the structured record for sym `"T_1e0231fb6037e57"` with pretty `"mozilla::gfx::Matrix5x4::(anonymous)::(anonymous)"` at loc `"01889:4-10"` in Matrix.h, we see a diff because we end up seeing 2 variations for "anonymous struct":
- `"mozilla::gfx::Matrix5x4::(anonymous)::(anonymous)::(anonymous struct at /builds/worker/checkouts/gecko/gfx/2d/Matrix.h:1889:5)"`
- `"mozilla::gfx::Matrix5x4::(anonymous)::(anonymous)::(anonymous struct at /builds/worker/workspace/obj-build/dist/include/mozilla/gfx/Matrix.h:1889:5)"`

The latter is of course the symlink created by the build process for the public include path and the former is the actual source location.  The situation here is weird because we end up with 1 copy of the structured record for each path for each platform, so our symmetry breaking logic ends up non-deterministic and we inverted which got to be the canonical variation and which got to be the non-canonical variation.

This is largely unrelated to the changes happening here, so this should just end up as a future work enhancement, quite possibly to do something slightly more intelligent in terms of creating the pretty identifier for anonymous namespaces and structs.

The diff for structured records can be hard to read, but if we pass them through `jq .` and diff those, we find that we get differences like

theres-waldo:
> Thanks for doing the dev run, this makes it much easier to dig into the impacts of the changes!

Just a heads up, I did two deploys to dev.searchfox.org, one or Monday with this patch, and a second one on Wednesday for this patch + an additional patch to use clangd's full HeuristicResolver implementation (that's on [this branch](https://github.com/theres-waldo/mozsearch/tree/full-heuristic-resolver)). In the second one a few additional uses are resolved (though not that many, because I haven't yet added new _uses_ of HeuristicResolver in MozsearchIndexer that are needed to make some of the other test cases in bug 1781178 work).

asutherland:
## Heuristics and/or Concrete

>     1. Currently, in `IndexConsumer::VisitCXXDependentScopeMemberExpr()`, I both (1) produce a heuristic result (that's what's new in this patch), **and** (2) record the location for possibly producing concrete results when visiting instantiations (if we have any) during the `AnalyzeDependent` phase (what the code was doing before).
>        I'm wondering, should we add some logic to **only** produce the heuristic result if we didn't end up producing concrete results from the `AnalyzeDependent` phase?
>        I feel like I could go either way on this. On the one hand, the indexer might only know about a subset of the instantiations actually occurring in the project, and the concrete results might therefore miss something that the heuristic results catch. On the other hand, if e.g. we're working with an overload set, it's possible that all instantiations end up using only a subset of the overloads, and the heuristic results would add noise by providing all overloads. In other words, I don't think either approach is _categorically_ more accurate than the other.

Aside: It would be great to have a specific test case that could exercise the permutations being discussed here.

I understand there to be broadly 3 use-cases here:
1. The source record use case where the user is looking at the source listing for a use and would really like a "go to def" option to go to the exact implementation and it's potentially a lot of legwork for the user to figure out which implementation to go to if there are multiple "go to def" options or if there is no "go to def" option and they end up having to use the "search" choice and end up having to pick implementations from there.
2. The source record use case where the user is looking at the source listing for a definition and would really like to be able to get a "search" endpoint listing of the callers of that exact definition.
   - From the source listing, they might then appreciate if the listing included something like the current approach for C++ virtual method overrides where when looking at the specific symbol for a subclass's override, they can click on a link to go to the superclass's symbol for the virtual method, which then lists all of the overrides of that method.
3. The "search" use-case where a user types in a pretty identifier name.  This situation is notable because we expect the user to get a soup of all of the symbols relating to the overload set here because searchfox's "pretty" identifiers never include any ability to distinguish any further than that.  So anything we do will tend to be good enough for this situation.

For the first two cases, it does seem like it would definitely be preferable for our source code rendering code which operates in terms of source records and jumpref lookups to be able to use concrete results if available and not end up broadening that to the overload set.

As discussed in my previous comment, we do have a merge stage for the multi-platform mozilla-central case, but there's nothing stopping us from always running merge logic even in single-platform cases to simplify things.  I could imagine we could do something so that we annotate the overload set source records we emit so that they will get replaced by concrete records when present for the same token with the same pretty.  For now I'll call this a "weak source record", but there are potentially more clever things we could do; more a little below.

### Our clang indexer can't / shouldn't need global awareness

I expect our clang indexer plugin won't always have enough information to make **only** determinations itself[1], so the best course of action is likely to go with **and**, but eventually make sure that we emit sufficient metadata that we can distinguish between the different qualities of "use".  The most straightforward approach to this is for merging (which only happens once all C++ indexing is complete) to have the concrete source records clobber the weak overload set source records, but we could handle things later on in subsequent passes as well if we need the crossref's process global awareness of everything.

While adding something like a "weak" property could work, there are also options where we could further refine the semantics of the overload set "use".  In [Bug 1419018](https://bugzilla.mozilla.org/show_bug.cgi?id=1419018) :kats suggested distinguishing between lhs/rhs for uses and in [Bug 1733217](https://bugzilla.mozilla.org/show_bug.cgi?id=1733217) I dug into that a little more proposing an additional "usage" field.  We could also introduce a new "kind" besides "use" like "potential use" which currently would result in an additional section in the "search" results output, but I think that would probably be a surprising/unpleasant UX since the "use" kind tends to be quite voluminous and people might never get down to "potential use".

That said, similar to your prior enhancement of https://github.com/mozsearch/mozsearch/pull/613 I think it would make sense to land this as-is and then figure out how to best address the "weak" or "usage" use-cases.  What's in this patch is already strictly an amazing step forward for searchfox; the only alternative users have right now is fulltext search which is going to have order of magnitude more false positives.

1: Technically, because of the way we do locking on the analysis files, the clang indexer could actually try and reason about the existing records from other ingestions of the file, but I don't think that we would want to do that because:
- The clang indexer never actually knows when it's going to be the last one to touch that file.
- I think a major performance optimization we could undertake for indexing is to get rid of the file locking and instead have all the clang indexers send their output over a (unix) socket to a rust daemon that can handle de-duplication on the fly.  The current approach of locking the file, loading in the current contents, sorting/merging, then writing the file back out is likely the reason that a build that normally takes ~12 minutes instead takes ~54 minutes.

## HeuristicResolver

>     2. The `HeuristicResolver` facility added in this patch is a slimmed-down version of clangd's (currently much more featureful) facility [of the same name](https://searchfox.org/llvm/rev/be9c91843bab5bb46574c27836bfcd9ad6fc9ef5/clang-tools-extra/clangd/HeuristicResolver.h#45).
>        My current approach is to port over pieces of this facility to Searchfox as we need them to make specific testcases work.
>        However, one could imagine a different approach of just... copying the class from clangd wholesale, as a piece of code which has been battle-tested in a different project? (It doesn't have any dependencies on anything else in clangd, only public clang APIs.)
>        Let me know if you have a preference here, I'm happy to switch approaches.
>        (Granted, even better would be to upstream `HeuristicResolver` to `libClangTooling` or somesuch and expose it as a public clang API, and have both clangd and Searchfox consume it from there. I'd like to explore this, but I'd prefer not to block our progress on Searchfox enhancements on this LLVM change, because LLVM patches can be slow to be reviewed (and then we'd need to adopt the new LLVM version in which it appears and so on.))

I have no problem if you want to copy the classes over now or in the future when you make your next change.  I see the files already have per-file license headers and their Apache 2.0 license is compatible with mozsearch's MPL2.  They also look nicely standalone, so it's not like we'd spend a lot of time trying to disentangle them from other clangd code.  One thing we might do is put them in a "from-clangd" subdirectory with a README that explains what they are.  This could help make it clear to anyone wanting to make changes that they're touching a file that has an upstream without us having to add "HEY DON'T MODIFY THIS FILE" comments every time we copy over a new revision from clangd.  But again, no need to do that now.

I definitely don't think we want to reinvent any wheels clangd has invented, or to make it harder to upstream enhancements made in searchfox to clangd.

asutherland (approved):
As noted in my comments, I think this is amazing as-is and can probably already land (and have the MozsearchIndexer.cpp changes propagated to m-c).  I'm happy to continue discussing potential next steps as it relates to weak source records or "usages", etc. here or on bugzilla.  I did mention some potential spin-offs, but there's no need for you to file anything you're not personally interested in/planning on implementing.  (Like, in particular, the contextsym merging isn't something you need to do or file.)

Thanks for all your work on searchfox!  This is very exciting!

theres-waldo:
> ## Heuristics and/or Concrete
> 
> > 1. Currently, in `IndexConsumer::VisitCXXDependentScopeMemberExpr()`, I both (1) produce a heuristic result (that's what's new in this patch), **and** (2) record the location for possibly producing concrete results when visiting instantiations (if we have any) during the `AnalyzeDependent` phase (what the code was doing before).
> >    I'm wondering, should we add some logic to **only** produce the heuristic result if we didn't end up producing concrete results from the `AnalyzeDependent` phase?
> >    I feel like I could go either way on this. On the one hand, the indexer might only know about a subset of the instantiations actually occurring in the project, and the concrete results might therefore miss something that the heuristic results catch. On the other hand, if e.g. we're working with an overload set, it's possible that all instantiations end up using only a subset of the overloads, and the heuristic results would add noise by providing all overloads. In other words, I don't think either approach is _categorically_ more accurate than the other.
> 
> Aside: It would be great to have a specific test case that could exercise the permutations being discussed here.

Here are a couple of (somewhat contrived) examples to illustrate the two cases:

> On the one hand, the indexer might only know about a subset of the instantiations actually occurring in the project, and the concrete results might therefore miss something that the heuristic results catch.

An example here might be:

```c++
template <typename Surface>
struct DrawingContext {
  void draw(Rectangle);
  void draw(Triangle);
  void draw(Circle);
};

template <typename Surface, typename Shape>
void foo(DrawingContext<Surface>& d, Shape& s) {
  d.draw(s);
}
```

Let's suppose that, for whatever reason, the indexer has only seen instantiations of `foo` with `Shape = Circle` (could be due to imperfections in the indexer implementation in identifying and merging callers from other translation units, or maybe callers with other shapes just haven't been written yet). The concrete results would only include `draw(Circle)`, while the heuristic results would include all three `draw()` overloads. In this case the heuristic results are better because conceptually any of the overloads is a valid target.

> On the other hand, if e.g. we're working with an overload set, it's possible that all instantiations end up using only a subset of the overloads, and the heuristic results would add noise by providing all overloads.

An example here might be:

```c++
template <typename OutputStream>
struct Serializer {
  void serialize(std::string);
  void serialize(int);
  void serialize(URL);
  // bunch of other overloads
  void serialize(FontMetrics);
};

template <typename OutputStream>
void serializeFont(Serializer<OutputStream> s, Font f) {
  // ...
  s.serialize(f.GetMetrics());
}
```

This time, all instantiations use `serialize(FontMetrics)`, and the concrete results will only include that, but the heuristic results will include all overloads most of which are irrelevant.

(Why aren't the heuristics smart enough to figure out the target will always be `serialize(FontMetrics)`, you ask? The answer is that currently we don't attempt to perform overload resolution prior to instantiation. That's something we can consider adding in the future, especially in cases like this where the arguments themselves are not dependent. However, you can easily imagine a slight variation where, say, `Font` has a template parameter too (which `serializeFont()` is also parameterized on), so now the arguments **are** dependent, which is not something the C++ overload resolution algorithm handles (and in particular trying to run clang's implementation of the overload resolution algorithm on dependent inputs will quickly run into asserts and such). We could probably devise and implement a heuristic overload resolution algorithm that handles dependent inputs and gives the expected results in cases like this, but now we're in "small research project" territory.)

---

Not sure if you just wanted to see the test cases to illustrate the tradeoff, or if you meant adding them to the test suite -- I'm happy to do the latter if you'd like.

theres-waldo:
> I could imagine we could do something so that we annotate the overload set source records we emit so that they will get replaced by concrete records when present for the same token with the same pretty. For now I'll call this a "weak source record", but there are potentially more clever things we could do; more a little below.
>
> [...]

Yeah, a "weak" annotation of some sort makes a lot of sense to me. Eclipse CDT uses something similar for e.g. find-references results where it's less sure that a use is definitely referencing the entity you're searching for, and annotates those as a "potential result" in the results list.

I'll file a follow-up for exploring this further.

theres-waldo:
> > 2. The `HeuristicResolver` facility added in this patch is a slimmed-down version of clangd's (currently much more featureful) facility [of the same name](https://searchfox.org/llvm/rev/be9c91843bab5bb46574c27836bfcd9ad6fc9ef5/clang-tools-extra/clangd/HeuristicResolver.h#45).
> >    My current approach is to port over pieces of this facility to Searchfox as we need them to make specific testcases work.
> >    However, one could imagine a different approach of just... copying the class from clangd wholesale, as a piece of code which has been battle-tested in a different project? (It doesn't have any dependencies on anything else in clangd, only public clang APIs.)
> >    Let me know if you have a preference here, I'm happy to switch approaches.
> 
> I have no problem if you want to copy the classes over now or in the future when you make your next change. I see the files already have per-file license headers and their Apache 2.0 license is compatible with mozsearch's MPL2. They also look nicely standalone, so it's not like we'd spend a lot of time trying to disentangle them from other clangd code. One thing we might do is put them in a "from-clangd" subdirectory with a README that explains what they are. This could help make it clear to anyone wanting to make changes that they're touching a file that has an upstream without us having to add "HEY DON'T MODIFY THIS FILE" comments every time we copy over a new revision from clangd. But again, no need to do that now.
> 
> I definitely don't think we want to reinvent any wheels clangd has invented, or to make it harder to upstream enhancements made in searchfox to clangd.

Sounds good! I'll leave this patch as-is, and expand `HeuristicResolver` to be a full copy of clangd's in my next patch.

asutherland:
> > On the one hand, the indexer might only know about a subset of the instantiations actually occurring in the project, and the concrete results might therefore miss something that the heuristic results catch.
> 
> An example here might be:
> 
> ```c++
> template <typename Surface>
> struct DrawingContext {
>   void draw(Rectangle);
>   void draw(Triangle);
>   void draw(Circle);
> };
> 
> template <typename Surface, typename Shape>
> void foo(DrawingContext<Surface>& d, Shape& s) {
>   d.draw(s);
> }
> ```
> 
> Let's suppose that, for whatever reason, the indexer has only seen instantiations of `foo` with `Shape = Circle` (could be due to imperfections in the indexer implementation in identifying and merging callers from other translation units, or maybe callers with other shapes just haven't been written yet). The concrete results would only include `draw(Circle)`, while the heuristic results would include all three `draw()` overloads. In this case the heuristic results are better because conceptually any of the overloads is a valid target.

Thanks for the specific example, this really enhances my ability to discuss this aspect of template indexing!

I agree that the heuristic result's overload set is fundamentally interesting in this case.  I think the concrete uses and the ability to distinguish them would also be useful, both at the point of the use for context menu purposes (source record) and in the search results (target records).

Re: CDT and "potential results", we also see something similar from GitHub and Sourcegraph, although those are more about the indexer; the terms "precise" and "search based" in particular are used to distinguish between "we used something that understands the full semantics of the language" versus "we used tree-sitter so it's not perfect".  As we potentially overhaul our understanding of JS from our current "we assume all JS we ever see is loaded into a single global" to using the more precise scip-typescript in some cases (or all cases, but with a pile of hacks), it could very much make sense for searchfox to grow a concept of "symbol confidence".

Specifically, if implemented, we'd:
- Establish a total ordering for symbol confidence along these lines (which includes cases you may not care about but I'm including for my own future reference) from most confident to least confident:
  - `concrete`: Definitely saw this (use) happen.  Most things in C++ would be concrete and we could default to this.
    - For JS via scip-typescript, exact cross-references generated from that could be concrete, but we'd add distinct confidences for when we add references beyond what scip-typescript provides.
  - `template-overload`: This is part of the template overload set.  Going very specific here to start with because I think this is conceptually unique to templates for C++ so far, at least compared to virtual method overrides where we explicitly do have a symbol with a declaration location.  (Well, non-templated overloads are similar, but arguably if we want to make that overload set more explicit, it should still be distinct from the template overload set for UI purposes.)
  - `js-global`: We added/upgraded this edge because of high confidence about files being loaded into the same global, even though scip-typescript didn't know it.  For example, using the machinery I proposed in https://bugzilla.mozilla.org/show_bug.cgi?id=1775130.
  - `repo-linked`: We're using a heuristic defined in the repository definition to add this edge (potentially replacing some other edge), like an edge from nsIRunnable subclass constructors to their run methods.
  - `js-symbolic`: Symbolic execution was used to glue `js-global` or higher confidence things together.
  - `js-soup-qualified`: This is a guess based on our classic JS analyzer's modeling of everything existing in a single global for qualified symbols.   So for `var Singleton = { aMethod() {} }`, this would correspond to the current `Singleton#aMethod` symbol.  This works well with singletons because they will use the fully qualified name which reduces collisions.
  - `js-soup-flat`: The flattened symbol model of the classic JS analyzer modeling, corresponding to `#aMethod` from the prior example.  This works better than the qualified variant for classes where the type isn't known (which is all the time for the classic JS analyzer).
- Add a `confidence` (maybe shorted to `con`?) attribute on source records that would be a parallel array to `sym` with each confidence entry being associated with the sym array entry at the same index.
- Add a `confidence`/`con` attribute on target records that is singular because symbols in target records are scalar.
- Source record merging would choose higher confidence values over lower confidence values when merging.  There would only ever be a single confidence value for a given symbol in a source record.
- Target records would be subject to merging and would similarly favor higher confidence values over lower confidence values.
- The `output-file.rs`/`format.rs` machinery would propagate the confidence values through to a `data-confidences` attribute on the HTML (so the context menu logic could consult it).
- The context menu would, whenever it sees a token with confidences including "template-overload":
  - Prefix any "go to def" variations with "concrete" or "potential" as appropriate.
    - It might also be necessary to have some overload protection here as it's possible for both the concrete and potential sets to be quite large.  It likely would make sense to not show any "go to def"s if the concrete set is large (>5?), or to show the concrete and suppress the "potential"s if the concrete set is reasonable but the potential set is large (>5).  Numbers could potentially be added to the "Search" variations to clarify the number of symbols in the overload sets.
  - Build a "search" menu variation for both "concrete" and "potential"(?) sets whenever it sees that there is a "template-overload" confidence involved.  The potential set would include the concrete confidence symbols as well, as those would have replaced the "template-overload" confidences.
- The "search" endpoint as exposed by `router.py` could further tuple the qualtype for uses so that "template-overload" would be separate from the normal (concrete) uses.
- The "query" endpoint's faceting could expose the confidence directly as one of its mechanisms for faceting and avoid needing to separate the concrete uses from potential uses.

We don't need to implement this right now, but I always like to have a working destination in mind, and I may try noodling around with it in my hobby time, as it seems quite useful to have an incremental approach for improving our JS implementation incrementally rather than an all-or-nothing transition.

> > On the other hand, if e.g. we're working with an overload set, it's possible that all instantiations end up using only a subset of the overloads, and the heuristic results would add noise by providing all overloads.
> 
> An example here might be:
> 
> ```c++
> template <typename OutputStream>
> struct Serializer {
>   void serialize(std::string);
>   void serialize(int);
>   void serialize(URL);
>   // bunch of other overloads
>   void serialize(FontMetrics);
> };
> 
> template <typename OutputStream>
> void serializeFont(Serializer<OutputStream> s, Font f) {
>   // ...
>   s.serialize(f.GetMetrics());
> }
> ```
> 
> This time, all instantiations use `serialize(FontMetrics)`, and the concrete results will only include that, but the heuristic results will include all overloads most of which are irrelevant.

I think the proposal above could still work for this.
 
> (Why aren't the heuristics smart enough to figure out the target will always be `serialize(FontMetrics)`, you ask? The answer is that currently we don't attempt to perform overload resolution prior to instantiation. That's something we can consider adding in the future, especially in cases like this where the arguments themselves are not dependent. However, you can easily imagine a slight variation where, say, `Font` has a template parameter too (which `serializeFont()` is also parameterized on), so now the arguments **are** dependent, which is not something the C++ overload resolution algorithm handles (and in particular trying to run clang's implementation of the overload resolution algorithm on dependent inputs will quickly run into asserts and such). We could probably devise and implement a heuristic overload resolution algorithm that handles dependent inputs and gives the expected results in cases like this, but now we're in "small research project" territory.)

Yeah, that seems like a lot of work.  I think it's very reasonable to say that ugly/complex code is going to result in an ugly/complex experience.  That said, I think reasonable things searchfox could help do in situations like this are:
- Provide ways to visualize / facet / filter things.
- Letting humans provide superseding graph edges through magic comments or repository definitions which can skip over situations like this where the static control flow graph goes through what effectively is a super node and everything else is dynamic.  For example, the strict control flow graph for our IPC calls would be a mess where everything is connected to everything else.  But we explicitly introduce a magic link from `SendFoo` to `RecvFoo` which hides the complex stuff under the hood.

> Not sure if you just wanted to see the test cases to illustrate the tradeoff, or if you meant adding them to the test suite -- I'm happy to do the latter if you'd like.

I think both are useful, but I think it would be particularly useful to add them to the test suite, as this makes it easier to dig in to what's currently happening.  Also, it means in the future people can just link to the in-(mozsearch-)tree examples rather than have to potentially find this thread or sketch them up from whole cloth.  Also it helps make sure details are filled in.

To do some quick investigation myself, I tried to make your shapes example more concrete in this `template_shapes.cpp` (which you should feel free to further iterate on or ignore, etc.):
```cpp
class GenericSurface {};

class Rectangle {};
class Triangle {};
class Circle {};

template <typename Surface> struct DrawingContext {
  void draw(Rectangle);
  void draw(Triangle);
  void draw(Circle);
};

template <typename Surface, typename Shape>
void foo(DrawingContext<Surface> &d, Shape &s) {
  d.draw(s);
}

int main(void) {
  GenericSurface surface;
  DrawingContext<GenericSurface> context;
  Circle circle;

  foo(context, circle);
}
```

HighCommander4:
> it could very much make sense for searchfox to grow a concept of "symbol confidence"

I filed [bug 1833552](https://bugzilla.mozilla.org/show_bug.cgi?id=1833552) for distinguishing between heuristic and concrete results in the context of templates. Please feel free to transmogrify it into a more general issue about symbol confidence, or to make it depend on such an issue etc.

> currently we don't attempt to perform overload resolution prior to instantiation. That's something we can consider adding in the future

I filed [bug 1833548](https://bugzilla.mozilla.org/show_bug.cgi?id=1833548) about further improvements in this space. I think there is some low-hanging fruit here (simple heuristics like arity) that can be tackled without solving the more general problem of overload resolution with dependent arguments.

> I think it would be particularly useful to add [an example illustrating the behaviour for heuristic vs. concrete results] to the test suite

Will do, I filed [bug 1833529](https://bugzilla.mozilla.org/show_bug.cgi?id=1833529) for this.

theres-waldo:
> it could very much make sense for searchfox to grow a concept of "symbol confidence"

I filed [bug 1833552](https://bugzilla.mozilla.org/show_bug.cgi?id=1833552) for distinguishing between heuristic and concrete results in the context of templates. Please feel free to transmogrify it into a more general issue about symbol confidence, or to make it depend on such an issue etc.

> currently we don't attempt to perform overload resolution prior to instantiation. That's something we can consider adding in the future

I filed [bug 1833548](https://bugzilla.mozilla.org/show_bug.cgi?id=1833548) about further improvements in this space. I think there is some low-hanging fruit here (simple heuristics like arity) that can be tackled without solving the more general problem of overload resolution with dependent arguments.

> I think it would be particularly useful to add [an example illustrating the behaviour for heuristic vs. concrete results] to the test suite

Will do, I filed [bug 1833529](https://bugzilla.mozilla.org/show_bug.cgi?id=1833529) for this.

PR #629, opened 2023-05-15
Bug 1833072 - Use clangd's full HeuristicResolver implementation

theres-waldo:


theres-waldo:
As discussed in https://github.com/mozsearch/mozsearch/pull/628, in this patch I replace the minimal `HeuristicResolver` implementation that was just enough to make the bug 1828223 testcase work, with a copy of clangd's `HeuristicResolver` implemetation.

The testcase added in this patch does not cover every new capability that the full `HeuristicResolver` gives us, it's just an example of the test cases listed in bug 1781178 that is fixed by this change alone. (There won't be a very large variety of test cases fixed by this change alone, because most other test cases also require new calls _into_ `HeuristicResolver` from `MozsearchIndexer`. Those are forthcoming in future patches.)

theres-waldo:
> There won't be a very large variety of test cases fixed by this change alone, because most other test cases also require new calls _into_ `HeuristicResolver` from `MozsearchIndexer`.

Well, the testcase in bug 1833073 is also fixed by this change alone, so I included that as an additional testcase in this PR.

theres-waldo:
(I deployed this branch to dev.searchfox.org, but it shouldn't be materially different from the previous deploy which was also "full heuristic resolver", it just wasn't cleaned up the way these patches are, e.g. it didn't have the separate `from-clang` directory yet)

asutherland (approved):
Awesome!

Thanks for clearly delineating the bits that are clangd!  Also thanks for the care in the patch formulation; while in this case the aggregate diff over the 3 commit wasn't particularly hard to read, it is very nice to be able to drill down into the patches and not have to see the line number noise[1], etc.

1: I think the line number noise is a good trade-off in general in terms of fidelity and realizing when we're changing something, but I should note that we always have the option to add a normalizing step to the pipeline, like [cmd_prod_filter](https://github.com/mozsearch/mozsearch/blob/master/tools/src/cmd_pipeline/cmd_prod_filter.rs) which exists to avoid having the production checks needing to constantly be updated.

PR #633, opened 2023-05-26
Bug 1621789 - Handle DependentScopeDeclRefExpr in MozsearchIndexer

theres-waldo:


theres-waldo:
Couple of notes on this one:

 * It feels slightly brittle to handle different sets of `Decl` types in `VisitCXXDependentScopeMemberExpr` vs. `VisitDependentScopeDeclRefExpr`. (For example, while I can't off the top of my head think of a testcase where a `CXXDependentScopeMemberExpr` would resolve to an `EnumConstantDecl`, I can't confidently rule it out either.) I'm going to make this more robust in a follow-up by factoring out a helper that handles various `Decl` types which various `Visit()` functions delegate to.
 * I haven't added `TemplateStack` logic to `VisitDependentScopeDeclRefExpr`. I can explore this (and unlike `OverloadExpr`, we don't need to wait on [bug 1833552](https://bugzilla.mozilla.org/show_bug.cgi?id=1833552) because for `DependentScopeDeclRefExpr` we may get concrete results that the heuristics failed to turn up), but I'd suggest doing so in a follow-up.

theres-waldo:
Deployed to dev.searchfox.org and confirmed to fix the original testcase in m-c

asutherland (approved):
Woo!

Thanks for calling out both of those, indeed they address questions raised in my mind while checking out the code[1].  They both sound like good potential follow-ups!

1: Although a great thing about you adding test coverage for every enhancement is that it's not really as important that I understand how the C++ sausage is compiled, if you will.

theres-waldo:
> I'm going to make this more robust in a follow-up by factoring out a helper that handles various `Decl` types which various `Visit()` functions delegate to.

Posted #634 for this.

theres-waldo:
> I haven't added `TemplateStack` logic to `VisitDependentScopeDeclRefExpr`. I can explore this (and unlike `OverloadExpr`, we don't need to wait on [bug 1833552](https://bugzilla.mozilla.org/show_bug.cgi?id=1833552) because for `DependentScopeDeclRefExpr` we may get concrete results that the heuristics failed to turn up), but I'd suggest doing so in a follow-up.

Filed [bug 1835692](https://bugzilla.mozilla.org/show_bug.cgi?id=1835692) for this.

PR #634, opened 2023-05-29
Bug 1835694 - Factor out a helper for handling heuristic results in MozsearchIndexer

theres-waldo:


theres-waldo:
I guess I need a bug number for this since it's going to land on m-c.

clang-plugin/MozsearchIndexer.cpp
```diff
@@ -1985,6 +1985,29 @@ class IndexConsumer : public ASTConsumer,
     return true;
   }
 
+  void visitHeuristicResult(SourceLocation Loc, const NamedDecl *ND) {
```
asutherland, line 1988: Not required, but it would be great if you could add a comment here that contextualizes why this method exists and our expectations about whether we expect to have potentially redundant calls to this method.  Here's my understanding from your comment at https://github.com/mozsearch/mozsearch/pull/633#issuecomment-1565082052 plus other context that you should feel free to use whatever pieces of make sense:

Boilerplate-reducing helper for heuristic template results that should be distinguished from concrete results when bug 1833552 is fixed.  We don't expect this method to be intentionally called multiple times for a given (Loc, NamedDecl) pair because our callers should be mututally exclusive expression types.  However, it's fine if this method is called multiple times for a given pair because we explicitly de-duplicate records with an identical string representation (which is a good reason to have this helper, as it ensures identical representations).

asutherland (approved):
Thanks!  I'm very much enjoying the constant stream of enhancements and how nicely factored they are!

PR #635, opened 2023-05-29
Bug 1835696 - Handle DependentNameTypeLoc in MozsearchIndexer

theres-waldo:


theres-waldo:
I don't know how to tell GitHub that this PR depends on #634, such that the list of commits excludes the commit from that PR, but that is my intention.

asutherland:
> I don't know how to tell GitHub that this PR depends on #634, such that the list of commits excludes the commit from that PR, but that is my intention.

This is an unfortunate limitation of GitHub's PR mechanism as I understand it; it doesn't really support stacks.  That said, when making a pull request, you do have control over what the target branch is.  If your branches lived in mozsearch/mozsearch instead of your fork of mozsearch, I think that means you could target this PR at the branch corresponding to https://github.com/mozsearch/mozsearch/pull/634.  But then your branches would have to be in this repo, and that would be confusing.

In general the "simplest" options are either:
1. Land your stack in order and rebase the revisions that aren't semantically part of the PR first.  So you'd land https://github.com/mozsearch/mozsearch/pull/634 and then rebase this PR so it's against the version of that commit that was actually landed on the master branch.  Then force-push to update this PR.
2. When you've got a stack, just use a single PR and make sure to label the individual commits with the specific bug numbers and land them all in a single PR.  I don't think there needs to be an exact 1:1 correspondence between Bugzilla bug and PR, it's fine for multiple bugs to be fixed by one PR.  As a reviewer I'm generally fine with this.  The only exception would be if your stack was really a stack of multiple orthogonal smaller stacks that were touching the same files a bunch.

So, for example, I'd be fine if you had just had this PR and skipped creating 634.  Given that I've already commented on 634, though, it would probably nice to do the 2-step dance from point 1 for this.

asutherland (approved):
Woo woo!  The template train keeps on chugging! :steam_locomotive::train::train::train::train:

theres-waldo:
Deployed to dev.searchfox.org and confirmed to fix the original testcase in m-c

PR #640, opened 2023-07-15
Bug 1843722 - Update DocumentTitler for SYM_INFO

arai-a:
Also support finding a destructor from multiple symbols

### DIFF
--- a/static/js/code-highlighter.js
+++ b/static/js/code-highlighter.js
@@ -111,9 +111,8 @@ var DocumentTitler = new (class DocumentTitler {
    * heuristic, but this should ideally be handled by either:
    * - Directly annotating the DOM with the symbol element that is inducing
    *   the nesting.
-   * - Switching from the SYM_INFO scheme to using a symbols dictionary and
-   *   ensuring the symbol data similarly identifies any nesting associated with
-   *   the symbol.
+   * - Augmenting the SYM_INFO symbol dictionary to provide information about
+   *   nesting in parallel to the "jumps".
    *
    * The current heuristic logic is:
    * - Pick the last observed symbol preceding a `(`.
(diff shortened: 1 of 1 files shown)


static/js/code-highlighter.js
```diff
@@ -126,8 +125,8 @@ var DocumentTitler = new (class DocumentTitler {
       return { long: bestPretty, short: bestShortPretty };
     }
 
-    const symElems = elem.querySelectorAll("[data-i]");
-    scan: for (const symElem of symElems) {
+    const defs = elem.querySelectorAll(".syn_def");
```
asutherland, line 128: Clever constraint refinement here!

I think it would probably be good to add a comment like the following since part of why this works is misleading, feel free to use or edit the comment as you desire:

format.rs maps "def", "decl", and "idl" syntax kinds to "syn_def" so this covers all "kinds" of interest.

static/js/code-highlighter.js
```diff
@@ -139,8 +138,11 @@ var DocumentTitler = new (class DocumentTitler {
       // when we see a colon anywhere.
       let sawClass = false;
       let sawColon = false;
+      // Also check if there's tilde, to distinguish a destructor from its
+      // class symbol.
+      let sawTilde = false;
```
asutherland, line 143: Nice enhancement!

asutherland (approved):
Thank you, thank you, thank you!  I loved this feature and honestly hadn't realized I had broken it before I stepped back from doing searchfox maintenance and dev work[1].  (But I quite happily and confusingly can review this! :)

Note that although I'm suggesting comment changes here I'm going to approve the pull request because they're nice to have and I can always do a PR myself assuming you agree with my suggested comments.

1: I wonder if the resurrection of servo would make it easier to add automated tests for the front-end logic?  Searchfox has surprisingly comprehensive and thorough tests for everything but the front-end now.  That said, there's obviously no need to deal with that in this stack.

asutherland:
As commented at https://github.com/mozsearch/mozsearch/pull/641#issuecomment-1637197060 this (and that PR) are now on https://dev.searchfox.org/ and it works great.  Many thanks!

PR #647, opened 2023-08-17
Bug 1848995 - Restore local modifications to sax-js

staktrace:
It appears our local copy of sax-js had an (undocumented) modification from the original. This restores the modification and records the change more explicitly.


asutherland:
Thanks for both the original sax update and extracting out the delta so future updates become more possible[1]!

1: Knocking on wood and hoping we never find a reason to update this code again, which hopefully having the patch around will guarantee! ;)

PR #650, opened 2023-09-17
Bug 1852805 - Update searchfox SSO workflow docs from maws.

asutherland:
This is basically just translating the canonical docs at https://mozilla-hub.atlassian.net/l/cp/GH1kL2zg into our docs.


asutherland:
The tl;dr is that people with existing credentials will need to install AWS CLI v2 which is not available via PIP, then run `aws configure sso` and take steps to set `AWS_PROFILE` which I suggest adding to the "env/bin/activate" script.  After that, you just run `aws sso login` instead of the weird `$(eval ...)` thing.

Please feel free to push improvements to the docs.

PR #651, opened 2023-09-19
Add support to specify an alias to a repository.

emilio:
This might be useful if we decide to rename /wubkat to /webkit or similar.

### DIFF
--- a/scripts/nginx-setup.py
+++ b/scripts/nginx-setup.py
@@ -162,6 +162,12 @@ for repo in config['trees']:
             # If this fails just leave head_rev as None and skip the optimization
             pass
 
+    alias = tree_config['alias']
+    if alias:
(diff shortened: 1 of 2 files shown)


asutherland (approved):

asutherland (changes requested):
I think [what we ended up doing for oldtree_name with a rewrite-rule](https://searchfox.org/mozsearch/rev/20de4d9dfc3d10d4b4c4996a45e9eb09b433a785/scripts/nginx-setup.py#316) (and the trailing `/`) is probably the way to handle this since a permanent redirect helps migrate people.

Note that the oldtree mechanism doesn't quite accomplish this use-case as-is because it assumes the old hashes need to be mapped, but we could condition those pieces on `oldgit_path` being present or something like that.

PR #652, opened 2023-09-19
rust: Test rust-analyzer rather than save-analysis.

emilio:
This uncovers a regression in generated file handling which is
non-trivial to address, but that I'll look into in the future hopefully.


asutherland (approved):
It definitely makes sense for our tests to actually test!

As noted on matrix, for my divergent scip-indexer.rs on my hobby stack I did end up maintaining the existing symbol scheme for rust (I think), so it probably would be good for you to form an opinion on whether you prefer these new symbols or the original rust-analyzer symbols you were previously generating.

emilio:
I don't care particularly about the symbols, fwiw, happy to go either way :)

PR #654, opened 2023-09-22
Document use of docker, add support for "trees" repo, provision scip-typescript and scip-python

asutherland:
This is slightly more of my hobby-stack touching the docker scripts than I was expecting when I said what I was going to do in https://github.com/mozsearch/mozsearch/pull/653 but I think this is probably a reasonable subset and gets all the re-provisioning out of the way at once.  I have locally re-provisioned with `./build-docker.sh` with these changes and it just ran fine.  I will perform automated re-provisioning of the AMIs if/when this PR gets r+'ed to land and I land it.

The most helpful change to the docker scripts is that it defines a persistent volume that gets mounted at the "vagrant" user's home directory.  Without this change, anything that happens in the docker container is entirely ephemeral and evaporates when the canonical run-docker.sh session exists.  This didn't matter too much for "make build-test-repo", but does start to matter if you've done "make build-mozilla-repo" and would like to not have to redo that every time.  This is what https://bugzilla.mozilla.org/show_bug.cgi?id=1812838#c1 was talking about.

The docker docs are mainly an attempt to update my WSL2 notes at https://docs.google.com/document/d/10uOwXBENgHzTAyOogSKu6cdxl-T6yQ7ars1CLYUJe2k/edit that I'd been pointing people at.  I've also added some handwaving for macOS/OS X.  (People for whom Vagrant is already working it should still keep working for them.)

This also includes mechanics for have a "trees" repo and a corresponding "tree-configs" dir so that one can run `make build-trees`.  I used this to experiment with scip-typescript on the m-c "devtools" subtree and my "gaia-email-libs-and-more" fork (via symlink) and scip-python on the m-c "python" subtree without having to deal with all of m-c and the indexing time that takes.  It's not something I would particularly point other people at right now, but since it involved touching the docker scripts in order to support some of these files being symlinks, it seemed simplest to just land it.

As part of the "trees" commit, and therefore included here, this installs "rtx" to help install node.js and scip-typescript and scip-python.  Relatedly, the other day I was interested in seeing if I could use "scip-ruby" to index "mastodon", but scip-ruby cannot be installed the same way and scip-ruby can't easily index mastodon because it's a monorepo or something like that.  (There's an issue in the scip-ruby repo about this, but I don't want to xref it here because that would be confusing for that issue.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ repos
 /config
 addons
 /tests/tests/checks/explanations
+/trees/*
+!/trees/README.md
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ addons
 /tests/tests/checks/explanations
 /trees/*
 !/trees/README.md
+/tree-configs/*
+!/tree-configs/README.md
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,18 @@ build-mozilla-repo: check-in-vagrant build-clang-plugin build-rust-tools
 	/vagrant/infrastructure/web-server-setup.sh ~/mozilla-config config1.json ~/mozilla-index ~
 	/vagrant/infrastructure/web-server-run.sh ~/mozilla-config ~/mozilla-index ~
 
+build-trees: check-in-vagrant build-clang-plugin build-rust-tools
+	mkdir -p ~/trees-index
+	/vagrant/infrastructure/indexer-setup.sh /vagrant/tree-configs config.json ~/trees-index
+	/vagrant/infrastructure/indexer-run.sh /vagrant/tree-configs ~/trees-index
+	/vagrant/infrastructure/web-server-setup.sh /vagrant/tree-configs config.json ~/trees-index ~
+	/vagrant/infrastructure/web-server-run.sh /vagrant/tree-configs ~/trees-index ~ WAIT
+	/vagrant/infrastructure/web-server-check.sh /vagrant/tree-configs ~/trees-index "http://localhost/"
+
+serve-trees: check-in-vagrant build-clang-plugin build-rust-tools
+	/vagrant/infrastructure/web-server-setup.sh /vagrant/tree-configs config.json ~/trees-index ~
+	/vagrant/infrastructure/web-server-run.sh /vagrant/tree-configs ~/trees-index ~ WAIT
+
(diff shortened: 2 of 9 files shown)


asutherland:
@staktrace You may also want to take a look at this; my vague recollection is that you were on macOS/OS X at least some of the time, but my memory is infamously bad.  I'd definitely be interested to know if you still maintain a vagrant checkout and whether you want it to keep working, etc.  (Noting again that I don't think I've done anything to break vagrant, but I also have stopped testing on vagrant, although I think I did validate things still worked back when I landed the original docker changes.)

staktrace:
Thanks for the heads up! I do work on macOS, but I haven't used the vagrant setup in a while now. I don't really care if it breaks; if I need to spin it up I can use docker.

asutherland:
An interesting complication I just ran into around the addition of the persistent volume is that because we build and extract so many things into the vagrant user's home directory as part of the provisioning process (ex: rust-analyzer), the persistent volume really needs to be destroyed and rebuilt each time `build-docker.sh` is run, otherwise we aren't actually updating everything.

Specifically, on first "mount", the persistent volume copies what already exists in the docker image where it is being mounted, but will not update to reflect changes on future mounts.  So I just experienced having a very old version of "rust-analyzer" because we symlink from /usr/local/bin to the rust-analyzer in the home directory.

Other options beyond destroying the persistent volume would be to have the installed things like rust-analyzer be in a path not covered by the persistent volume.  But I think in practice, it's probably better to just destroy the volume.  This is more consistent with what we did with vagrant and reduces the odds that what a developer thinks is happening diverges from what is actually happening.  It does mean when I run `make build-mozilla-repo` which I've been doing a lot these days, I have to re-download some stuff, but there are upsides to that anyways.

One question is whether `build-docker.sh` should recognize that there's already something there and be like "hey, run `destroy-docker.sh` first, then run this again".

asutherland:
I've updated build-docker.sh in the "Support use of a volume mount for persistence" commit to handle the need to clean up the volume.  We don't do any prompting and I didn't create a "destroy-docker.sh" command.

I also have now properly guarded that cleanup logic so it only tries to run if the container and/or volume exist.

asutherland:
We'll potentially be pointing external parties at this soon and I want to make sure any experimentation they perform is using our docker best practice and I'm experiencing some branch juggling complications, so I'm going to land this now but will treat this as post-landing review and respond to action items arising from it.

PR #655, opened 2023-09-25
Bug 1855092 - Provision emoji font on web-server and indexer

asutherland:


asutherland:
Going to land and provision this immediately since it's extremely low risk to the indexers but handy for my hobby stack for class-diagram stuff that's useful to people beyond me.

PR #656, opened 2023-09-29
Bug 1812838 - Elide comments from provisioning user data

asutherland:
There's a limit on the data we send when triggering the indexer.  Right now we send a concatenation of our various provisioning scripts (wrapped inside a shell script HERE directive).  My recent other commit changes pushed us over this limit, so my pragmatic fix is to use a regexp to elide comment lines, avoiding removing hashbang lines which can be load bearing.  We leave single newlines for whitespace, compressing runs of newlines.

A `--dry-run` command with hacky parsing was added to aid in validating the changes.

Provisioning successfully completed with these changes and indexer runs then successfully completed on those provisioned jobs, so landing.

### DIFF
--- a/infrastructure/aws/trigger-provision.py
+++ b/infrastructure/aws/trigger-provision.py
@@ -9,11 +9,18 @@
 from __future__ import absolute_import
 import boto3
 from datetime import datetime, timedelta
+import re
 import sys
 import os.path
 
-kind = sys.argv[1]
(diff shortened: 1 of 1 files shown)


asutherland:
This fixes the problem I noted in https://github.com/mozsearch/mozsearch/pull/654#issuecomment-1739437660

PR #659, opened 2023-10-16
Provision scip CLI 0.3.0 to get `scip print`

asutherland:
`scip print` is extremely useful for debugging / understanding what's going on with SCIP indices.  Note that it can take some time to run on large indices and the output will be voluminous, so you'll always want to pipe to a pager.

Originally we were updating to 0.2.3 but 0.3.0 makes the print command also support JSON output via `--json` which seems potentially useful.

### DIFF
--- a/infrastructure/common-provision-pre.sh
+++ b/infrastructure/common-provision-pre.sh
@@ -176,7 +176,7 @@ if [ ! -d git-cinnabar ]; then
 fi
 
 # Install scip
-SCIP_VERSION=v0.2.0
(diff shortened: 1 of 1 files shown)


asutherland:
This re-provisioned fine via `./build-docker.sh` so I'm going to land.

The standard disclaimer is that just doing `./build-docker.sh` will reuse any existing provisioning layers that didn't have their content hash changed of the script or their args, so it's not 100% the same as a full reprovision from scratch.

PR #660, opened 2023-10-16
Avoid using stale XPIDL libs.

asutherland:
When running `make build-mozilla-repo` from a persistent VM volume, it's necessary to be ready to delete the XPIDL libs and re-extract them to avoid being inconsistent with the tree.


asutherland:
I did not do a great job in getting everything in the "make build-mozilla-repo" super-happy PR before, but on the upside I'm not spending a lot of time re-summarizing commit messages for the PR summary!  Landing this because this did not break `make build-test-repo` and I regularly have been running with this for `make build-mozilla-repo` too.

PR #663, opened 2023-10-25
 Bug 1859373  - Refactor rust-indexer's SCIP logic into a more general scip-indexer with integrated use of tree-sitter for nesting and context/contextsym

asutherland:
This is the contents of my hobby stack relating to scip-indexer.rs with the following notable changes:
- I already landed a bunch of somewhat orthogonal changes ahead of time to avoid them cluttering up this stack.
- In my hobby-stack I had switched over from js-analyze to the scip-typescript SCIP indexing.  That's no good, so that's not included here.  As noted in a commit, the most pragmatic approach might be to add a make target like `make build-jstest-repo` to run the test repo in an alternate configuration that uses scip-typescript instead of js-analyze.  I've also proposed elsewhere performing a hybrid union of everything, but that would still be a subsequent step because the scip-typescript output just isn't useful for our purposes yet.
- This does not include changes from later in the stack where I add new functionality for C++ indexing and/or cross-referencing that added new fields and accordingly the struct instantiations in scip-indexer needed to be updated.  In general they were all empty values anyways that were trivially added, so there's no problem if scip-indexer.rs changes in ways that completely bit-rot my hobby-stack.


asutherland:
I've triggered a "dev" channel run of config1 using this branch (and the corresponding mozsearch-mozilla branch).  There was no pre-existing "dev" channel server, so if dev.searchfox.org responds, it works.  If it doesn't respond, the run failed and I'll check it out tomorrow.

In general, everything related to rust should look just like what's happening on asuth.searchfox.org.

asutherland:
The run seems to have completed successfully, https://dev.searchfox.org/mozilla-central/source/ works.

An interesting example via :mccr8 is [dev.searchfox.org search for LayoutPixel](https://dev.searchfox.org/mozilla-central/search?path=&q=LayoutPixel) versus the [stock searchfox.org search for LayoutPixel](https://searchfox.org/mozilla-central/search?q=LayoutPixel&path=) where LayoutPixel is defined as:

```rust
pub struct LayoutPixel;
```

asutherland:
The run did result in some warnings that all seem to uniformly suggest that the processing pipeline is getting confused by the use of `!` in rust macro identifier names:
```
+ /home/ubuntu/mozsearch/tools/target/release/scip-indexer /mnt/index-scratch/config.json mozilla-central --subtree-name rust --subtree-root . --platform android-armv7 objdir-android-armv7/rust.scip
[2023-10-25T06:07:38Z INFO  scip_indexer] Processing symbols/definitions for 'xpcom/rust/xpcom/xpcom_macros/src/lib.rs'
[2023-10-25T06:07:38Z WARN  scip_indexer] Unable to parse SCIP symbol: rust-analyzer cargo lazy_static 1.4.0 lazy_static!
    :Err(MissingDescriptor)
[2023-10-25T06:07:38Z WARN  scip_indexer] Unable to parse SCIP symbol: rust-analyzer cargo quote 1.0.28 format/format_ident!
    :Err(MissingDescriptor)
[2023-10-25T06:07:38Z WARN  scip_indexer] Unable to parse SCIP symbol: rust-analyzer cargo quote 1.0.28 quote!
    :Err(MissingDescriptor)
[2023-10-25T06:07:38Z WARN  scip_indexer] Unable to parse SCIP symbol: rust-analyzer cargo syn 2.0.18 token/Token!
    :Err(MissingDescriptor)
[2023-10-25T06:07:38Z WARN  scip_indexer] Unable to parse SCIP symbol: rust-analyzer cargo xpcom_macros 0.1.0 bail!
    :Err(MissingDescriptor)
[2023-10-25T06:07:38Z WARN  scip_indexer] Unable to parse SCIP symbol: rust-analyzer cargo core https://github.com/rust-lang/rust/library/core builtin/macros/derive!
```

It's conceivable that may be a failure of rust-analyzer to escape or otherwise normalize the `!`, in which case I'll need to either explicitly downgrade just this class of warning or just downgrade them all from warning.  Right now I'd lean towards downgrading everything because I don't expect to actively triage any new warnings that come from scip-indexer at this time.  (Although maybe someone else would?)

tools/src/bin/scip-indexer.rs
```diff
@@ -0,0 +1,1222 @@
...
+        //     - for rust that does give us size and alignment which is nice
+        //     - for JS we get the inferred type as well, plus the extracted
+        //       comment as a second string after the type info.
+
+        for scip_sym_info in &doc.symbols {
+            // Process each symbol to:
+            // - Derive a canonical mozsearch symbol name and map it.
+            // - Derive structured analysis information that we will emit in the
+            //   next pass if/when we see the definition.
+            if let Ok(scip_sym) = scip::symbol::parse_symbol(&scip_sym_info.symbol) {
```
emilio, line 438: What do you think of factoring this out / deindent this a little bit?

Something like:

```
let Ok(scip_sym) = scip::symbol::parse_symbol(..) else {
    warn!(...);
    continue;
};

// ### Extract M...
```

tools/src/file_format/config.rs
```diff
@@ -79,6 +79,26 @@ pub struct TreeConfigPaths {
...
+    /// Definitions of SCIP-based indexes to ingest.  Currently it's expected
+    /// that the build script will handle downloading or generating the indexes.
+    #[serde(default)]
+    pub scip_subtrees: BTreeMap<String, ScipSubtreeConfig>,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct ScipSubtreeConfig {
+    /// The path the SCIP index can be found at.
+    pub scip_index_path: String,
```
emilio, line 91: I'm not sure if we have a precedent for this kind of indexer info inside the tree config. It just feels like something that should probably be a build output perhaps? Anyways not opposed to it, it works.

emilio (approved):
Looks great!

tools/src/bin/scip-indexer.rs
asutherland: Oh wow, I'm not sure I realized [let-else](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) existed!  This is fantastic!  I'll definitely adapt this here and potentially there are some other spots where I currently use a match-based idiom for the same purpose that could be made more concise with let-else,

Thanks!

tools/src/file_format/config.rs
asutherland: Yeah, I'm not 100% on this either.  I think a lot of my decision here was just "having these defined in JSON could be handy and this is a JSON file we already have and use and where we already do the environment variable translations".  A per-tree JSON file or even something like an /etc-style `scip-trees.d` structure where we allow the build system to just dump JSON blobs in a directory and then we union them up together with jq might work much better.  (And the directory could also be pre-populated from the tree's config dir for statically known things.)

I removed it from the stack because we're not using scip-typescript, but I guess I already have a motivating use case for this in the hunk I removed:

```bash
# Find all the JS islands identified by "package.json"; if they have a jsconfig.json
# we would not want to `--infer-tsconfig`, but currently these do not exist in the
# m-c tree, so we don't need to care.
#
# Our find command prints out just the dirname where we found the "package.json",
# which will be "." for the root and which we will want to ignore, so we use grep
# for that.  (Specifically, the package.json at the root of m-c is useless for
# our purposes, it just expresses devDependencies and we do not want
# typescript to treat the entire tree as a single JS project root because it
# categorically is not.)
JS_SCIP_ROOT=$INDEX_ROOT/js-scips
SCRATCH_CONFIG_FILE=$WORKING/config-scratch.json
mkdir -p $JS_SCIP_ROOT
pushd $GIT_ROOT
find * -name package.json -printf '%h\n' | grep -v '^\.$' | while read -r dir; do
    # Use the path as the basis for our name for this index.  "foo/bar/baz" will
    # become "foo__bar__baz".
    SCIP_KEY=${dir//\//__}
    SCIP_FILE=${JS_SCIP_ROOT}/${SCIP_KEY}.scip
    pushd $dir
    # Run the indexer and add an entry to our config file if it succeeds.
    rtx exec nodejs@18 -- scip-typescript index --output ${SCIP_FILE} --infer-tsconfig && \
      jq ".trees[\"mozilla-central\"][\"scip_subtrees\"][\"$SCIP_KEY\"] = { \"scip_index_path\": \"$SCIP_FILE\", \"subtree_root\": \"$dir\"}" $CONFIG_FILE > $SCRATCH_CONFIG_FILE && \
      mv $SCRATCH_CONFIG_FILE $CONFIG_FILE || true
    popd
done
popd
```

The horrible bash there is iteratively updating the environment-evaluated config file we store in the index root in order to cram additional entries into `scip_subtrees` iteratively.


The initial rationale for introducing this mechanism (and before the above bash nightmare):
1. We use the name of the key for this record in `scip_subtrees` above as a human chosen identifier that we include in the symbols we generate to concisely namespace them.
2. I started out without this and it felt like it was going to result in a lot of extra bash hassles in terms of propagating the values around and argument passing, and that there was potential where we might have scip-indexer directly understand what to do from the config file.  I didn't move forward on scip-indexer doing that because the existing per-platform m-c dataflow really favors the existing approach.

asutherland: I'd definitely be interested if you think an `outputs.d` mechanism would make sense and/or what other data you might see us including in such a self-descriptive build byproducts mechanism.

That said, I think I might favor just leaving this setup as-is for this landing with the understand we will replace it.  Noting that in this stack I somehow broke the indexing of `python/` in the tree that works on asuth.searchfox.org, so maybe I can change it up while addressing that.

tools/src/bin/scip-indexer.rs
```diff
@@ -926,21 +944,72 @@ fn analyze_using_scip(
...
+                        scip_symbol_to_structured.insert(norm_scip_sym.clone(), fake);
+                        our_symbol_to_scip_sym.insert(norm_scip_sym_ustr, norm_scip_sym.clone());
+                        scip_symbol_to_structured.get(&norm_scip_sym).unwrap()
+                    } else {
+                        warn!(
+                            "Unable to find structured data for symbol: {}",
+                            norm_scip_sym
+                        );
+                        continue;
+                    }
```
nicolas-guichard, line 1038: > At least for scip-typescript, it's very common to not have symbol defs for
> locals.  We now create synthetic structured records for this case that aren't
> really useful at all, but at least provide hover highlighting that's okay.

I also see this happen a lot with scip-java, not only for locals, but also for packages and types imported from dependencies.
For instance [this file](https://github.com/nicolas-guichard/mozsearch/blob/scip-java/tests/tests/files/java_project/lib/src/main/java/java_project/Library.java) and [this file](https://github.com/nicolas-guichard/mozsearch/blob/scip-java/tests/tests/files/java_project/lib/src/test/java/java_project/LibraryTest.java) are both part of the java_project package, and declare that with a `package java_project;`. scip-java lists both of those as occurrences of the symbol `semanticdb maven . . java_project/` but never actually provides a matching `SymbolInformation`.
In the [test file](https://github.com/nicolas-guichard/mozsearch/blob/scip-java/tests/tests/files/java_project/lib/src/test/java/java_project/LibraryTest.java) , this also applies to all external symbols (`org/`, `org/junit/`, ..., `org/junit/jupiter/api/Assertions#assertTrue(+4).`.

Would it be an issue to provide fake symbols in every cases, and not only for locals?

scripts/rust-analyze.sh
```diff
@@ -54,6 +56,8 @@ if [ -d "$RUST_ANALYSIS_IN" ]; then
   fi
 
   $MOZSEARCH_PATH/tools/target/release/rust-indexer \
+    "$CONFIG_FILE" \
+    "$TREE_NAME" \
```
nicolas-guichard, line 60: rust-indexer hasn't changed, so it doesn't expect those arguments.
```suggestion
```

tests/searchfox-config.json
```diff
@@ -15,7 +15,8 @@
       "git_path": "$MOZSEARCH_PATH",
       "git_blame_path": "$WORKING/searchfox/blame",
       "codesearch_path": "$WORKING/searchfox/livegrep.idx",
-      "codesearch_port": 8081
+      "codesearch_port": 8081,
+      "scip_subtrees": {}
```
nicolas-guichard, line 19: To have the Rust indexing back in searchfox too:
```suggestion
      "scip_subtrees": {
        "rust": {
          "scip_index_path": "$WORKING/searchfox/objdir/tools/rust.scip",
          "subtree_root": "tools"
        }
      }
```

nicolas-guichard (commented):
Hi!

This looks good to me.

Is the plan is to strip the SCIP-related bits from rust-analyze.sh and rust-indexer.rs? Or to remove them altogether and only keep the SCIP path long-term?

I've rebased my scip-java-based tree on top of this and it seems to work as expected.

tools/src/file_format/config.rs
emilio: Yeah, leaving as is for now sounds fine to me.

Some more descriptive build output kind of thing sounds worth it. But we'd need to teach each build script to generate it...

asutherland: > Some more descriptive build output kind of thing sounds worth it. But we'd need to teach each build script to generate it...

We could provide wrapper scripts or a wrapper-driver for `scip-typescript`, etc. that help.

That said, the current configs just approximate the same information that could be found by naming the file `SCIP_SUBTREE_NAME.scip` and placing it at the ~root path of where we indexed which provides us with `subtree_root`.  The most natural place for those files to end up is unfortunately in the source tree (and I think the scip-foo binaries already like to do this and can get mad if you try and point the output elsewhere), and we don't zip that up because we already know the source tree contents, so we could also have a `find` script over the source tree for `*.scip` that derives the info for us (assuming that the index names are already appropriate).

But this seems like something we can iterate on based on as part of @nicolas-guichard's work.

asutherland:
> Is the plan is to strip the SCIP-related bits from rust-analyze.sh and rust-indexer.rs? Or to remove them altogether and only keep the SCIP path long-term?

After ensuring there's no regressions that we need to roll back from, I think we would remove them and keep the SCIP path long term.

The rationale is:
- The table on [the root page of searchfox.org](https://searchfox.org/) lists ESR115 and ESR102 our oldest semantically indexed branches for rust.
- Looking at [the artifacts for an ESR115 searchfox task we do have a SCIP artifact](https://treeherder.mozilla.org/jobs?repo=mozilla-esr115&searchStr=searchfox&selectedTaskRun=dn4jbd0rR1iYANMrh1nTLw.0).  Note that one must manually click on "Artifacts and Debugging Tools" tab in the header of the bottom panel on the UI.
- Looking at [the artifacts for an ESR102 searchfox task we do NOT have a SCIP artifact](https://treeherder.mozilla.org/jobs?repo=mozilla-esr102&searchStr=searchfox&selectedTaskRun=JHiSZ28DTGKwlH65_7pLww.0) (and this makes sense), but per [the tracking page for ESR releases](https://whattrainisitnow.com/release/?version=esr) we can see ESR102 is no longer supported and the [calendar which has an expandable "click to toggle past releases table data"](https://whattrainisitnow.com/calendar/) shows ESR 102 is far enough in the past that we'll probably lose our build jobs for it soon anyways.  Ah, and yeah, https://bugzilla.mozilla.org/show_bug.cgi?id=1854084 says the jobs have already been pulled and indeed our [last job in treeherder](https://treeherder.mozilla.org/jobs?repo=mozilla-esr102&searchStr=searchfox&selectedTaskRun=CwlSzoiVQqyDbHGCYwBj3A.0) is from Sep 26th which means we're just coasting on the the extant artifacts, but as they expire esr102 will start erroring out.  I'll file a bug to remove the esr102 semantic support.
 
> I've rebased my scip-java-based tree on top of this and it seems to work as expected.

Awesome!

scripts/rust-analyze.sh
asutherland: Good eye!  I had a patch later in my stack with the comment "Fold this: Undo erroneous additional args to rust-analyze.sh" that I missed!  (I'm pretty sure this was a situation where I made the changes in the wrong editor buffer and only noticed it later when I went to run the searchfox repo tests.)

tools/src/bin/scip-indexer.rs
asutherland: Amending this, to avoid causing bit-rot for @nicolas-guichard I'm not going to do this right now, but will be very happy if @nicolas-guichard wants to make changes like this.

tools/src/bin/scip-indexer.rs
asutherland: > Would it be an issue to provide fake symbols in every cases, and not only for locals?

Not an issue.  We should definitely be emitting source and target records in these cases and my choice to have us `continue` here I think was probably a bit stop-gap; I expect if I'd tried to continue to improve the JS use-case that I would have changed this.  I believe I may have been under the (mistaken) impression that the SCIP file should be providing definitions in all cases, and this likely informed my attempt to assume we should always have a "structured" record.  (And thereby avoid me needing to handle the case where a structured record would not exist.)

It's even okay if we emit structured records that we infer in the analysis files that we produce.  To ensure we only emit the record once and are able to merge/de-duplicate it, we have only emitted the structured records at the type's point of definition.  The scip-indexer has things easier than `MozsearchIndexer.cpp` that operates in parallel, though, so it could arbitrarily pick a file to put structured records in as long as it can be consistent about the choice of file and "loc" so that `merge_files` in `merger.rs` can merge things consistently.  Maybe we could introduce a synthetic file like `__package__` or similar in order to hold inferred structured records whose scope is not inherently limited to a single file.  (I could believe we might end up wanting to emit structured records for the local symbols too.)

The Java "package" situation seems quite analogous to the C++ namespace situation where there's no single point of definition.  However, as I look at our handling of C++ namespaces right now, I see we [emit the namespaces as a "def"](https://searchfox.org/mozilla-central/rev/a1827d62f97d9721993ba489f4af7dea422d418f/build/clang-plugin/mozsearch-plugin/MozsearchIndexer.cpp#1670-1671) which feels wrong.  We should probably weaken that to a "decl" as a first step and maybe even introduce a specific kind like I did with "forward" for C++ forwards.  That said, there's no urgency to change the C++ namespaces to "decl" right now since it doesn't really affect any user functionality.

asutherland:
I'm going to land this and the mozsearch-mozilla changes now; there are 2 outstanding issues I will try and quickly prepare and land fixes for, but since that might not get done tonight, I would like to land now to make sure I'm not blocking @nicolas-guichard's work.  Those issues are:
- The warning issue with `!` in macros I cite above in https://github.com/mozsearch/mozsearch/pull/663#issuecomment-1779660640
- Actually consuming the python data.

Deployment-wise, I am going to:
- Move the spare release1 server to the dev channel to keep it as a backup that is safe from me triggering an indexer run near the utc22 run.
- Trigger an indexing run.

asutherland:
I addressed the macro warning in https://github.com/mozsearch/mozsearch/pull/664 tracked as https://bugzilla.mozilla.org/show_bug.cgi?id=1862092 by upgrading our scip crate from my 0.2.x fork to the stock 0.3.1.  There seem to be a number of notable semantic changes that may end up being useful assuming scip-java is emitting the new fields.

PR #665, opened 2023-11-06
Podman support and Nix flake

nicolas-guichard:
As requested by @asutherland in https://bugzilla.mozilla.org/show_bug.cgi?id=1490144#c7 I'm sharing here my Podman/Nix-based setup:

> For your stack, it's great that you're adding podman and nix support! For nix, I'm not really familiar with nix, although I understand it's popular and something I should be more familiar with! It would be great to split those out into their own pull request(s), and in particular since I'm not familiar with nix, it would be great if you could make sure to[1]:
> 
> - Provide brief context about what the .envrc file is in the commit message. (In particular, since it's a fairly generic dotfile name, it'd be good for the history for the file to help provide that context.)
> - Add some additional context to the docs about what nix is and when people would want to use it. It might make sense to split this out into a separate markdown file under docs/ with a pointer. Specifically, I think a lot of existing Firefox contributors may be in a similar situation as me where we could benefit from some pointers about nix and how using it might be useful with searchfox and whether one can usefully use it on Ubuntu/Fedora/etc. or whether one would really only want to use it if switching to nix OS.

I didn't add too much to the docs about Nix, only notes in the places where it is useful. If I were duplicating the Docker provisioning logic into Nix then more would be needed, but at the moment it's just a convenient way (for me at least) to install podman, awscli2, boto3 and rich. The commit message has more details.
For a pointer, I think [zero-to-nix.com](https://zero-to-nix.com) is great to start with.
You can enjoy most of the features of Nix without using NixOS. I'd even recommend using Nix on another distro to begin with: it's better to be a bit familiar with the tool before having the world depend on it.

[Direnv](https://direnv.net/) is a per-directory development environment manager.
Basically using Nix and Direnv together enables this:
```
➜  Mozilla > aws
aws: command not found

➜  Mozilla > cd mozsearch
direnv: loading ~/Mozilla/mozsearch/.envrc
direnv: using flake
direnv: nix-direnv: using cached dev shell
TL;DR:
- './build-docker.sh' to build the container
- './run-docker.sh' to enter the container
- 'cd /vagrant; make build-test-repo' inside to build
- open http://localhost:16995 in a browser
direnv: export […] +AWS_PROFILE […] ~PATH […]

➜  Mozilla/mozsearch > aws

usage: aws [options] <command> <subcommand> [<subcommand> ...] [parameters]
To see help text, you can run:

  aws help
  aws <command> help
  aws <command> <subcommand> help

aws: error: the following arguments are required: command

➜  Mozilla/mozsearch > cd ..
direnv: unloading

➜  Mozilla > aws  
aws: command not found
```

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,3 @@
+# Configuration file for direnv.net
+# Uses Nix to provide podman, awscli and a couple of Python dependencies
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ addons
 !/trees/README.md
 /tree-configs/*
 !/tree-configs/README.md
+.direnv
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ that:
   This makes sure that the vagrant user in the container can read/write inside
   the bind mount at /vagrant.
 
+> Note for Nix users: the devShell in flake.nix provides both of those requirements.
+
 #### Installing on macOS/OS X
 
 I don't think anyone has tried this yet, but it seems like there are a variety
--- a/README.md
+++ b/README.md
@@ -35,14 +35,24 @@ If you do want to use it and you work at Mozilla, you should acquire a license
 through Service Desk before doing anything else.  If you already have a license,
 that's fine too.
 
+#### Alternative to Docker: Podman
+
+Podman can be used almost as a drop-in replacement to Docker. Just make sure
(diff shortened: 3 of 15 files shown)


asutherland (approved):
Thanks for the context, it's appreciated!  I'll land this now, then.

PR #667, opened 2023-11-06
Java and Kotlin support via scip-java

nicolas-guichard:
Up on dev.searchfox.org for testing already, see eg. [this search for org.mozilla.geckoview.GeckoSession.PromptDelegate.AutocompleteRequest](https://dev.searchfox.org/mozilla-central/search?q=symbol:S_java_java_org%2Fmozilla%2Fgeckoview%2FGeckoSession%23PromptDelegate%23AutocompleteRequest%23) which returns Java and Kotlin files.

I already see an issue with tree-sitter nesting for Kotlin in the link above, it apparently doesn't like anonymous classes.

Some more testing is also probably needed, I only added to the snapshot what was needed to build the tests repo.
I'd like to replace the java/kotlin tests with something that mixes the two languages.


asutherland:
For lines with fallback symbols like https://dev.searchfox.org/mozilla-central/rev/6e28819cef5e12a11316da483baa8a3fe9843dce/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/GeckoSession.java#16
```java
import android.graphics.Bitmap;
```

The symbol arising for `Bitmap` seems to be `semanticdb maven . . android/graphics/Bitmap#` which looks like it might be the raw SCIP symbol maybe due to the fallback logic involved.  The most important thing I think is for us to make sure there is no whitespace in the symbol identifier[1], but that should also ideally follow our symbol scheme.

1: I think I've fixed a bug about how we parse the `identifiers` file so this may be less of a problem than it used to be, but I'm also not 100% sure I landed that on trunk yet, it may only be on my hobby-branch.

nicolas-guichard:
> The symbol arising for `Bitmap` seems to be `semanticdb maven . . android/graphics/Bitmap#` which looks like it might be the raw SCIP symbol maybe due to the fallback logic involved. The most important thing I think is for us to make sure there is no whitespace in the symbol identifier[1], but that should also ideally follow our symbol scheme.

I've updated 9e34d5e to re-use the logic that splits an SCIP symbol into parts for those symbols too. So `semanticdb maven . . android/graphics/Bitmap#` should now gets transformed into `S_java_java_android/graphics/Bitmap#`.

I've also added a commit to handle the literal objects that we encounter in the m-c tests (eg. at [AutocompleteTest.kt#46](https://dev.searchfox.org/mozilla-central/source/mobile/android/geckoview/src/androidTest/java/org/mozilla/geckoview/test/AutocompleteTest.kt#46)) where we have methods that appear as locals.

tools/Cargo.toml
```diff
@@ -61,6 +61,8 @@ tree-sitter = "0.20.9"
 tree-sitter-python = "0.20.2"
 tree-sitter-rust = "0.20.3"
 tree-sitter-typescript = "0.20.2"
+tree-sitter-java = "0.20.2"
+tree-sitter-kotlin = { git = "https://github.com/nicolas-guichard/tree-sitter-kotlin.git", branch = "field-names" }
```
nicolas-guichard, line 65: TODO: replace with https://github.com/fwcd/tree-sitter-kotlin branch=main once fwcd/tree-sitter-kotlin/pull/106 is merged. Then with a proper release once available.

nicolas-guichard:
I've reworked the example to have a single project and better showcase how Kotlin and Java code can reference each other and added some checks.

asutherland (approved):
This is very exciting!  Thank you!

Thanks also for the added test checks!

One important note about the check names which we probably will need to address at some point is that although the test driver does understand the directory hierarchy and propagates the entire relative path for the check `tests/tests/checks/inputs/analysis/java/InlineObject.kt/method__json` to the "snapshots" subtree, unfortunately the test driver will get confused if there's any other check whose filename is `method__json` regardless of the path hierarchy.  That's why in some cases you may see awkward redundant encoding of information already in the directory hierarchy into the filename itself.  Note that I'm not suggesting you should change any check filenames here... just that we may need to enhance the test driver in the future.

tools/src/bin/scip-indexer.rs
```diff
@@ -774,7 +855,12 @@ fn analyze_using_scip(
                 // so only process the first element.
```
nicolas-guichard, line 855: This does not work as expected with Java's interfaces which enables some kind of multiple inheritance.
Take for instance [org.webrtc.AndroidVideoDecoder](https://dev.searchfox.org/mozilla-central/search?q=symbol:S_java_java_org%2Fwebrtc%2FAndroidVideoDecoder%23). It implements both `VideoDecoder` and `VideoSink` but only the former gets picked up here.

There is worse though, for [org.mozilla.thirdparty.com.google.android.exoplayer2.ExoPlayerImpl](https://dev.searchfox.org/mozilla-central/search?q=symbol:S_java_java_org%2Fmozilla%2Fthirdparty%2Fcom%2Fgoogle%2Fandroid%2Fexoplayer2%2FExoPlayerImpl%23) which has this class diagram:
```mermaid
classDiagram
    class Player
    <<interface>> Player

    class ExoPlayer
    <<interface>> ExoPlayer

    class BasePlayer
    <<abstract>> BasePlayer

    class ExoPlayerImpl
    <<final>> ExoPlayerImpl

    BasePlayer <|-- ExoPlayerImpl : extends
    ExoPlayer <|-- ExoPlayerImpl : implements

    Player <|-- ExoPlayer : extends

    Player <|-- BasePlayer : implements
```
The only superclass of `ExoPlayerImpl` picked up by Searchfox here is `Player`. Iow the order in which the relationships appear in scip_sym_info is not guaranteed to be from closest to farthest.

The cleanest fix would be to have the SCIP data distinguish between direct parents and other ancestors − or provide an option to only list direct parents.

tools/src/bin/scip-indexer.rs
```diff
@@ -774,7 +855,12 @@ fn analyze_using_scip(
...
                 if let Some(rel) = scip_sym_info.relationships.first() {
                     if let Some(rel_sinfo) = scip_symbol_to_structured.get(&rel.symbol) {
```
nicolas-guichard, line 857: This assumes that the SymbolInformation of the supers/overrides has already been seen, which isn't necessarily the case. For instance with that code no class relationship information is generated:
```java
class Derivative extends Base {}
class Base {}
```

This could be fixed in a number of ways:
1. Extracting the processing of relationships in another pass to make sure that `scip_symbol_to_structured` has been properly populated already.
2. Doing like in 57ca05a2 and adding entries to `scip_symbol_to_structured` as needed, and when processing the relevant `SymbolInformation` again updating the data if an empty entry was already created. `Structured{Super,Override}Info::pretty` may get out of sync.
3. Only keep the symbol in `Structured{Super,Override}Info`, that removes the issue of getting out of sync altogether.

My preferred option is option 3, because I don't see where the `Structured{Super,Override}Info::pretty` and `StructuredSuperInfo::props` are currently used anyway, and that's duplicated data that belongs more to the target than to the relationship (unless the objective is to use relative paths like `super::SomeTrait`).

nicolas-guichard (commented):
After some extra testing I found a couple more issues with inheritance.

tools/src/bin/scip-indexer.rs
asutherland: Option 3 is fine.  The current structured record type is largely just what my first draft of the structured representation was for C++ and was biased towards what clang provided and some JS-based prototyping I was doing for diagramming benefited from[1].  We definitely do not need to populate things at analysis/indexing time that can be handled by the cross-reference process.

The "pretty" is something I think we can move towards removing/eliminating since it does canonically come from the symbol definition itself and we do largely ignore it.  (And indeed, for the reciprocal relationships to super and override, we only emit the symbol.)

The "props" is something I would like to retain, but it only needs to be used for cases where the information is specific to the relationship between the current symbol and the upward-edge to the super/override.  There is no need to propagate information from the related symbol to the current symbol as part of indexing/analysis.  If we end up wanting that information on the current symbol for runtime performance reasons, we can do that during the crossref process as well.

1: In my JS prototype, diagramming happened client-side and with us having to perform a network request per symbol lookup, so there was a bias towards including information about the targets of edges in a symbol since it could potentially provide an order of magnitude of reduction in diagram lookups.  These are not concerns we have anymore.

tools/src/bin/scip-indexer.rs
asutherland: > The cleanest fix would be to have the SCIP data distinguish between direct parents and other ancestors − or provide an option to only list direct parents.

Agreed.  I also think that while it would be great if SCIP-producing upstreams expressed this nuance, it may not realistically be something we can depend on or require of upstreams as I think for the general VS code language server use case there is no need to care, even though searchfox might care.

The crossref mechanism can potentially undo the transitive inclusion process here.  Perhaps we could use the "props" array to include something like "possibly-transitive" in cases when we don't know if the edge is direct or transitive.  crossref could then eventually process these edges, normalizing them by retaining the edge on cases that seem direct (and removing the "possibly-transitive" prop) and removing the edges that seem like they're only included because they're transitive.  I don't think we need to do that in this PR, but it would be nice as a follow-up[1].

In particular, I don't believe the "search" endpoint would be bothered by the transitive closure here, although it does make it more likely that the search results will be overwhelming.  It's only experimental things like [inheritance-diagram diagrams](https://asuth.searchfox.org/mozilla-central/query/default?q=inheritance-diagram%3Aworkercontrolrunnable+graph-layout%3Aneato) and `crossref_expand` (which currently is not usable both because the "query" endpoint UI needs improvement, but also because I think it has bugs) which would care.

asutherland:
The `--no-emit-inverse-relationships` commit looks good too; a very appropriate, targeted use of a check to provide additional clarity for our current behavior and confidence that when/if things change in the future, we know the review process will notice!

tools/src/bin/scip-indexer.rs
asutherland: To clarify, I think it's fine to remove this constraint that we only process the first element in this PR.  Also fine to do it in another PR, but I think the upsides of having the full set of direct parents probably outweighs the downsides of having to deal with the full transitive set.

asutherland:
Per zoom call I'm landing this now and iterative improvements like traversing all parents will happen in follow-up PRs.

tools/Cargo.toml
```diff
@@ -61,6 +61,8 @@ tree-sitter = "0.20.9"
 tree-sitter-python = "0.20.2"
 tree-sitter-rust = "0.20.3"
 tree-sitter-typescript = "0.20.2"
+tree-sitter-java = "0.20.2"
+tree-sitter-kotlin = { git = "https://github.com/fwcd/tree-sitter-kotlin.git" }
```
fwcd, line 65: Sorry for the trouble, but I might have broken this by reverting https://github.com/fwcd/tree-sitter-kotlin/pull/106 upstream. The PR bumped the parser from 22 MB to 70 MB, which is too large (GitHub starts emitting warnings once files exceed 50 MB), we need to take some more time to get this down to a more reasonable size.

In the meantime please pin this to a commit (e.g. https://github.com/fwcd/tree-sitter-kotlin/commit/5b7ca7a611eac3234cd0a38b6257effa8888f89d).

asutherland: @fwcd Thanks so much for reaching out about this!  I will set the pin!

asutherland: I've pinned in https://github.com/mozsearch/mozsearch/pull/698

PR #669, opened 2023-11-19
Bug 1621788 - Add "Remove the Permalink" button for the latest revision.

arai-a:
This changes:
  * Add "Remove the Permalink" link to the Navigation box for the latest revision
    * This is a link to the same thing as "Go to latest version" in older revision
    * This is disabled when the URL is not permalink
    * This is enabled when the URL is permalink
    * The enabled/disabled status is dynamically updated when clicking "Permalink" and "Remove the Permalink"

(left: navigation box for non-permalink, right: navigation box for permalink)
remove-permalink

The "Permalink" item isn't disabled, because it has "copy" button, and I think the button should be available regardless of the current URL.

### DIFF
--- a/static/js/panel.js
+++ b/static/js/panel.js
@@ -12,6 +12,7 @@ var Panel = new (class Panel {
     this.accelEnabledCheckbox = document.getElementById("panel-accel-enable");
 
     this.permalinkNode = this.findItem("Permalink");
+    this.unpermalinkNode = this.findItem("Remove the Permalink");
     this.logNode = this.findItem("Log");
     this.rawNode = this.findItem("Raw");
 
@@ -76,6 +77,20 @@ var Panel = new (class Panel {
       });
     }
 
+    const updateUnpermalink = () => {
+      if (/^\/[^\/]+\/rev\//.test(document.location.pathname)) {
+        if (this.unpermalinkNode) {
+           this.unpermalinkNode.classList.remove("disabled");
+        }
+      } else {
+        if (this.unpermalinkNode) {
+           this.unpermalinkNode.classList.add("disabled");
+        }
+      }
+    };
+
+    updateUnpermalink();
+
     if (this.permalinkNode) {
       this.permalinkNode.addEventListener("click", event => {
         if (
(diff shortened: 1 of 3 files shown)


static/js/panel.js
```diff
@@ -76,6 +77,20 @@ var Panel = new (class Panel {
       });
     }
 
+    const updateUnpermalink = () => {
+      if (document.location.pathname.startsWith("/searchfox/rev/")) {
```
arai-a, line 81: just realized this is wrong. this works only on test website.
I'll fix this part

asutherland:
This works great, thank you!

UX-wise, in https://bugzilla.mozilla.org/show_bug.cgi?id=1621788#c1 where I said:
> For UX consistency we could certainly detect these situations and have something that says "You're seeing the latest version right now via a permalink. `[Remove the Permalink]`".

I believe I was proposing doing something with the the textual description of the current revision under the search box at the top of the page that can look like:
> Showing [a9cb8b07](https://searchfox.org/mozilla-central/commit/a9cb8b07d4cf4290e693bc31ba23412d0f042cb8): Bug [1863685](https://bugzilla.mozilla.org/show_bug.cgi?id=1863685) - update expectations for subtests in test video-tag.https.html to cover OSX and windows x64.

But I think the approach in this patch is probably more likely to be useful to people because:
- It's more discoverable in the navigation sidebar next to the permalink button.  My mental model is that no one pays any attention to the revision description line under the search box, but that's not scientific, just anecdotal experience that when people have reported problems about stale revisions they haven't realized the affordance is there.
- In https://bugzilla.mozilla.org/show_bug.cgi?id=1517978 for hyperannotate I am planning to potentially overhaul the space under the search field and while I think I may provide an affordance as originally envisioned, I don't think that changes the utility of having something in the navigation panel for this.

So I'm not sure that "Remove the Permalink" necessarily makes sense as a string for this more terse context, especially since if we get rid of the "the" we potentially have room for a hotkey.  Although I'm not sure what hotkey we would use; if we had one on "Go to latest version" then that would be the clear choice.  I believe we cribbed the "y" key from github, but it seems like on github you have to hit "w" and then select the branch in question to go to the latest version.  I think it's probably highly desirable that the "y" key is idempotent rather than toggling states.  I suppose I might propose "h" as a QWERTY-centric choice as the "h" key is adjacent to the "y" and on the same right side of my split keyboard and "l" is already used by github as "jump to line" so that would be confusing to reuse.

I do think having the "Remove the Permalink" option disabled/enabled does seem like it provides nice context awareness of what's going on given the existing "permalink" affordance.  @emilio @smaug---- do you have thoughts on this?  Specifically:
1. Should we add a hotkey for "go to latest version" that we also add for this case?
2. Does it make sense that this is a separate menu item from "Permalink"?  The other alternatives would seem to be:
  -  Have a single menu item where clicking "Permalink" cycle the menu item to "Remove permalink" or something like that.  This runs into the eternal UI question of "is this menu item describing the current state, or what clicking on it will do?".
  - Have 2 menu items that are effectively radio buttons `(o) Latest Revision [H]` and `( ) Permalink [Y]`.  They could also both still have the copy icons.  This could also potentially want an additional row that's just a label, or maybe we add a `<title>` to each.

asutherland:
I'm going to land this as-is and we'll see if people notice or have thoughts.  Thank you so much for your continued enhancements to searchfox!

PR #670, opened 2023-11-19
Bug 1818027 - Handle spread inside object in JS.

arai-a:
Added dedicate path for `SpreadExpression` inside `ObjectExpression` and `ObjectPattern`, to handle `({...foo})` syntax in JS


asutherland:
Thanks!  Normally it would be good to add a [testing check](https://github.com/mozsearch/mozsearch/blob/master/docs/testing-checks.md) for a change like this, but given the potential move to scip-typescript we'd potentially have to redo all the checks and test files anyways, so it makes sense to punt on that for now.

PR #672, opened 2023-11-21
Fixes to SCIP ingestion: multiple inheritance & ignoring declaration order when parsing relationships

nicolas-guichard:
Extra fixes to the SCIP ingestion as discussed in /pull/667.

### DIFF
--- a/clang-plugin/MozsearchIndexer.cpp
+++ b/clang-plugin/MozsearchIndexer.cpp
@@ -1064,7 +1064,6 @@ public:
 
         J.objectBegin();
 
-        J.attribute("pretty", getQualifiedName(BaseDecl));
         J.attribute("sym", getMangledName(CurMangleContext, BaseDecl));
 
         J.attributeBegin("props");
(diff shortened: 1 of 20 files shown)


asutherland (approved):
Thanks!  And thank you for the attention to making sure the docs stay in sync.

Once this is landed please also submit a phabricator request to keep the https://searchfox.org/mozilla-central/source/build/clang-plugin/mozsearch-plugin/MozsearchIndexer.cpp that's in-tree for mozilla-central in sync.  This is just because we like to keep the files in-sync; there's no need for us to uplift to older branches because having the extra field data isn't a problem (but there's also no harm if the change gets uplifted).

Restating:
- The removal of "pretty" will not cause a problem for ingesting data that still includes that field because the default behavior for serde_json is to accept unknown fields unless the container attribute [`deny_unknown_fields`](https://serde.rs/container-attrs.html#deny_unknown_fields) is used.  We don't use that field and have no reason to ever use that field because it would complicate our lives.

PR #673, opened 2023-11-24
Bug 1865048 - Java/Kotlin cross-language support

nicolas-guichard:


nicolas-guichard:
I'm putting this here as draft to discuss some design details, and some implementation details too but I don't really expect a full review at this moment.
(This is also a bit of a brain dump.)

### Actual binding analysis

0199e6d does really two things (and maybe I'll split it into two commits):
 1. It looks at `__attributes__((annotate("binding_to_something"/"bound_from_something", "symbol")))` and generates slotOwner/bindingSlots accordingly those attributes can be set manually but they don't have to. I will probably change that a bit to avoid having the "binding_to_java_member_kind" annotation. Those attributes can be set manually (which is done in 04b616e), but also:
 1. It looks for patterns which match what the WrapForJNI annotation generates and add those  __attributes__ to the AST where appropriate. This code silently bails out as soon as it can't traverse the AST as expected anymore. (And I definitely miss the `?` Rust operator in C++.)

This has a few false-positives, for instance the C++ type [mozilla::widget::detail::StringArray](https://searchfox.org/mozilla-central/source/widget/android/EventDispatcher.cpp#195) is a binding to an array of Java Strings. It gets a `slotOwner.sym = "[Lorg/java/String;#"`. I don't think it hurts because those will get discarded later on.

I also have an issue with overloaded bindings for instance how to properly match C++ [GeckoViewSupport::PrintToPdf](https://searchfox.org/mozilla-central/source/widget/android/GeckoViewSupport.h#114) overloads with the corresponding Java [GeckoSession::Window::PrintToPdf](https://searchfox.org/mozilla-central/source/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/GeckoSession.java#1444) overload. For now I sort them by definition order in C++ and expect that to match with the declaration order in Java, but this could be wrong. This is not too common though, so I think a first version we could do with that heuristic and properly document it to be fixed sometime later.

### Target LLVM version

3e14c46 stems from something that is unclear to me: what is our target Clang/LLVM/C++ standard version? In the Docker container clang/llvm14 is available and llvm-config returns -std=c++14. The mozilla-central build seems to be using clang/llvm17 and C++17 already.
This yields two incompatibilities:
 * dyn_cast_or_null was deprecated and replaced by dyn_cast_if_present, but is still available → we could just use dyn_cast_or_null everywhere
 * in clang17 clang::Optional was removed in favor of std::optional (which is C++17)

### Subtrees?

A more interesting design question is about the concept of subtrees: at the moment SCIP symbol names incorporate the subtree name, which has a couple of issues:
 * identical symbols from multiple subtrees don't get linked together, for instance if we had 2 java subtrees A and B, `S_java_A_org/junit/Test#` won't get linked to `S_java_B_org/junit/Test#`. Same thing with Rust crates etc.
 * when analyzing java bindings in the clang plugin we don't know the name of the subtree where the Java symbol is supposed to be

I removed the whole `S_<lang>_<subtree>_` prefix in a7605380 but the `S_<lang>_` part doesn't really hurt here. What's your position on removing the `<subtree>_` part from SCIP-derived symbols?

### IDL vs direct bindings

Finally 8fcf83e shows that the bindingSlots/slotOwner system was built with IDL in mind instead of direct bindings: when processing an IDL file only bindingSlots are generated in the analysis files and the slotOwner are deduced from that later on. Here I sometimes have a slotOwner and need to add that to the right bindingsSlot in the other direction.

I'm also thinking of adding a field ownerLang besides the existing slotLang in BindingSlotInfo to specify that those bindings are IDL bindings or direct bindings to another language. That could be used in the search page to better display things.

asutherland:
> ### Actual binding analysis
> I also have an issue with overloaded bindings for instance how to properly match C++ [GeckoViewSupport::PrintToPdf](https://searchfox.org/mozilla-central/source/widget/android/GeckoViewSupport.h#114) overloads with the corresponding Java [GeckoSession::Window::PrintToPdf](https://searchfox.org/mozilla-central/source/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/GeckoSession.java#1444) overload. For now I sort them by definition order in C++ and expect that to match with the declaration order in Java, but this could be wrong. This is not too common though, so I think a first version we could do with that heuristic and properly document it to be fixed sometime later.

Agreed that it makes sense to defer worrying about overloads for the time being.  Especially since human-authored searches will usually operate in "pretty" space, as do the diagram context menu options.

asutherland:
> ### Subtrees?
> 
> A more interesting design question is about the concept of subtrees: at the moment SCIP symbol names incorporate the subtree name, which has a couple of issues:
> 
>     * identical symbols from multiple subtrees don't get linked together, for instance if we had 2 java subtrees A and B, `S_java_A_org/junit/Test#` won't get linked to `S_java_B_org/junit/Test#`. Same thing with Rust crates etc.
> 
>     * when analyzing java bindings in the clang plugin we don't know the name of the subtree where the Java symbol is supposed to be
> 
> 
> I removed the whole `S_<lang>_<subtree>_` prefix in [a760538](https://github.com/mozsearch/mozsearch/commit/a7605380a7941157dceac4aacb1ed056e6ff7bed) but the `S_<lang>_` part doesn't really hurt here. What's your position on removing the `<subtree>_` part from SCIP-derived symbols?

I agree that for the Java/Kotlin case where there's a built-in, rigidly-enforced package namespace there's no need to have more than one subtree.

I introduced the subtree mechanism for JS/TS where:
- It's very common for code to use very limited namespacing based on the root of the (relative) build/source tree.  (Packages may be pulled in from npm which does have a namespace, but that's usually conceptually decoupled from the project source tree other than avoiding name collisions with top-level dependencies.)  For example, [searching on package.json files](https://searchfox.org/mozilla-central/search?q=path%3Apackage.json&path=) there's a number of independent roots under "browser/components" and "devtools".
- It's desirable to have the symbol names, when they involve file paths, to be relative to that build/source tree root.  The subtree name basically just amounts to an alias that lets us keep the symbol names much shorter, which can be a win in terms of:
   - Symbol readability: It's desirable for the symbol names to be comprehensible to humans, and so being able to shorten a path-based mapping like "browser-components-pocket" to just "pocket" is a win.
   - Symbol stability: The subtree serving as an alias means that symbols could rename stable if "browser/component/pocket" moves elsewhere in the tree.
   - Shorter symbols as a performance win: Trimming long strings out of the symbol names has clear upsides, especially when they would otherwise be a prefix.
   
What do you think about:
- Have the scheme be `S_<SCIP namespace>_`.
- Have the "SCIP namespace" default to whatever our language mapping chooses.  For Java/Kotlin that could be "java" or even "jv".  For the other languages it would be our current language mapping ("rs", "py", "js").
- scip-indexer could take an explicit namespace to use as an argument, so for the mozilla-central examples of "browser/components/pocket", we'd explicitly define that to be "pocket".

This still basically works for the `scip_subtrees` `config.json` scheme mechanism, although as I noted in response to Emilio's review comment at https://github.com/mozsearch/mozsearch/pull/663#discussion_r1372089592 that was basically a hack that already ended up being unsuitable, so it's not like there's any real need to keep that mechanism.

asutherland:
> ### IDL vs direct bindings
>
> Finally [8fcf83e](https://github.com/mozsearch/mozsearch/commit/8fcf83e94605d7d763e5ef045d9626e8e08987bf) shows that the bindingSlots/slotOwner system was built with IDL in mind instead of direct bindings: when processing an IDL file only bindingSlots are generated in the analysis files and the slotOwner are deduced from that later on. Here I sometimes have a slotOwner and need to add that to the right bindingsSlot in the other direction.

It definitely makes sense that it's necessary to expand the crossref's linkage logic as, indeed, it has only been using for XPIDL and IPDL so far and those very much are IDL languages.

> I'm also thinking of adding a field ownerLang besides the existing slotLang in BindingSlotInfo to specify that those bindings are IDL bindings or direct bindings to another language. That could be used in the search page to better display things.

I hope to find some more time to better understand the JNI idiom, but I want to raise 2 possibilities in brief that I'll be thinking about as I do:
- We could add `DirectCaller` and `DirectCallee` `BindingSlotKind` enum values for the direct call situation and not use `slotOwner` at all.  At least as long as we only expect there to exist a single paired language.  If we have both C++ bindings and Rust bindings that could both work, then it probably does make sense to treat the Java code as the slotOwner.
- We could add "linkage" records as another type of JSON output the indexers emit rather than having potentially different semantics for what `bindingSlots` and `slotOwner` mean before and after crossref.  Or at least, I may be reading too much into the changes in https://github.com/mozsearch/mozsearch/commit/8fcf83e94605d7d763e5ef045d9626e8e08987bf that clear the `binding_slots` and `slot_owner`.  Ideally if we are going to continue to perform linkage from their contents, the logic would be explicitly idempotent.  So if we're iterating over `binding_slots` on symA and see a `DirectCaller` of symB, then our `xref_link_slots` would track that we find-or-create a `DirectCallee` of symA when we look up symB.

That said, we can iterate on all of this without too much trouble, so I'm not concerned about getting it perfect as long as the context-menu behavior for users from source listings seems sane.

nicolas-guichard:
> We also have the "wubkat" and "nss" trees that we build on the EC2 indexers and where the version used is whatever we provisioned the indexer to use as "clang" and "clang++" based on how [indexer-setup.py uses those bare names](https://github.com/mozsearch/mozsearch/blob/d5aa2ec55ae3619c4d9a20afd8e6154067fd1a57/scripts/indexer-setup.py#L27-L28) and the [wubkat build script](https://github.com/mozsearch/mozsearch-mozilla/blob/440ec30f15bf8776fe9d9df3a213c7670349bc5a/wubkat/build#L24-L25) and [nss build script](https://github.com/mozsearch/mozsearch-mozilla/blob/440ec30f15bf8776fe9d9df3a213c7670349bc5a/nss/build#L10-L12) just use those.
> 
> ESR115 being on clang 16 is slightly annoying, but as long as we don't make any breaking changes to MozsearchIndexer.cpp's JSON format that require us to uplift changes to MozsearchIndexer to the ESR tree, I think it's fine to say that our minimum clang version is 17 as long as wubkat doesn't break.

OK then I'll open another PR to update clang to v17 in the AMI (it currently still is v14). If the wubkat and nss trees compile fine with that, we would only have to care about one single version.

> What do you think about:
>  * Have the scheme be `S_<SCIP namespace>_`.
>  * Have the "SCIP namespace" default to whatever our language mapping chooses.  For Java/Kotlin that could be "java" or even "jv".  For the other languages it would be our current language mapping ("rs", "py", "js").
>  * scip-indexer could take an explicit namespace to use as an argument, so for the mozilla-central examples of "browser/components/pocket", we'd explicitly define that to be "pocket".

That should work indeed! I think we should keep the language name in all cases though, instead of replacing it with the subtree name: I mean have the scheme be `S_{rs,py,js,jvm}_[<optional subtree name>_]`.

> * We could add "linkage" records as another type of JSON output the indexers emit rather than having potentially different semantics for what `bindingSlots` and `slotOwner` mean before and after crossref.  Or at least, I may be reading too much into the changes in [8fcf83e](https://github.com/mozsearch/mozsearch/commit/8fcf83e94605d7d763e5ef045d9626e8e08987bf) that clear the `binding_slots` and `slot_owner`.

I think you are reading a bit too much from those `clear`s indeed ;-)
Everything that gets removed from binding_slots on [`piece.binding_slots.clear();`](https://github.com/mozsearch/mozsearch/commit/8fcf83e94605d7d763e5ef045d9626e8e08987bf#diff-c3382b8b0b9865f726f9cfcf1090c8525b2eed959136b87232229cfd8582756cR376) is then [pushed back from `xref_link_slots` in the later loop](https://github.com/mozsearch/mozsearch/commit/8fcf83e94605d7d763e5ef045d9626e8e08987bf#diff-c3382b8b0b9865f726f9cfcf1090c8525b2eed959136b87232229cfd8582756cR403). The clear is just there to avoid duplicates and would not be needed if `AnalysisStructured::binding_slots` was a `Set` for instance. The semantics of `bindingSlots` and `slotOwner` don't change and the crossref step remains idempotent: this only aims to ensure that each entry in `binding_slots` has a matching `slot_owner` (which was already the case) and vice versa that each `slot_owner` gets a matching entry in the relevant `binding_slots`. If everything is already populated both ways then nothing should change (save for the ordering in `binding_slots`, and some duplicates because I forgot to dedup `xref_link_slots` but that's a bug).

nicolas-guichard:
I believe this is now ready for review, along with mozsearch/mozsearch-mozilla/pull/220 and deployed on dev.searchfox.org for testing already.

I listed a few examples on https://bugzilla.mozilla.org/show_bug.cgi?id=1865048.

How do you want to go about the clang plugin changes? I'm thinking of having them reviewed here then having a single commit on mc to sync the changes to the in-tree version?

asutherland:
> How do you want to go about the clang plugin changes? I'm thinking of having them reviewed here then having a single commit on mc to sync the changes to the in-tree version?

Yeah, this is how we usually do it.

The exceptional case is that sometimes :glandium or someone will provide patches to the C++ indexer for m-c as they are trying to add support for a new version of clang, in which case they land on m-c first and then we sync them over to mozsearch from there.

asutherland:
@theres-waldo I've added you as a reviewer here because I'm interested in your thoughts/feedback on the specific changes and the meta approach, but especially given how busy this week is at the (metaphorical) office, there's no obligation to review, although if you'd like us to delay landing until you've had more time to review, please do let us know.

asutherland (approved):
This is super fantastic and amazing, thank you!  Note that mozilla-central will be in soft-freeze until Monday so while it's okay to prep an m-c patch before Monday, I won't be able to land it until then.

The below is just notes for future me, you don't need to read them.

### Restating
- This implementation is centered on using
  [C++ annotations](https://en.cppreference.com/w/cpp/language/attributes) to
  describe binding relationships so that it's possible to manually author
  binding relationships.  Automatically detected/inferred bindings are first
  mapped into annotations, ensuring consistency of data-flow.
  - Our C++ annotations use the pre-existing
    [Annotate](https://searchfox.org/llvm/source/clang/include/clang/Basic/Attr.td#865-883)
    attribute that's
    [sorta documented](https://clang.llvm.org/docs/AttributeReference.html#annotate).
    (If we didn't use an existing attribute, we'd need to modify
    [Attr.td](https://searchfox.org/llvm/source/clang/include/clang/Basic/Attr.td)
    to understand our attribute and use a custom clang build, which is not
    desirable!)
  - Our annotations look like
    `__attribute__((annotate("binding_to", "LANGUAGE", "KIND", "SEARCHFOX SYMBOL")))
    where:
    - `LANGUAGE` corresponds to the `BindingSlotLang` where "jvm" is used for
      both Java and Kotlin because they exist in the same namespace and that's
      really what we care about.
    - `KIND` is "class"/"method"/"getter"/"setter"/"const" corresponding to
      `BindingSlotKind`.
    - `SEARCHFOX SYMBOL`: The expected corresponding searchfox symbol for the
      referenced counterpart of the symbol we're annotating.  For Java/Kotlin,
      this will look like `S_jvm_` followed by scip-indexer's translation of
      the SCIP descriptors generated by scip-java that tries to resemble the
      SCIP string syntax but should not be assumed to match exactly.
- MozsearchIndexer doesn't have our clang-format rules run against it because [everything under build/clang-plugin/ is ignored](https://searchfox.org/mozilla-central/rev/655f49c541108e3d0a232aa7173fbcb9af88d80b/.clang-format-ignore#1-2) because it's assumed to match LLVM's coding standards and [searchfox's plugin lives at build/clang-plugin/mozsearch-plugin](https://searchfox.org/mozilla-central/source/build/clang-plugin/mozsearch-plugin).  (And the [LLVM guidelines indeed say not to brace simple single statement bodies](https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements).)

### Minutae

Binding Operations helpers that aren't excessively obvious for the clang plugin:
- hasReverseQualifiedName: takes [crbegin, crend, NamedDecl] where the NamedDecl
  is expected to be the innermost type and then it can walk up the context and
  make sure the names line up.
- fieldNamed: Locates NamedDecl with the given name (and has an identifier),
  null otherwise.
- namedFieldValue: It it finds a field named "name" on the given RecordDecl that
  dyncasts to a VarDecl, it pierces to the definition, then tries to get its
  initialization expression and cast that to a string, returning the contents of
  the string.  This is used for dealing with structs/classes that name themselves
  with idioms like `static constexpr char name[] = "foo";`
- AbstractBinding: Helper base class for emission of StructuredBindingSlotInfo.
  Its subclasses define `ANNOTATION` fields that name the annotation we add to
  decls.  Specific subclasses:
  - BindingTo: Annotation of "binding_to"
  - BindingAs: Annotation of "binding_as".
- setBindingAttr: Creates the
  [Annotate](https://searchfox.org/llvm/source/clang/include/clang/Basic/Attr.td#865-883)
  attribute with our 3 args of [lang, kind, symbol].
- readBinding: Reads the contents of an
  [Annotate](https://searchfox.org/llvm/source/clang/include/clang/Basic/Attr.td#865-883)
  attribute, producing an `AbstractBinding` struct which can be used to
  initialize the subclasses.  Non-conforming annotations are mapped to nullopt.
- getBindingTo: Produces a BindingTo from the first present and valid
  "binding_to" binding on the given Decl, nullopt otherwise.
- getBoundAs: Produces a vector of BindingAs from all valid "binding_as"
  annotations found on the given Decl.
- FindCallCall: Exposed as `FindCallCall::search(Stmt*)`, uses a recursive
  visitor to find CallExpr instances where it applies heuristic rules that
  recognize patterns like
  `return mozilla::jni::Method<UpdateOverscrollOffset_t>::Call(Compositor::mCtx, nullptr, a0, a1);`
  using the `Method` template magic with the trait
  `struct UpdateOverscrollOffset_t` exists and has a name that
  `namedFieldValue` can recognize like
  `static constexpr char name[] = "updateOverscrollOffset";`
- javaScipSymbol: Derives the symbol we think scip-indexer would generate for
  the given binding [using its mapping from the structured descriptors](https://github.com/mozsearch/mozsearch/blob/a4a1d3d08d7cef34f91be39cf276972fe7842659/tools/src/bin/scip-indexer.rs#L466)
  which looks like SCIP string syntax because we didn't want to reinvent the
  wheel but our consistency is with scip-indexer.rs, not SCIP proper.
- addSlotOwnerAttribute: If getBindingTo finds a binding, writes out a valid `slotOwner` structured
  record object attribute to the provided JSON OStream, writes nothing otherwise.
- addBindingSlotsAttribute: If getBoundAs returns a non-empty vec, writes out a valid `bindingSlots`
  structured record object array, otherwise writes nothing.
- demangleJniCallPart: destructively consumes the JNI mangled function name,
  returning the transformed symbol name and leaving the signature intact (but
  which could then be demangled).  The processing loop consumes single-`_`
  delimited tokens, with `__` then being seen by the loop as a token starting
  with `_` which leads to an early return.
- scipSymbolFromJnicallFunctionName: Uses demangleJniCallPart plus other
  heuristics (strip off `Java_` prefix, do some transforms) to normalize the
  symbol plus adds a basic ordinal overload using a static map (which works
  fine with our per-compilation-unit lifetime).
- findBindingToJavaClass: Looks for classes subclassing `mozilla::jni::ObjectBase`
  per `isMozillaJniObjectBase` and if it finds one, pulls out the name via
  `nameFieldValue`, builds the symbol in a forward direction from the name via
  `javaScipSymbol`, builds the `BindingTo` binding, then sets the binding
  annotation via `setBindingAttr`.
- findBindingToJavaFunction: Uses `scipSymbolFromJnicallFunctionName` to figure
  out if it's a JNI function if it demangled successfully, then builds the
  binding as `BoundAs` knowing it has to be a method, and annotates via
  `setBindingAttr`.
- findBindingToJavaMember: Very nice illustrative pattern comments on this and
  the subsequent helpers.  First validates that the parent has had a binding
  annotated onto it and that it has a body.  Then uses `FindCallCall::search` on
  the body (discussed above) and demonstrated by the comment to build a
  `BindingTo` and annotate it.
- findBindingToJavaConstant: Given a VarDecl which can only be a constant,
  validates parent is a binding, then uses the field to build a `BindingTo` for
  the constant and annotate it.
- findBoundAsJavaClasses: Wow does this feel elegant compared to other options!
  We look for a grandparent class of `mozilla::jni::NativeImpl`, bind the class
  based on extracted "name", then find the "methods", traverse their inits, and
  then walk and bind each of those methods with a bit of boilerplate.
- emitBindingAttributes: Helper to call addSlotOwnerAttribute and then
  addBindingSlotsAttribute given a single decl.

MozsearchIndexer changes:
- Existing emitStructuredInfo methods gain calls to the `emitBindingAttributes` helper
  so that if they see the annotations (which could have been manually specified
  in the source or automatically added), they get emitted.
- New emitStructuredInfo for `VarDecl` for static class members.
- VisitNamedDecl
  - if it's a RecordDecl and if it's a definition
    - if it's a CXXRecordDecl
      - findBindingToJavaClass (for "binding_to")
      - findBoundAsJavaClasses (for "binding_as")
    - then we call emitStructuredInfo
  - if it's a FunctionDecl and if it's a definition
    - if it's a CXXMethodDecl
      - findBindingToJavaMember -> FindCallCall
    - else
      - findBindingToJavaFunction
    - then we call emitStructuredInfo
  - if it's a VarDecl and it's not part of a template and we're not in a
    template traversal and the context of this VarDecl is a CXXRecordDecl,
    call findBindingToJavaConstant and use our brand new `emitStructuredInfo`
    for VarDecls.

router.py:
- Differentiate slotOwner traversal based on ownerLang so that only "idl" gets
  mapped to "IDL" and for non-IDL languages it's a "Definitions".
- Differentiate bindingSlots traversal similarly; "Definitions" for IDL versus
  new "Bindings".

idl-analyze.py:
- The ownerLang is always "idl", naturally.

Test config:
- We stop using `scip_subtrees` in favor of directly invoking scip-indexer in
  the "build" step because we don't want the subtree_name which is now optional.

Checks:
- We changed the scip-indexer mangling so all the SCIP indexer symbols lose the
  tree name.
- We of course now have all the new "ownerLang" fields.
- Some reordering due to making `xref_link_slots` a BTreeMap so the ordering is
  consistent.

scip-indexer.rs:
- subtree_name is now optional; symbol building moved into a helper because
  the existing direct `format!` invocations don't want to all bifurcate.

PR #677, opened 2023-12-04
Update to clang 17

nicolas-guichard:
Attempt at fixing the wubkat index, and making the clang version we use closer to the one the Firefox build uses.
nixpkgs doesn't have llvmPackages_17 yet, so updating to 16 there for the time being.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -51,9 +51,10 @@
             pkg-config
 
             # Dependencies required to build clang-plugin
+            # Update to llvmPackages_17 when available
             clang
-            llvmPackages_14.libllvm
-            llvmPackages_14.libclang
+            llvmPackages_16.libllvm
+            llvmPackages_16.libclang
 
             scip
             protobuf
--- a/infrastructure/common-provision-pre.sh
+++ b/infrastructure/common-provision-pre.sh
@@ -12,12 +12,12 @@ set -o pipefail # Check all commands in a pipeline
 # Note that for the most recent LLVM/clang release (ex: right now v13), you
 # would actually want to leave this empty.  Check out https://apt.llvm.org/ for
 # the latest info in all cases.
-CLANG_SUFFIX=-14
+CLANG_SUFFIX=-17
 # Bumping the priority with each version upgrade lets running the provisioning
 # script on an already provisioned machine do the right thing alternative-wise.
 # Actually, we no longer support re-provisioning, but it's fun to increment
 # numbers.
-CLANG_PRIORITY=412
+CLANG_PRIORITY=413
 # The clang packages build the Ubuntu release name in; let's dynamically extract
 # it since I, asuth, once forgot to update this.
 UBUNTU_RELEASE=$(lsb_release -cs)
--- a/tests/tests/checks/snapshots/analysis/cpp/template_shapes.cpp/check_glob@Foo_Project-2.snap
+++ b/tests/tests/checks/snapshots/analysis/cpp/template_shapes.cpp/check_glob@Foo_Project-2.snap
@@ -7,7 +7,7 @@ expression: "&json_results"
     "loc": "00022:7-11",
     "source": 1,
     "syntax": "decl,function",
-    "type": "void (class Rectangle)",
+    "type": "void (Rectangle)",
     "pretty": "function DrawingContext::draw",
     "sym": "_ZN14DrawingContext4drawE9Rectangle"
   },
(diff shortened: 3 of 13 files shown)


asutherland (approved):

nicolas-guichard:
The clang plugin builds and runs fine with clang16/17, since that's what mc uses already, so that shouldn't break.
The test repo shows some differences in type names, mostly `class ClassName` → `ClassName` or `SomeClass::inner_type` → `inner_type`.
There are also some changes in the javascript checks. I don't understand why, do you?

asutherland:
> The test repo shows some differences in type names, mostly `class ClassName` → `ClassName` or `SomeClass::inner_type` → `inner_type`.

This isn't super surprising; in my hobby stack I have some changes to make us explicitly do more to get and emit the canonical type because of some hacky type parsing I do for the class diagram functionality.  Those will change back (and gain new "class " prefixes, etc.) when I put those changes up for review.

> There are also some changes in the javascript checks. I don't understand why, do you?

I think that's from the non-deterministic aspect of our provisioning where we just download the [latest JS shell that provides parsing/reflection for js-analyze.js](https://github.com/mozsearch/mozsearch/blob/2b632f035830b4a3f68ca1a8316d9a2e7ae010a0/infrastructure/indexer-update.sh#L48).

nicolas-guichard:
Alright, then with this doubts “removed” I let you either merge this now and handle any error, or leave it to me and I'll merge it tomorrow.

asutherland:
Oh, the JS column number changes are https://bugzilla.mozilla.org/show_bug.cgi?id=1862692 moving from 0-based to 1-based and we probably will need to deal with that.  I may try a quick pass at fixing that today after landing this since I think this could explain apparent regressions in our JS behavior.

asutherland:
I rebuilt docker locally and provisioned on AWS and both provisioning emails indicated success.  I am going to briefly address the js-analyzer issue and then re-trigger config1 and config5 so we can see what happens.

asutherland:
config1 experienced a build failure for some NSPR tests it looks like, but [we explicitly mark nss as continue in config1.json](https://github.com/mozsearch/mozsearch-mozilla/blob/440ec30f15bf8776fe9d9df3a213c7670349bc5a/config1.json#L10) so it just is experienced as a warning:

```
../../../pr/tests/testfile.c:576:31: error: incompatible function pointer types passing 'PRInt32 (void *)' (aka 'int (void *)') to parameter of type 'void (*)(void *)' [-Wincompatible-function-pointer-types]
  576 |                               DirTest, &thrarg,
      |                               ^~~~~~~
../../../pr/tests/testfile.c:93:36: note: passing argument to parameter 'start' here
   93 |                             void (*start)(void *arg),
      |                                    ^
5 warnings and 2 errors generated.
MOZSEARCH: /mnt/index-scratch/nss/git /mnt/index-scratch/nss/analysis/ /mnt/index-scratch/nss/dist/
make[2]: *** [../../../config/[rules.mk:450](http://rules.mk:450/): testfile.o] Error 1
make[2]: Leaving directory '/mnt/index-scratch/nss/nspr/Linux6.2_x86_64_clang_glibc_PTH_64_DBG.OBJ/pr/tests'
make[1]: *** [Makefile:143: build_nspr] Error 2
make[1]: Leaving directory '/mnt/index-scratch/nss/git'
make: *** [Makefile:51: nss_build_all] Error 2
+ handle_tree_error 'tree build script'
+ local 'msg=tree build script'
+ echo 'warning: Tree '\''nss'\'' error: tree build script'
warning: Tree 'nss' error: tree build script
```

It would be good to see if there's an easy way to make the build happy.  https://github.com/nss-dev/nss#building-nss suggest there's a new way to build using its `build.sh` but depending on ninja and gyp.  We do already provision ninja and it does seem like gyp is available to provision (although not super recent?), so that might not be too bad.

While briefly looking for existing NSS bugs I saw that https://treeherder.mozilla.org/jobs?repo=nss exists which means that NSS could define searchfox jobs of its own, so if this ends up being tricky we could perhaps just remove the semantic indexing for NSS and see if there are users who are interested in setting up semantic indexing on taskcluster if they miss it.

asutherland:
We did get a clean wubkat config5/release5 run in there although the error emails make it seem like we might not have.

The first run I triggered failed on determining the correct revision to use due to a run of failures on the webkit buildbot (although it really should have determined an older revision, but maybe we need to look at changing a paging setting in our request), but the run after that succeeded.  Then today's lambda cron job seems to have failed in/around bmalloc which hopefully is something transient that will be fixed on their trunk.

PR #679, opened 2023-12-18
Bug 1583635 - C++ macro expansion

nicolas-guichard:


asutherland:
This is very exciting, thank you!

I did a try push on m-c at https://treeherder.mozilla.org/jobs?repo=try&revision=b55d47dc249d67ad73af4e687ecf0f26483caafc to try and get some UX discussion going[1] and it seems like there are some cases where a call to `getSpelling` can crash.  You may already be aware of those, but I thought I'd mention the try push.

1: I don't think we'll want a hover implementation for production, but this was definitely the most pragmatic solution for prototyping!

nicolas-guichard:
I didn't know that getSpelling could fail… Thanks for letting me know I'll look at that tomorrow.

I've added an alternate representation with expand/collapse buttons on the left of line numbers. I think it's generally better. (That first impl uses CSS :has, so test with FF 121+, I hear it's due for release shortly 😉)

Because this is per-line, it shows the context around the macro too, and if multiple macros are used on the same line they all get expanded together.

When a macro call spans multiple lines currently it will display the expansion in the middle of the call which isn't great. In that case the expansion should be displayed below the end of the call I guess. But then this doesn't really work with the idea of expanding all macros on the same line… How would this be expanded for instance?
```cpp
MACRO(
  "called",
  "on",
  "multiple",
  "lines"
) with ANOTHER_MACRO(
  "called", "on", "the", "last",
  "line"
)
```
Should everything be expanded together then? Is that the kind of edge cases we don't really care about?

asutherland:
Thank you for also prototyping the expansion toggle / inline approach!

I think my current proposal would be that for an initial production landing macro expansions should be:
- Initiated by the context menu, so potentially our first menu item would be `Expand macro MULTI_LINE_MACRO`.
- Use the `BlamePopup` triggered exclusively from the context menu because:
  - We have a good screen reader accessibility mechanism in place with the BlamePopup that I think can enable us to provide critical locality to the screen reader.  I think this would still potentially require some iteration, though.  In particular, it wouldn't surprise me if we ended up needing to parent the popup on the context menu but temporarily add `visibility: hidden` to the popup for non-screen-reader users.
  - It lets us largely sidestep the issue of the stateful-ness of expanded macros and the ability to link to the expanded lines which I think users might expect.
  - It also lets us sidestep the expectation that the user can click on the tokens in the expanded source like they could in a normal source listing.
  - We also sidestep the potential inconsistency for screen-reader purposes of the expanded source.  I would expect the expanded macro lines to be table rows just like the rest of the source listing.
  - The popup approach I think should also sidestep potential a lot of potential performance issues.
  - It does also avoid the multiple macros on a single-line problem.
- Do style the popup source listing to look more like the current page theme style, specifically on light mode it should have a light background for the source.

I do think there is longer term potential for doing the expansion approach, but as noted above, I think we'd want to integrate that into the table rendering mode.  And I think we'd want solutions for linking to the expanded lines as well in a way that doesn't bit-rot.  (Line number links against the HEAD without a revision currently bit-rot quite quickly, but only a line number link against head would be able to provide the semantic info necessary to have the macro expansions.  I do have some plans for that, but we need not address them here.)

nicolas-guichard:
I've replaced the trigger of the macro expansion popup with an entry in the context menu, where the entry is a preview of the actual expansion. I think it's especially useful for small expansions like constants instead of requiring another click.

I think it would be interesting to support cases where the macro expansions differ from platform to platform. The context menu solution could easily be extended to have “linux64 expansion:”, “macosx64 expansion:”, “android-armv7 expansion:”, … entries when they differ.

asutherland:
> I've replaced the trigger of the macro expansion popup with an entry in the context menu, where the entry is a preview of the actual expansion. I think it's especially useful for small expansions like constants instead of requiring another click.

Yes, this is fantastic!  I was thinking along these lines as well and your choice to place the value on its own really makes it work.

I think the max-width you've added is necessary, but I think we should probably double it to 600px.  Briefly testing on [mozilla::nsDisplayListBuilder::nsDisplayListBuilder](https://searchfox.org/mozilla-central/rev/008ce958417fd724e02748740d669aafd49debb3/layout/painting/nsDisplayList.cpp#646) from https://bugzilla.mozilla.org/show_bug.cgi?id=1743252 which is a good pathological case, it lays out at about 584px in width for me and that feels like it's at the upper bound of what is reasonable.  (Note that I've provided a permalink but one will need to go to the latest revision and then maybe seek to the right line.)

> I think it would be interesting to support cases where the macro expansions differ from platform to platform. The context menu solution could easily be extended to have “linux64 expansion:”, “macosx64 expansion:”, “android-armv7 expansion:”, … entries when they differ.

Yes, I think for the limited number of platforms we have right now and how the BlamePopup does not currently have a tab-like mechanism, this is probably the best choice.

Currently only [structured records currently have a mechanism for platform variations in the merge stage](https://github.com/mozsearch/mozsearch/blob/549ce9bf97ffba95595f710cbb8aefefe621c069/tools/src/file_format/merger.rs#L59-L84), but it does seem like we could extend the merge logic without too much trouble.  `src_data` could become a tuple of `(AnalysisSource, i_file)` so we retain knowledge of the platform it came from at merge time for source records.

A hacky thing we could do for encoding the expansions would be to have the JSON field be "expansions" which could be `[expansion1, optional platform for expansion 1 if there's only 1 expansion, expansion 2, mandatory platform for expansion 2]` so pre-merge it would just be `["the expansion"]` but if we merge then it's `["the expansion", "linux64", "other expansion", "win64,osx64,arm64"]`.  Or maybe an omitted platform is presumed to mean "all other platforms"?  Or maybe we have a more rigid format but with the C++ indexer emitting a platform of "" or "*" because it doesn't know, and the merge step fills that in.

asutherland:
I had triggered an indexer run of your branch against your most-recent (green) m-c try run earlier today that failed on the checks stage because of a change in HTML.  I re-triggered a run with https://github.com/asutherland/mozsearch-mozilla/tree/macro-expansion which disabled the limited other production checks we had, although it looks like your more recent push to the PR may have mooted the HTML changes.  In any event, that should ideally finish soon and I'll check out the results tomorrow (and maybe do some minor CSS tweaks on the server via vim over ssh, although it's also easy to update the git checkout, of course).

nicolas-guichard:
> it looks like your more recent push to the PR may have mooted the HTML changes

Indeed my most recent push was not compatible with the try push you ran the indexer against. I've triggered another indexer run with a more recent try push.

Some interesting examples:
 * [`TARGET_XPCOM_ABI`](https://dev.searchfox.org/mozilla-central/source/toolkit/xre/nsAppRunner.cpp#1236) has different values on the Android, Linux/macOS and Windows builds.
 * [`M_PI_2`](https://dev.searchfox.org/mozilla-central/source/dom/system/nsDeviceSensors.cpp#271) has a different value on macOS.
 * [`MOZ_DECLARE_REFCOUNTED_TYPENAME`](https://dev.searchfox.org/mozilla-central/source/dom/cache/Action.h#87) declares two member functions which can be highlighted/searched for from the popup. There no need to stop at declarations, cdf0d77 could be extended to the other cases too.
 * GTest [`TEST`](https://dev.searchfox.org/mozilla-central/source/netwerk/test/gtest/TestStandardURL.cpp#334) shows the weakness of my first attempt at supporting highlighting/context menu: the `Test` tokens from `::testing::Test` get a huge `data-symbols` list because `Test` is basically a substring of every other symbol declared in that expansion (the check that chooses which symbol to associate with each token is [here](https://github.com/mozsearch/mozsearch/blob/e77d45e4ceb76e790a50710384d24aa7c817a285/tools/src/format.rs#L321-L323)). This can be improved.

static/css/mozsearch.css
```diff
@@ -777,9 +792,15 @@ span[data-symbols].hovered {
...
-  z-index: 200;
+  z-index: 101;
   cursor: auto;
 }
+.blame-popup code {
+  background-color: var(--page-background);
+  display: block;
+  padding: 0.5em;
+  white-space: pre;
+}
```
asutherland, line 803: I think we need/want `color: var(--page-foreground);` in here too; otherwise for me (using the default/light color theme), identifier tokens are white text on a white background.

asutherland: Note that I vim'ed this change on the web server for the moment.

asutherland:
This is amazing!  Indeed, I hadn't appreciated that you had addressed the platform differences issue.  Wowie wow wow!

jesup:
This is great!

One question: how does this interact with conditional macros?  I.e. platform or build-type variants of macros?  For example:
`#ifdef NS_BUILD_REFCNT_LOGGING
#  define NS_LOAD_NAME_BEFORE_RELEASE(localname, _name) \
    const char* const localname = _name
#else
#  define NS_LOAD_NAME_BEFORE_RELEASE(localname, _name)
#endif
`
or

`#ifdef DEBUG
#  define NS_CHECK_FOR_RIGHT_PARTICIPANT(_ptr) _ptr->CheckForRightParticipant()
#else
#  define NS_CHECK_FOR_RIGHT_PARTICIPANT(_ptr)
#endif
`

asutherland:
> One question: how does this interact with conditional macros? I.e. platform or build-type variants of macros?

In each searchfox build using MozsearchIndexer.cpp, the expansion is archived in a "source" record at the point of macro expansion for the macro identifier token.  In the merge stage, we build an aggregate, keying by the platform only.  This works fine for your examples above because we only have one build per platform right now.  Currently we would arbitrarily clobber differing expansions if we had both a DEBUG and non-DEBUG build.  (But we don't have those.)

The current implementation would also potentially have trouble with the idiom of things like:
```cpp
#define BINDING_GOOP(x) HEADER_MAGIC(x)
#include "bindings.inc"
#define BINDING_GOOP(x) BODY_MAGIC(x)
#include "bindings.inc"
#undef BINDING_GOOP
```

We definitely have a bunch of uses like that in clang-plugin, but we don't build clang-plugin for searchfox, I think.  We also do have a built use [in xptcprivate.h](NS_LOAD_NAME_BEFORE_RELEASE), but we don't seem to end up with any content in [the raw-analysis data for xptcstubsdef.inc](https://dev.searchfox.org/mozilla-central/raw-analysis/xpcom/reflect/xptcall/xptcstubsdef.inc).  Which is maybe weird because after my changes in https://bugzilla.mozilla.org/show_bug.cgi?id=1852419 we [absolutely have semantically marked up HTML for LLVM's Attributes.inc](https://searchfox.org/llvm/source/__GENERATED__/include/llvm/IR/Attributes.inc) and quite reasonably then [the underlying analysis records](https://searchfox.org/llvm/raw-analysis/__GENERATED__/include/llvm/IR/Attributes.inc).  But it's not like this patch regresses things, there's [no semantic data on searchfox.org m-c either](https://searchfox.org/mozilla-central/source/xpcom/reflect/xptcall/xptcstubsdef.inc).

Oh, aha, so, the merge stage I guess is falling down there.  If I download the raw linux64 analysis data, I can see that on m-c trunk we do have the multiple invocations:
```json
{"loc":"00001:0","source":1,"syntax":"def,file","pretty":"file xpcom/reflect/xptcall/xptcstubsdef.inc","sym":"FILE_xpcom/reflect/xptcall/xptcstubsdef@2Einc"}
{"loc":"00001:0","target":1,"kind":"def","pretty":"xpcom/reflect/xptcall/xptcstubsdef.inc","sym":"FILE_xpcom/reflect/xptcall/xptcstubsdef@2Einc"}
{"loc":"00001:0-10","source":1,"syntax":"decl,function","type":"nsresult (void)","pretty":"function nsXPTCStubBase::Stub3","sym":"_ZN14nsXPTCStubBase5Stub3Ev"}
{"loc":"00001:0-10","source":1,"syntax":"def,function","type":"nsresult (void)","pretty":"function nsIXPTCStubBase::Stub3","sym":"_ZN15nsIXPTCStubBase5Stub3Ev"}
{"loc":"00001:0-10","source":1,"syntax":"use,macro","pretty":"macro STUB_ENTRY","sym":"M_072695aebb227fe5"}
{"loc":"00001:0-10","source":1,"syntax":"use,macro","pretty":"macro STUB_ENTRY","sym":"M_8b3657a08868bed2"}
{"loc":"00001:0-10","source":1,"syntax":"use,macro","pretty":"macro STUB_ENTRY","sym":"M_958099a08868bed2"}
{"loc":"00001:0-10","structured":1,"pretty":"nsIXPTCStubBase::Stub3","sym":"_ZN15nsIXPTCStubBase5Stub3Ev","kind":"method","parentsym":"T_nsIXPTCStubBase","overrides":[],"props":["instance","virtual","user"]}
{"loc":"00001:0-10","target":1,"kind":"decl","pretty":"nsXPTCStubBase::Stub3","sym":"_ZN14nsXPTCStubBase5Stub3Ev"}
{"loc":"00001:0-10","target":1,"kind":"def","pretty":"nsIXPTCStubBase::Stub3","sym":"_ZN15nsIXPTCStubBase5Stub3Ev"}
{"loc":"00001:0-10","target":1,"kind":"use","pretty":"STUB_ENTRY","sym":"M_072695aebb227fe5"}
{"loc":"00001:0-10","target":1,"kind":"use","pretty":"STUB_ENTRY","sym":"M_8b3657a08868bed2"}
{"loc":"00001:0-10","target":1,"kind":"use","pretty":"STUB_ENTRY","sym":"M_958099a08868bed2"}
```

And for the most recent patch try run:
```json
{"loc":"00001:0","source":1,"syntax":"def,file","pretty":"file xpcom/reflect/xptcall/xptcstubsdef.inc","sym":"FILE_xpcom/reflect/xptcall/xptcstubsdef@2Einc"}
{"loc":"00001:0","target":1,"kind":"def","pretty":"xpcom/reflect/xptcall/xptcstubsdef.inc","sym":"FILE_xpcom/reflect/xptcall/xptcstubsdef@2Einc"}
{"loc":"00001:0-10","source":1,"expansions":"asm(\".section\t\\\".text\\\"\\n\\t\"\n    \".align\t2\\n\\t\"\n    \".if\t\"\n    \"3\"\n    \" < 10\\n\\t\"\n    \".globl\t_ZN14nsXPTCStubBase5Stub\"\n    \"3\"\n    \"Ev\\n\\t\"\n    \".hidden\t_ZN14nsXPTCStubBase5Stub\"\n    \"3\"\n    \"Ev\\n\\t\"\n    \".type\t_ZN14nsXPTCStubBase5Stub\"\n    \"3\"\n    \"Ev,@function\\n\"\n    \"_ZN14nsXPTCStubBase5Stub\"\n    \"3\"\n    \"Ev:\\n\\t\"\n    \".elseif\t\"\n    \"3\"\n    \" < 100\\n\\t\"\n    \".globl\t_ZN14nsXPTCStubBase6Stub\"\n    \"3\"\n    \"Ev\\n\\t\"\n    \".hidden\t_ZN14nsXPTCStubBase6Stub\"\n    \"3\"\n    \"Ev\\n\\t\"\n    \".type\t_ZN14nsXPTCStubBase6Stub\"\n    \"3\"\n    \"Ev,@function\\n\"\n    \"_ZN14nsXPTCStubBase6Stub\"\n    \"3\"\n    \"Ev:\\n\\t\"\n    \".elseif    \"\n    \"3\"\n    \" < 1000\\n\\t\"\n    \".globl     _ZN14nsXPTCStubBase7Stub\"\n    \"3\"\n    \"Ev\\n\\t\"\n    \".hidden    _ZN14nsXPTCStubBase7Stub\"\n    \"3\"\n    \"Ev\\n\\t\"\n    \".type      _ZN14nsXPTCStubBase7Stub\"\n    \"3\"\n    \"Ev,@function\\n\"\n    \"_ZN14nsXPTCStubBase7Stub\"\n    \"3\"\n    \"Ev:\\n\\t\"\n    \".else\\n\\t\"\n    \".err\t\\\"stub number \"\n    \"3\"\n    \" >= 1000 not yet supported\\\"\\n\\t\"\n    \".endif\\n\\t\"\n    \"movl\t$\"\n    \"3\"\n    \", %eax\\n\\t\"\n    \"jmp\tSharedStub\\n\\t\"\n    \".if\t\"\n    \"3\"\n    \" < 10\\n\\t\"\n    \".size\t_ZN14nsXPTCStubBase5Stub\"\n    \"3\"\n    \"Ev,.-_ZN14nsXPTCStubBase5Stub\"\n    \"3\"\n    \"Ev\\n\\t\"\n    \".elseif\t\"\n    \"3\"\n    \" < 100\\n\\t\"\n    \".size\t_ZN14nsXPTCStubBase6Stub\"\n    \"3\"\n    \"Ev,.-_ZN14nsXPTCStubBase6Stub\"\n    \"3\"\n    \"Ev\\n\\t\"\n    \".else\\n\\t\"\n    \".size\t_ZN14nsXPTCStubBase7Stub\"\n    \"3\"\n    \"Ev,.-_ZN14nsXPTCStubBase7Stub\"\n    \"3\"\n    \"Ev\\n\\t\"\n    \".endif\");","syntax":"use,macro","pretty":"macro STUB_ENTRY","sym":"M_072695aebb227fe5"}
{"loc":"00001:0-10","source":1,"expansions":"virtual nsresult\nStub3() = 0;","syntax":"use,macro","pretty":"macro STUB_ENTRY","sym":"M_8b3657a08868bed2"}
{"loc":"00001:0-10","source":1,"expansions":"virtual nsresult\nStub3() override;","syntax":"use,macro","pretty":"macro STUB_ENTRY","sym":"M_958099a08868bed2"}
{"loc":"00001:0-10","source":1,"syntax":"decl,function","type":"nsresult (void)","pretty":"function nsXPTCStubBase::Stub3","sym":"_ZN14nsXPTCStubBase5Stub3Ev"}
{"loc":"00001:0-10","source":1,"syntax":"def,function","type":"nsresult (void)","pretty":"function nsIXPTCStubBase::Stub3","sym":"_ZN15nsIXPTCStubBase5Stub3Ev"}
{"loc":"00001:0-10","structured":1,"pretty":"nsIXPTCStubBase::Stub3","sym":"_ZN15nsIXPTCStubBase5Stub3Ev","kind":"method","parentsym":"T_nsIXPTCStubBase","overrides":[],"props":["instance","virtual","user"]}
{"loc":"00001:0-10","target":1,"kind":"decl","pretty":"nsXPTCStubBase::Stub3","sym":"_ZN14nsXPTCStubBase5Stub3Ev"}
{"loc":"00001:0-10","target":1,"kind":"def","pretty":"nsIXPTCStubBase::Stub3","sym":"_ZN15nsIXPTCStubBase5Stub3Ev"}
{"loc":"00001:0-10","target":1,"kind":"use","pretty":"STUB_ENTRY","sym":"M_072695aebb227fe5"}
{"loc":"00001:0-10","target":1,"kind":"use","pretty":"STUB_ENTRY","sym":"M_8b3657a08868bed2"}
{"loc":"00001:0-10","target":1,"kind":"use","pretty":"STUB_ENTRY","sym":"M_958099a08868bed2"}
```

As is made somewhat clear by the JSON, we [define the symbol identifier for a macro as a hash based on the location of its definition](https://searchfox.org/mozilla-central/rev/b580e3f77470b2337bc8ae032b58a85c11e66aba/build/clang-plugin/mozsearch-plugin/MozsearchIndexer.cpp#2126-2127) which will inherently vary for cases like the above where the influence of a differing macro is felt by a macro definition happening at a different point.  This means we could potentially also key by the macro symbol identifier as well for those cases; this could be nice because by keying on the symbol we can also provide a "go to the definition that caused this expansion" part to the UI (eventually).

But I think things would potentially fall down where the macro we are expanding effectively has its expansion parameterized by some other macro which does not impact the hash used for the macro being expanded.  At that point we'd also need to check if the expansion strings are identical, and if they're not, maybe just put them in a list.  Or better yet, discard the expansion and replace it with a string like: "If you wrote this nightmarish set of macros, you clearly thought you understood what was going on; [enjoy the whirlwind!](https://en.wiktionary.org/wiki/sow_the_wind,_reap_the_whirlwind)"

asutherland:
I did a brief investigation to try and figure out how things might be going wrong in the merge-analyses step but it's not jumping out at me.  I see the `xpcom/reflect/xptcall/xptcstubsdef.inc` entry in the per-platform lists like `analysis-files-linux64.list` and also in the aggregate `xpcom/reflect/xptcall/xptcstubsdef.inc`.  And if I download the linux64 and armv7 files and run merge-analyses locally in docker with this branch checked out, the merge seems to run correctly and without problems.

Unfortunately, because we don't do anything [like we do to set the JOBLOG_PATH for parallel for output-file](https://github.com/mozsearch/mozsearch/blob/549ce9bf97ffba95595f710cbb8aefefe621c069/scripts/output.sh#L24) for our [invocation of merge-analyses via parallel](https://github.com/mozsearch/mozsearch-mozilla/blob/4c6336115ee1a74c61ca6afc29a41cd9fdb6c45d/shared/process-gecko-analysis.sh#L43), I think the parallel output/joblogs ended up in [the TMPDIR of /mnt/index-scratch/tmp](https://github.com/mozsearch/mozsearch/blob/549ce9bf97ffba95595f710cbb8aefefe621c069/infrastructure/aws/main.sh#L104) which was the indexer's SSD and which I don't think we try to move to the persistent volume in the success case.  We probably should be trying to capture the parallel output of those runs in a similar fashion.

nicolas-guichard:
I have rebased this and made some improvements:
- The mapping of tokens inside macro expansions is now way more robust than my initial hack based on `item.pretty.contains(&input[token.start..token.end])`. Instead I record the location of each identifier in the macro expansion, apply the code reformatting to them and add them to the source records in a new `inExpansionAt` field.
- When the expansion is the same on every platform display only `Expansion: …` instead of `Expansion plat0 plat1 plat2 plat3: …`.

The code in format.rs could probably be improved to be less redundant and more digest.

nicolas-guichard:
Also as you mentioned in #758 I didn't share screenshots initially, here goes until I deploy this to kdab.searchfox.org:
Context menu with per-platform expansions
Platform-specific expansion
Context menu in expansion

asutherland:
Very exciting about implementing high fidelity expansion!  Given @arai-a's recent introduction of a an amazing [webtest](https://github.com/mozsearch/mozsearch/blob/master/docs/webtest.md) mechanism ([existing tests](https://searchfox.org/mozsearch/source/tests/webtest)), it would be great if we can add some basic webtest regression tests.

I will wait on the kdab.searchfox.org update to dig into the review because I find having that context available can be really handy.  That will also be helpful since it turns out we can fairly easily get an accessibility review from the accessibility team, albeit via slack (~1week ETA).  Also, @theres-waldo and @arai-a, please also feel free to do review passes if you're interested as you both have very relevant expertise!

arai-a (commented):
Excellent!
This will help investigating code with very long macro.
I'm looking forward to see this on production!

nicolas-guichard:
This new version has:
- the expansion context menu entries behind feature flag
- a basic webtest that verifies the expansion popup can be opened and interacted with
- the max-width moved from the whole context menu to only the expansion -related entries
- I've added `BlameStripHoverHandler.keepVisible = true;` to the macro display action otherwise moving the mouse in/out of the popup would hide the popup

nicolas-guichard:
Trying to index a Try push revealed I didn't correctly handle cases like `#include SOME_DEFINE`. Now fixed, I'll push to Try again.

nicolas-guichard:
The Try run and the indexing was successful and can be tested at kdab.searchfox.org.

Updating my examples from [my previous comment](https://github.com/mozsearch/mozsearch/pull/679#issuecomment-1866207558):
- [TARGET_XPCOM_ABI](https://kdab.searchfox.org/mozilla-central/source/toolkit/xre/nsAppRunner.cpp#1224) has different values on the Android/iOS, Linux/macOS and Windows builds.
- [M_PI_2](https://kdab.searchfox.org/mozilla-central/source/dom/system/nsDeviceSensors.cpp#271) has a different value on macOS/iOS.
- [MOZ_DECLARE_REFCOUNTED_TYPENAME](https://kdab.searchfox.org/mozilla-central/source/dom/cache/Action.h#87) declares two member functions which can be highlighted/searched for from the popup.
- GTest [TEST](https://kdab.searchfox.org/mozilla-central/source/netwerk/test/gtest/TestHeaders.cpp#5) works and unlike my first attempt the symbols are properly distinguished. Note: The expansions of the `TEST` macro in my previous example [TestStandardURL.cpp](https://kdab.searchfox.org/mozilla-central/source/netwerk/test/gtest/TestStandardURL.cpp#28) seems to be broken. I traced that to the `replace_aliases` step because `"URL_"` appears in the macro expansions.
- [`#include MACRO`](https://kdab.searchfox.org/mozilla-central/source/gfx/graphite2/src/inc/Main.h#33) is supported

arai-a:
Excellent!
This will really help the workflow when investigating the code with full of macro :)

Then, I have some requests about the macro expansion popup.
There are some very long macro in SpiderMonkey (e.g. 600+ lines), and the popup content can overflow vertically.
Can we set a max-height with a vertical scroll bar?
https://kdab.searchfox.org/mozilla-central/source/js/src/frontend/ParserAtom.cpp#39
popup-long

Also, there seems to be some issue with content width handling in the popup, where the
text can go outside of the padding, or something along that line
(the trailing comma hits the right side of the box)
https://kdab.searchfox.org/mozilla-central/source/js/src/jit/CacheIR.cpp#82
padding


Also, this is optional, but the "Go to definition" list in the context menu now contains symbols inside expanded macro,
and the first item is no longer the macro itself (`FOR_EACH_COMMON_PROPERTYNAME` in this case).
Can it be put as a first item?
go-to-definition-order

Fwiw, here's some list of some long macros that's heavily used in SpiderMonkey code, which will highly benefit from this feature:
  * https://kdab.searchfox.org/mozilla-central/source/js/src/vm/Opcodes.h#257
  * https://kdab.searchfox.org/mozilla-central/source/js/src/vm/CommonPropertyNames.h#23
  * https://kdab.searchfox.org/mozilla-central/source/js/src/frontend/ParseNode.h#63
  * https://kdab.searchfox.org/mozilla-central/source/js/src/gc/AllocKind.h#61-115
  * https://kdab.searchfox.org/mozilla-central/source/js/public/ProtoKey.h#68

and some more cases that may be interesting:
  * https://kdab.searchfox.org/mozilla-central/source/__GENERATED__/xpcom/ds/nsGkAtomList.h#8 (This header is included multiple times with different `GK_ATOM` macro.  I'm not sure how it should be handled, but right now only one expansion from https://kdab.searchfox.org/mozilla-central/source/xpcom/ds/nsGkAtoms.h#178 is shown)
  * https://kdab.searchfox.org/mozilla-central/source/parser/htmlparser/nsHTMLTagList.h#52 (Same as above, this file is included multiple times with different `HTML_TAG` macro)

nicolas-guichard:
Thanks for the feedback!

> There are some very long macro in SpiderMonkey (e.g. 600+ lines), and the popup content can overflow vertically.
> Can we set a max-height with a vertical scroll bar?

> Also, there seems to be some issue with content width handling in the popup, where the text can go outside of the padding, or something along that line

I've set `max-height: 80vh;` and `overflow-y: scroll;` to `.blame-popup` and moved the `max-width: 600px;` to only impact the actual blame entries so the width adapts to the actual code (also vimed on the server instead of relaunching the indexer).

nicolas-guichard:
> Also, this is optional, but the "Go to definition" list in the context menu now contains symbols inside expanded macro,
and the first item is no longer the macro itself (FOR_EACH_COMMON_PROPERTYNAME in this case).
> Can it be put as a first item?

Good point. This one I can't just vim on the server since the information about which symbol is the expanded one is lost in `format.rs`.
In fact I'm wondering if the inner symbols should be listed in the main context menu at all? The user can access them by opening the expansion after all.

Also I realize they are now referenced in searches, for instance searching for `symbol:E_<T_js::WellKnownAtomId>_Async` returns more result [now](https://kdab.searchfox.org/mozilla-central/search?q=symbol:E_%3CT_js%3A%3AWellKnownAtomId%3E_Async&redirect=false) than it did [before](https://searchfox.org/mozilla-central/search?q=symbol:E_%3CT_js%3A%3AWellKnownAtomId%3E_Async&redirect=false). I think we want to keep that even if I cleanup the context menus.

arai-a:
> I've set `max-height: 80vh;` and `overflow-y: scroll;` to `.blame-popup` and moved the `max-width: 600px;` to only impact the actual blame entries so the width adapts to the actual code (also vimed on the server instead of relaunching the indexer).

Thank you for the quick fix!
Now I can see the entire content of the expanded macro for very long case.
This is really helpful when working with macro-heavy code.

(style-nit) It might be nice if the scroll bar is added to the code block itself, instead of (or in addition to?) the popup, so that the thick border around the code block is always shown.

> In fact I'm wondering if the inner symbols should be listed in the main context menu at all? The user can access them by opening the expansion after all.

yeah, if there are only 2 or 3, showing them in the same context menu (possibly in separate section) might make sense,
but otherwise it would be better hiding them, to make other menu items for the macro itself (go to definition, search, etc) more easily accessible.

> Also I realize they are now referenced in searches, for instance searching for `symbol:E_<T_js::WellKnownAtomId>_Async` returns more result [now](https://kdab.searchfox.org/mozilla-central/search?q=symbol:E_%3CT_js%3A%3AWellKnownAtomId%3E_Async&redirect=false) than it did [before](https://searchfox.org/mozilla-central/search?q=symbol:E_%3CT_js%3A%3AWellKnownAtomId%3E_Async&redirect=false). I think we want to keep that even if I cleanup the context menus.

yes, showing those definitions and uses in the search is good improvement.


Also, currently the `#blame-popup` element is placed outside of `body` element. is it intentional (maybe there's some use case) ?
This seems to be causing inconsistent behavior, for example, the text content inside it is not selected by `Cmd+A`.
If this is not intentional behavior, it would be nice to put the element inside body.

text selection-and-element-position

asutherland:
This is so amazing!  Thank you!

> Good point. This one I can't just vim on the server since the information about which symbol is the expanded one is lost in `format.rs`. In fact I'm wondering if the inner symbols should be listed in the main context menu at all? The user can access them by opening the expansion after all.

I agree that the currently approach is likely too voluminous without additional heuristics that are probably out-of-scope for this initial implementation.  For my team, at least, [ActorsParent.cpp](https://kdab.searchfox.org/mozilla-central/rev/df781a59fbb92e3928d3e4d03ec79dc6910e7da8/dom/indexedDB/ActorsParent.cpp#674) it arguably is better to not list them at all if it's all-or-nothing:
image

I think potentially what we might end up wanting to continue to expose is definitions and declarations made within the macro expansion that aren't locals/temporaries.  However, I think I'd also argue that the context-menu isn't really the right place for them, it's just the only affordance we have to expose that information right now.  Probably a better affordance would be something similar to the "inlay type hints" used by VS code language servers/similar.  For example, our XPCOM interface declaration macros like [NS_DECL_NSIINPUTSTREAM](https://kdab.searchfox.org/mozilla-central/source/xpcom/io/nsMultiplexInputStream.cpp#51) emit a bunch of declarations, which is notable.  (Another place they could go would be in the fancy sidebar we're evolving towards.)

> Also I realize they are now referenced in searches, for instance searching for `symbol:E_<T_js::WellKnownAtomId>_Async` returns more result [now](https://kdab.searchfox.org/mozilla-central/search?q=symbol:E_%3CT_js%3A%3AWellKnownAtomId%3E_Async&redirect=false) than it did [before](https://searchfox.org/mozilla-central/search?q=symbol:E_%3CT_js%3A%3AWellKnownAtomId%3E_Async&redirect=false). I think we want to keep that even if I cleanup the context menus.

Yeah, 100% agreed with @arai-a that this is great.

asutherland:
> Also, currently the `#blame-popup` element is placed outside of `body` element. is it intentional (maybe there's some use case) ? This seems to be causing inconsistent behavior, for example, the text content inside it is not selected by `Cmd+A`. If this is not intentional behavior, it would be nice to put the element inside body.

It should be?  It looks like @emilio did this perhaps as an optimization to avoid DOM churn when he overhauled the class; currently we do it [here](https://searchfox.org/mozsearch/rev/1f5338c5137104176866feabc9fe633054eb07ab/static/js/blame.js#13):
```js
document.documentElement.appendChild(this.popup);
```

where [before that we used the blame strip as the parent](https://searchfox.org/mozsearch/rev/9e3930a9b461709ee0339a5b03684ed278a0cda9/static/js/blame.js#3-4,100,106):
```js
// The .blame-strip element for which blame is currently being displayed.
var blameElt;
...
    var parent = blameElt.parentNode;
...
    parent.appendChild(popupBox);
```

emilio:
Yeah from 4891238289b3d8774efb68e2f5bf9e8cb5e690ce's commit message:

> Right now the blame popup causes pretty expensive reflows, because it's hosted inside the line strip.

That said, probably putting it on `<body>` rather than `<html>` is ok.

asutherland (approved):
This is super awesome!  I think there's a few small things to burn down from the discussion, but I expect @arai-a will do a review sign-off that can pair with that.  r=asuth for the core implementation stuff.

nicolas-guichard:
> (style-nit) It might be nice if the scroll bar is added to the code block itself, instead of (or in addition to?) the popup, so that the thick border around the code block is always shown.

I've moved the max-height & scrollbar to the `<code>` element.

nicolas-guichard:
@asutherland I'm going to ask for another round of review for the core stuff, as I've made changes to accommodate the ordering of go-to vs expansion and the multiple expansions with different macro definitions use-case.

For instance if we have:
```
#ifdef LINUX
#define MACRO bla bla bla
#else
#define MACRO bla bla bla
#endif

MACRO
```
then instead of getting:
```
"expandsTo": {"plat0": "exp0", "plat1": "exp1", "plat2": "exp2"},
and
"inExpansionAt": {"plat0": [1, 2, 3], "plat1": [1, 2, 3], "plat2": [1, 2, 3]},
```
we keep which expansion is related to which `#define`:
```
"expandsTo": {"sym0: {"plat0": "exp0", "plat0": "exp2"}, "sym1": {"plat1": "exp1"}},
and
"inExpansionAt": {"sym0": {"plat0": [1, 2, 3], "plat2": [1, 2, 3]}, "sym1": {"plat1": [1, 2, 3]}},
```

Basically it now looks like this:
image

The idea is to have:
- go to macro #define 1
- expansion related to macro #define 1
- go to macro #define 2
- expansion related to macro #define 2

When opening the expansion we now know which #define it is related to:
image

Of course we would want to replace the `M_gibberish` identifier by something else, maybe a clickable `<pretty name> at <file>:<line>`?


I'll launch a Try run with the new clang-plugin-related changes.

asutherland:
> @asutherland I'm going to ask for another round of review for the core stuff, as I've made changes to accommodate the ordering of go-to vs expansion and the multiple expansions with different macro definitions use-case.

Really love the UX here!

> we keep which expansion is related to which `#define`:
> 
> ```
> "expandsTo": {"sym0: {"plat0": "exp0", "plat0": "exp2"}, "sym1": {"plat1": "exp1"}},
> and
> "inExpansionAt": {"sym0": {"plat0": [1, 2, 3], "plat2": [1, 2, 3]}, "sym1": {"plat1": [1, 2, 3]}},
> ```

Oh, neat!  Thanks for the illustrative examples!

> Of course we would want to replace the `M_gibberish` identifier by something else, maybe a clickable `<pretty name> at <file>:<line>`?

Yes, that sounds great, although I think the key thing is just to have the pretty identifier instead of the raw symbol if that turns out to be tricky.  (Although I think the SYM_INFO should make it doable... which reminds me I should probably try and do that for the diagram limits-hit listing, which currently exposes raw symbols.)

nicolas-guichard:
> > Of course we would want to replace the `M_gibberish` identifier by something else, maybe a clickable `<pretty name> at <file>:<line>`?
> 
> Yes, that sounds great, although I think the key thing is just to have the pretty identifier instead of the raw symbol if that turns out to be tricky. (Although I think the SYM_INFO should make it doable... which reminds me I should probably try and do that for the diagram limits-hit listing, which currently exposes raw symbols.)

I went with `Expansion of <pretty> as defined at <file>#<line> on <platform>:` for now.
`<pretty>` is hightlightable/clickable and brings up the context menu like any symbol and `<file>#<line>` is clickable too:
image

The Try push is finished and I just launched a just-mc indexing job with that.

nicolas-guichard:
The [`nsGkAtomList.h`](https://kdab.searchfox.org/mozilla-central/source/__GENERATED__/xpcom/ds/nsGkAtomList.h) case is better but not entirely what I expected: it seems there is a jumpref issue and while the analysis for [`nsGkAtomList.h`](https://kdab.searchfox.org/mozilla-central/raw-analysis/__GENERATED__/xpcom/ds/nsGkAtomList.h) itself looks good and has all the data we want most “go to definition” context menu entries are missing and only 1 remain. Same for [`HTML_TAG` in nsHTMLTagList.h](https://kdab.searchfox.org/mozilla-central/source/parser/htmlparser/nsHTMLTagList.h#52).

For [`HTML_HTMLELEMENT_TAG` in nsHTMLTagList.h](https://kdab.searchfox.org/mozilla-central/source/parser/htmlparser/nsHTMLTagList.h#52) however, since the `HTML_HTMLELEMENT_TAG` macro itself is only defined in a single place there is still only 1 single expansion shown. I could store all the symbols of macros involved in a given expansion instead of only the top-most one to have the data and display “expansion of `HTML_HTMLELEMENT_TAG` using the definition of `HTML_TAG` from `file1.cpp#24`” vs “expansion of `HTML_HTMLELEMENT_TAG` using the definition of `HTML_TAG` from `file2.cpp#42`”.

Also my logic for hiding inner symbols from the main context menu didn't take into account inner macros (which don't have an in-macro location since they are themselves expanded). Leading to this for instance:
image
I'll remove inner macros from there.

nicolas-guichard:
I've fixed the issue about nested macros like `HTML_HTMLELEMENT_TAG` by tracking all macro dependencies in macro expansions.
This gives us:
image

I've cleaned up the context menus but by using the inner context menu you can still access the dependencies:
image
In that case the “Go to definition of YO_ATOM" entry will either point to atom_magic.h#10 or atom_magic.h#16 depending on the expansion that was opened.

nicolas-guichard:
An indexer run with the latest changes just finished. Please give it a (final?) try.

Known issues remaining, but should be fixed in other PRs IMO:
- jumpref data is lacking in some cases, like [`nsGkAtomList.h`](https://kdab.searchfox.org/mozilla-central/source/__GENERATED__/xpcom/ds/nsGkAtomList.h) or [`nsHTMLTagList.h`](https://kdab.searchfox.org/mozilla-central/source/parser/htmlparser/nsHTMLTagList.h#52)
- the `BlamePopup` lives outside the `<body>` tag

Apart from that I think I fixed everything else we discussed, notably:
- some stylistic changes (max-height & scrollbar around expanded code)
- macro expansions are put below the “Go to definition” of the macro that was expanded (modulo the jumpref issue, in that case they are still shown above, same for `-DDEFINES`)
- the context menu doesn't have ∞ entries anymore, instead inner symbols are to be found in the expansion and inner macros in the context menu of the title, but they still count towards the expanded token when searching
- expansion of multiple macros defines in a single location ([`nsGkAtomList.h`](https://kdab.searchfox.org/mozilla-central/source/__GENERATED__/xpcom/ds/nsGkAtomList.h) case) is supported, even indirectly ([`HTML_HTMLELEMENT_TAG` in `nsHTMLTagList.h`](https://kdab.searchfox.org/mozilla-central/source/parser/htmlparser/nsHTMLTagList.h#52) case)

asutherland (approved):
This is going to be really transformative!  Thank you!  Let's wait on @arai-a's r+ too.

arai-a (approved):
Excellent!

One thing I noticed is that in the `nsGkAtomList.h`'s case, the "Go to definition" item lives in between expansions.
Is it the same issue as jumpref?

anyway, it's not necessary to address it here.

https://kdab.searchfox.org/mozilla-central/source/__GENERATED__/xpcom/ds/nsGkAtomList.h#8
go-to-definition-position

nicolas-guichard:
> One thing I noticed is that in the `nsGkAtomList.h`'s case, the "Go to definition" item lives in between expansions.
> Is it the same issue as jumpref?

Yes this is the jumpref issue: there should be 5 “Go to definition of GK_ATOM” jumps here with each expansion below the relevant one, but we only get 1 jump.
This leads to another difference: if you open the expansion below “Go to definition of GK_ATOM” it will show “Expansion of `GK_ATOM` on ${platforms}:” and you can click on `GK_ATOM` to get the “Go to definition” for every involved macro; for the other you only get “Expansion on ${platforms}:”.

PR #680, opened 2024-01-04
Made available as OpenSearch provider

coderonline:
This is supposed to add the possibility to add searchfox to the list of search providers within the browser. This is a suggestion and only tested off-tree right now. Please let me know if you like the idea of it so that I know if it makes sense to invest more time into it.

With this change the search box is supposed to look like this:
searchbox

### DIFF
--- a/static/mozsearch.xml
+++ b/static/mozsearch.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+  <OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
+  <ShortName>SearchFox</ShortName>
+  <Description>SearchFox code search</Description>
+  <InputEncoding>UTF-8</InputEncoding>
+  <Image width="16" height="16" type="image/x-icon">https://searchfox.org/mozilla-central/static/icons/search.png</Image>
+  <Query role="example" searchTerms="BrandFullName" />
+  <Url type="text/html" template="https://searchfox.org/mozilla-central/search?q={searchTerms}&amp;path=&amp;case=false&amp;regexp=false"></Url>
+</OpenSearchDescription>
--- a/tools/templates/header_search.liquid
+++ b/tools/templates/header_search.liquid
@@ -14,6 +14,7 @@
   <link href="/{{tree}}/static/css/mozsearch.css" rel="stylesheet" media="screen"/>
   <link href="/{{tree}}/static/css/icons.css" rel="stylesheet" media="screen" />
   <link href="/{{tree}}/static/css/font-icons.css" rel="stylesheet" media="screen" />
+  <link href="/{{tree}}/static/searchfox.xml" rel="search" title="{{ title }}" type="application/opensearchdescription+xml"/>
 </head>
 
 <body>


asutherland:
I think the expected use case right now is that people will use keyword bookmarks[1] to add per-tree searches.  Like I use "sf" for searchfox so I can do `sf query-payload` via my URL bar and "sfwk" for the "wubkat" tree.  That said, I don't believe we have documentation on searchfox.org mentioning the existence of this mechanism or how to do it, so discoverability is bad, whereas OpenSearch conceptually is discoverable, as you demonstrate with your screenshot.  (Although it seems like the icon badge only shows up in Firefox if the search bar has been customized onto the toolbar?)

It would definitely be nice to do better in terms of helping people learn how to more efficiently use searchfox.  https://github.com/mozsearch/mozsearch-mozilla/blob/master/help.html provides the contents seen on the root of searchfox.org.  Adding documentation there might be the most pragmatic way to go there.

I should note that @aminomancer has created an add-on version at https://addons.mozilla.org/en-US/firefox/addon/searchfox-engine/.  @aminomancer, do you have thoughts on the utility of hosting such an XML file on searchfox.org vs trying to document the keyword bookmark flow?  cc @emilio 

1: By right-clicking in the "Search mozilla-central" (or other tree) search field and choosing "Add a Keyword for this search..." in Firefox.

aminomancer:
> I think the expected use case right now is that people will use keyword bookmarks[1] to add per-tree searches. Like I use "sf" for searchfox so I can do `sf query-payload` via my URL bar and "sfwk" for the "wubkat" tree. That said, I don't believe we have documentation on searchfox.org mentioning the existence of this mechanism or how to do it, so discoverability is bad, whereas OpenSearch conceptually is discoverable, as you demonstrate with your screenshot. (Although it seems like the icon badge only shows up in Firefox if the search bar has been customized onto the toolbar?)

OpenSearch has a couple of other UI affordances too. The urlbar results popup will have an add search engine button at the bottom, where the other one-off search engines are. Right clicking in the urlbar will also show a menuitem that adds the engine. And there's also a special urlbar result type, "Search with {current site}" - I think it creates a temporary search engine and passes the search terms to that.

> I should note that @aminomancer has created an add-on version at [addons.mozilla.org/en-US/firefox/addon/searchfox-engine](https://addons.mozilla.org/en-US/firefox/addon/searchfox-engine/). @aminomancer, do you have thoughts on the utility of hosting such an XML file on searchfox.org vs trying to document the keyword bookmark flow? cc @emilio

Both seem worthwhile to me. Keyword bookmarks can do things SearchEngines can't, and vice versa. SearchEngines' integration in the UI is a lot better, with buttons in the urlbar and an indicator label while you're in a search mode. When I switch to searchfox mode and start typing a query, it shows me history results of URLs I visited on that domain. So I could type "urlbar" and the first result would be `UrlbarView.sys.mjs` instead of just `?q=urlbar`.

The biggest advantage of SearchEngines is the support for search suggestions. But searchfox doesn't have a suggestions API, so we can't take advantage of that. Someday though.

Anyway, making the default OpenSearch description search mozilla-central seems okay, since the searchbar on the root also searches mozilla-central. That probably covers the use cases of most users. Per-tree descriptions could be made, but it seems pretty tedious.

The web server could be configured to dynamically generate an OpenSearch description for each tree, and that seems to dovetail nicely with how the site is set up. That way if you're on the root or mozilla-central, it'll show the OSD for m-c, but if you're on mozilla-beta, it'll show the OSD for m-b. Rather than a static XML file, use `/{{tree}}/searchfox.xml`, and configure the server to output an appropriate XML file for that request.

static/mozsearch.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+  <OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
+  <ShortName>SearchFox</ShortName>
+  <Description>SearchFox code search</Description>
+  <InputEncoding>UTF-8</InputEncoding>
+  <Image width="16" height="16" type="image/x-icon">https://searchfox.org/mozilla-central/static/icons/search.png</Image>
+  <Query role="example" searchTerms="BrandFullName" />
+  <Url type="text/html" template="https://searchfox.org/mozilla-central/search?q={searchTerms}&amp;path=&amp;case=false&amp;regexp=false"></Url>
```
aminomancer, line 8:
```suggestion
  <Url type="text/html" template="https://searchfox.org/mozilla-central/search?q={searchTerms}"></Url>
```

These seem unnecessary. Am I overlooking something?

asutherland: For searchfox's caching of dynamic queries it arguably is preferable to use the full encoded form for the extra `&path=&case=false&regexp=false` suffix since the dynamically updated URL from the search form will include the suffix.  That said, because dynamic queries request JSON instead of HTML and the cache `Vary`s on that, it means that this really only matters for the case where 1) the person using OpenSearch looks for a query Foo, 2) some other person types in the search in the form generating an equivalent URL, 3) the person from 2 shares that URL with some other person who accesses it directly.  The person in case 1 will have generated the `Accept: text/html` payload already for the URL shared by the person from case 2, which will generate an instant cache hit for the person in case 3.  (Case 2 would only generate a cache hit for `Accept: application/json`.)

asutherland:
> The biggest advantage of SearchEngines is the support for search suggestions. But searchfox doesn't have a suggestions API, so we can't take advantage of that. Someday though.

There is, for example "WorkerP", https://searchfox.org/mozilla-central/complete/WorkerP [generated here](https://github.com/mozsearch/mozsearch/blob/549ce9bf97ffba95595f710cbb8aefefe621c069/tools/src/bin/web-server.rs#L245-L251).  Presumably that would need to be adapted.  (And in particular, one would only want to use the surfaced "id" values and without any prefixing, because the default search is usually what people always want.)

(I tend to favor the awesomebar results that come up when using the keyword bookmark, as you note happens.)

> The web server could be configured to dynamically generate an OpenSearch description for each tree, and that seems to dovetail nicely with how the site is set up. That way if you're on the root or mozilla-central, it'll show the OSD for m-c, but if you're on mozilla-beta, it'll show the OSD for m-b. Rather than a static XML file, use `/{{tree}}/searchfox.xml`, and configure the server to output an appropriate XML file for that request.

Ah, yeah, I think we would definitely want to do this.  I think the relevant mechanisms as it relates to this are:
- [output.sh invoking per-tree templates like "render help"](https://github.com/mozsearch/mozsearch/blob/549ce9bf97ffba95595f710cbb8aefefe621c069/scripts/output.sh#L82-L85)
- which depends on [the "render" pipeline command referencing the liquid template and invoking it](https://github.com/mozsearch/mozsearch/blob/549ce9bf97ffba95595f710cbb8aefefe621c069/tools/src/cmd_pipeline/cmd_render.rs)
- [web-server-setup.sh](https://github.com/mozsearch/mozsearch/blob/master/infrastructure/web-server-setup.sh) and its logic around "DEFAULT_TREE_NAME" putting files in place and handling the default-tree special case.

We potentially would be able to add regression checks for this too, like [here's the check input for the generated search.html file](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/inputs/web/templates/search_template__html) and [here's the resulting output](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/snapshots/web/templates/check_glob%40search_template__html.snap)

PR #684, opened 2024-01-30
Bug 1773165 - Modest diagramming improvements

asutherland:
Under the auspices of https://bugzilla.mozilla.org/show_bug.cgi?id=1773165#c3 I'm landing most of my outstanding hobby stack.  I've given this a review skim myself to make sure nothing weird rides along.  Most changes are in the "query" endpoint space as it relates to diagrams and pipeline-server infrastructure.

The only change to supported functionality muscle memory are:
- This does put icons back into place for the context menu which were accidentally removed in the past, with new icons added.  This was previously discussed with UX lead :smaug in the channel on August 15th, 2023 and in general there hasn't been any feedback about the differences when linking people to the asuth.searchfox.org server that has been serving the changes continually since then.
- Some minor enhancements to context menu behavior as it relates to binding slots and eliding implicit constructors that otherwise can overwhelm the menu.

In terms of the pre-alpha diagramming functionality, all diagramming modes are exposed by the context-menu UI, but only if the diagramming functionality is enabled.  (This is already behind a gate.)  The class-diagram functionality does depend on changes made to MozsearchIndexer to ensure we provide fully qualified type information (ex: all namespaces included, not just what the user typed).  (As do some of the ontology mapping heuristics.)  Right now this is parsed from pretty-space inside crossref in a way I'm not super proud of, but that works and has test coverage.  The ideal approach is almost certainly to leverage clang's understanding of the types instead, but the current approach works and it's likely the more clever semantic approach will involve a fair amount of work.

There are also some changes to the clang indexer output to provide information about function call arguments.  This is part of a "dataflow for babies" experiment which I plan to continue iterating on in my hobby stack and where it's advantageous to have this data available on the m-c searchfox jobs so I can iterate more easily.  While the higher level dataflow goals are longer term, the ability to potentially facet based on what's passed to argument calls is likely something that can bear fruit in the medium term.  (I am counting hyperblame as short term.)


asutherland:
The utc22 indexing run fully completed (indexer finished, web server became active) so there should be no phase problems with landing this now and we do want this in place before the clang indexer changes show up.  Landing.

PR #698, opened 2024-03-13
Bug 1885069 - pin tree-sitter-kotlin to retain field names from parser

asutherland:


asutherland:
local tests are happy

PR #699, opened 2024-03-18
Bug 1779939 - update git-cinnabar to 0.6.3

asutherland:
With the android repo merge into mozilla-central, git fetch on central was throwing.  This may potentially have been a transient issue with the server, but since there wasn't a reason to be on an old version of cinnabar, I updated us to the most recent release before checking if things were now okay.

I'm leaving us using a specific release for reproducability reasons, but honestly if I'd thought to check if there was a "release" branch (there is, it's up-to-date), I probably would have used that.  That might be the next thing to do if things ever break again.


asutherland:
Local provisioning has successfully gotten past the git-cinnabar point; I'm going to trigger automated re-provisioning now and expect it to pass, but I'll keep an eye on it.

PR #701, opened 2024-03-21
Update mozsearch.css to correct HCM/IC colors on Windows and Mac

anna-yeddi:
1. Moving most of the `prefers-contrast` queries to `forced-colors` to ensure the HCM only applies to WinOS. This would ensure the Increase Contrast mode on macOS is appearing as expected. 

2. Correcting the input field on `forced-colors` to use `Field`/`FieldText` color combination with `ButtonText` border, as recommended by @MReschenberg [in the comments to the original patch](https://github.com/mozsearch/mozsearch/pull/697#discussion_r1526755458)

3. I'm retaining the border for the blame dialog to be added on `prefers-contrast` modes across OSes to ensure the dialog is visually clearly discernible from the underlying page content.

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -151,7 +151,7 @@ table {
 }
 }
 
-@media (prefers-contrast) {
+@media (forced-colors) {
   :root {
     --blame-popup-background: Canvas;
     --blame-popup-color: CanvasText;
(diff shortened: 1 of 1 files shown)


emilio (approved):
Seems fine, but fwiw forced-colors can also happen in macOS and Linux...

anna-yeddi:
Thank you for the reminder for the onlookers!

Just in case, the full description of the `forced-colors` and `prefers-contrast` HCM queries could be found in the following docs: https://firefox-source-docs.mozilla.org/accessible/HCMMediaQueries.html

PR #711, opened 2024-05-02
Bug 1894632 - Calculate the non-UTF-8 file size correctly.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1894632


emilio (approved):

asutherland:
Thanks for the fix!

PR #715, opened 2024-05-06
Bug 1895316 - Linkify string literal with chrome:// and resource:// in JS files

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1895316
depends on https://github.com/mozsearch/mozsearch-mozilla/pull/232

This does the following:
  1. process the `chrome-map.json` into "URL to file path" map, and save it as `url-map.json`
  2. in `js-analyze.js`:
    1. add the link target at the top of the file, with the file path
    2. for each string literal with `chrome://` or `resource://`
      1. lazily load `url-map.json`, and lookup the file path for the URL
      2. add the link source for the string literal, with the target file path

with the today's m-c, the `chrome-map.json` is 1,685,516 bytes, and the processed `url-map.json` becomes 1,977,829 bytes.
So, for each JS file with `chrome://` or `resource://` string literal, the `js-analyze.js` loads and parses ~2MB JSON file.  If this is too problematic, I'll look into reducing the cost (e.g. use URL instead of file path, or split the JSON into smaller chunk, etc).

### DIFF
--- a/scripts/js-analyze.js
+++ b/scripts/js-analyze.js
@@ -1,5 +1,5 @@
 let nextSymId = 0;
-let localFile, fileIndex, mozSearchRoot;
+let localFile, sourcePath, urlMapFile, fileIndex, mozSearchRoot, urlMap;
 
 // The parsing mode we're currently using.
 let gParsedAs = "script";
@@ -460,6 +460,22 @@ function memberPropLoc(expr)
   // happen in locstr2)
   idLoc.start.column = idLoc.end.column - expr.property.name.length;
   return idLoc;
+}
+
+function ensureURLMap() {
+  if (urlMap) {
+    return;
+  }
+
+  try {
+    urlMap = JSON.parse(snarf(urlMapFile));
+  } catch (e) {
+    urlMap = {};
+  }
+}
+
+function atEscape(text) {
+  return text.replace(/[^A-Za-z0-9_/]/g, matched => "@" + matched.charCodeAt(0).toString(16).toUpperCase().padStart(2, "0"));
 }
 
 /**
@@ -1038,6 +1054,10 @@ let Analyzer = {
     }
 
     case "ImportDeclaration": {
+      if (stmt.moduleRequest && stmt.moduleRequest.source &&
+          stmt.moduleRequest.source.type === "Literal") {
+        this.maybeLinkifyLiteral(stmt.moduleRequest.source);
+      }
       break;
     }
 
@@ -1068,6 +1088,11 @@ let Analyzer = {
         } catch (ex) {
           logError(`Weirdness processing export, ignoring: ${ex}`);
         }
+      }
+
+      if (stmt.moduleRequest && stmt.moduleRequest.source &&
+          stmt.moduleRequest.source.type === "Literal") {
+        this.maybeLinkifyLiteral(stmt.moduleRequest.source);
       }
       break;
     }
@@ -1154,6 +1179,7 @@ let Analyzer = {
       break;
 
     case "Literal":
+      this.maybeLinkifyLiteral(expr);
       break;
 
     case "Super":
@@ -1446,7 +1472,41 @@ let Analyzer = {
       break;
     }
   },
+
+  maybeLinkifyLiteral(expr) {
+    if (typeof expr.value !== "string") {
+      return;
+    }
+
+    if (!expr.value.startsWith("chrome://") &&
+        !expr.value.startsWith("resource://")) {
+      return;
+    }
+
+    ensureURLMap();
+
+    if (!(expr.value in urlMap)) {
+      return;
+    }
+    const targetPath = urlMap[expr.value];
+
+    const name = "\"" + expr.value + "\"";
+    const loc = expr.loc;
+    const sym = "FILE_" + atEscape(targetPath);
+    this.source(loc, name, "file,use", "file " + targetPath, sym);
+    this.target(loc, name, "use", targetPath, sym);
+  },
 };
+
+function printFileTarget(path) {
+  print(JSON.stringify({
+    loc: "00001:0",
+    target: 1,
+    kind: "def",
+    pretty: "file " + path,
+    sym: "FILE_" + atEscape(path),
+  }));
+}
 
 // Helper for preprocessor directives so that JS assignments like `#error =`
 // won't match.  All of this is obviously optimized for clarity/not messing up
@@ -1967,7 +2027,11 @@ function analyzeFile(filename)
 fileIndex = scriptArgs[0];
 mozSearchRoot = scriptArgs[1];
 localFile = scriptArgs[2];
+sourcePath = scriptArgs[3];
+urlMapFile = scriptArgs[4];
 
 load(mozSearchRoot + "/sax/sax.js");
+
+printFileTarget(sourcePath);
 
 analyzeFile(localFile);
--- a/scripts/js-analyze.sh
+++ b/scripts/js-analyze.sh
@@ -15,5 +15,5 @@ TREE_NAME=$2
 
 cat $INDEX_ROOT/js-files | \
     parallel --halt 2 js -f $MOZSEARCH_PATH/scripts/js-analyze.js -- {#} \
-    $MOZSEARCH_PATH $FILES_ROOT/{} ">" $INDEX_ROOT/analysis/{}
+    $MOZSEARCH_PATH $FILES_ROOT/{} {} $INDEX_ROOT/url-map.json ">" $INDEX_ROOT/analysis/{}
 echo $?
--- a/scripts/mkindex.sh
+++ b/scripts/mkindex.sh
@@ -66,6 +66,10 @@ $MOZSEARCH_PATH/scripts/objdir-mkdirs.sh
 
 date
 
+$MOZSEARCH_PATH/scripts/process-chrome-map.py $GIT_ROOT $INDEX_ROOT/chrome-map.json $INDEX_ROOT/url-map.json || handle_tree_error "process-chrome-map.py"
+
+date
+
 $MOZSEARCH_PATH/scripts/js-analyze.sh $CONFIG_FILE $TREE_NAME || handle_tree_error "js-analyze.sh"
 
 date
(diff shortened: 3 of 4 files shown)


arai-a:
Example of the link with an import declaration.

resource-link

asutherland:
@arai-a This is very exciting!  This is probably a great practice opportunity for you to trigger an [AWS indexer run](https://github.com/mozsearch/mozsearch?tab=readme-ov-file#triggering-an-aws-indexer-run) against your branches targeting the "dev" channel (dev.searchfox.org) which is currently available for use.  You should get an email if anything happens.

asutherland:
> with the today's m-c, the `chrome-map.json` is 1,685,516 bytes, and the processed `url-map.json` becomes 1,977,829 bytes. So, for each JS file with `chrome://` or `resource://` string literal, the `js-analyze.js` loads and parses ~2MB JSON file. If this is too problematic, I'll look into reducing the cost (e.g. use URL instead of file path, or split the JSON into smaller chunk, etc).

Although it's always nice for the searchfox indexer run time to be fast, much of the searchfox rationale is that it's worth spending time during our batch offline indexing to make things faster for users when the indexing is done.  Also, since we run the JS indexing using parallel, I expect the cost to be amortized very quickly so it really shouldn't matter, and this is something we can check when your dev run finishes.

I created some scripts to help us analyze the runtime with the caveat that you need to install https://github.com/rcoh/angle-grinder (`cargo install ag`) in whatever context you want to use it.  Then you can do, assuming you've evaluated `MOZSEARCH=/path/to/mozsearch/checkout` and evaluated the mozsearch venv to get our pretty printing python dep installed:

```
 $  $MOZSEARCH/scripts/indexer-logs-fetch.sh 
download: s3://indexer-logs/index-2024-05-05T13:03+00:00_release5_config5.gz to ./index-2024-05-05T13:03+00:00_release5_config5.gz
download: s3://indexer-logs/index-2024-05-05T06:22+00:00_release1_config1.gz to ./index-2024-05-05T06:22+00:00_release1_config1.gz
download: s3://indexer-logs/index-2024-05-05T18:18+00:00_release1_config1.gz to ./index-2024-05-05T18:18+00:00_release1_config1.gz
download: s3://indexer-logs/index-2024-05-05T16:48+00:00_release2_config2.gz to ./index-2024-05-05T16:48+00:00_release2_config2.gz
download: s3://indexer-logs/index-2024-05-05T19:11+00:00_release4_config4.gz to ./index-2024-05-05T19:11+00:00_release4_config4.gz
```

This will delete one of the config1 logs it just downloaded.

Then you can run:
```
 $  $MOZSEARCH/scripts/indexer-logs-analyze.sh                                                            
Trees
...                                                                                
├── mozilla-central
│   └──                                                                
│         script                time since start   apparent duration  
│        ──────────────────────────────────────────────────────────── 
│         find-repo-files.py    0:00:00            0:02:37            
│         build.sh              0:02:37            0:14:57            
│         scip-analyze.sh       0:17:34            0:01:34            
│         js-analyze.sh         0:19:08            0:04:28            
│         idl-analyze.sh        0:23:36            0:00:15            
│         ipdl-analyze.sh       0:23:51            0:00:02            
│         crossref.sh           0:23:53            0:20:08            
│         output.sh             0:44:01            0:08:09            
│         build-codesearch.py   0:52:10            0:03:44            
│         compress-outputs.sh   0:55:54            0:04:27            
│         check-index.sh        1:00:21                                
│     
...
```

Here we can see that it currently takes the indexer about 4 minutes 28 seconds to perform the JS analysis path.  If there's somehow a major change to that number, we should probably check what's going on because that would be surprising, but it's certainly worth spending more time to improve our JS experience!

That said, because it seems very likely we'll want to move much of our JS experience to using scip-typescript in the future and our `scip-indexer.rs`, I probably would not spend a ton of time optimizing anything going into js-analyze.js (although I could believe we might want to continue having it around as a fallback or for hybridized analysis between the two).

arai-a:
Thank you for your suggestion and the detailed explanation!

I've triggered the AWS indexer, and now it's available in dev channel
https://dev.searchfox.org/mozilla-central/source/browser/components/BrowserContentHandler.sys.mjs

js-analyze takes `0:05:05`, which I guess acceptable amount of increase?

arai-a:
Just noticed that this works only for the case where the file pointed by the URL is also a JS file.
CSS and other resources aren't handled, because we don't perform analysis file for them, and they don't have target definition.
(`url-map.json` contains non-JS as well, and `source` for them is generated for strings in JS files. we just need `target`)

Possible options would be:
  * (a) generate analysis files for CSS and other resources with `target` data, and feed them to crossref
  * (b) generate the `target` data in crossref itself

(a) is simpler solution. CSS, HTML, and SVG can also have references inside them, and linkifying those references would also be nice.
(b) would work nicely for the case where the file has no reference inside it, e.g. PNG.

if (a) is okay, I'd like to go with the current patch's way for JS case, and address non-JS cases as a followup.

asutherland:
> if (a) is okay, I'd like to go with the current patch's way for JS case, and address non-JS cases as a followup.

In general it's always okay and even desirable to land enhancements incrementally and iteratively, especially given how hard it can be to find time to contribute to searchfox!  Far better to get a small improvement in now than to have a large improvement stuck on a branch!  (And to be clear, I think this will be a major improvement!)

I think the main concern in reviews as it relates to landing would be that we want to avoid regressions in data/UX or on the user's muscle memory, and especially to avoid churn over a short time.  The changes here are additive and the potential follow-ups will be additive, so there's definitely no problem with landing this now.

> Just noticed that this works only for the case where the file pointed by the URL is also a JS file. CSS and other resources aren't handled, because we don't perform analysis file for them, and they don't have target definition. (`url-map.json` contains non-JS as well, and `source` for them is generated for strings in JS files. we just need `target`)
> 
> Possible options would be:
> 
>     * (a) generate analysis files for CSS and other resources with `target` data, and feed them to crossref
> 
>     * (b) generate the `target` data in crossref itself
> 
> 
> (a) is simpler solution. CSS, HTML, and SVG can also have references inside them, and linkifying those references would also be nice. (b) would work nicely for the case where the file has no reference inside it, e.g. PNG.

I think it likely makes sense to do both; if we don't have analysis for a file in crossref, it would make sense to still emit a symbol for the file.

Going to do a quick review now; thank you for kicking off the dev channel try indexing!

scripts/js-analyze.js
```diff
@@ -1446,8 +1470,31 @@ let Analyzer = {
...
+    }
+
+    ensureURLMap();
+
+    if (!(expr.value in urlMap)) {
+      return;
+    }
+    targetPath = urlMap[expr.value];
+
+    this.source(expr.loc, "\"" + expr.value + "\"", "file,use", "file " + targetPath, "FILE_" + atEscape(targetPath));
```
asutherland, line 1490: It would be great to emit a target record here, too!  I'm not sure if you noticed, but there are `(file symbol)` spans to the right of the bread crumbs at the tops of files, like at https://dev.searchfox.org/mozilla-central/source/js/xpconnect/loader/XPCOMUtils.sys.mjs and if you click on them they offer, using an example of https://searchfox.org/mozilla-central/source/dom/workers/JSSettings.h [search which shows uses](https://searchfox.org/mozilla-central/search?q=symbol:FILE_dom%2Fworkers%2FJSSettings%402Eh&redirect=false), [calls-to diagram which shows the include dependency graph that ends up including the file](https://searchfox.org/mozilla-central/query/default?q=calls-to%3A%27dom%2Fworkers%2FJSSettings.h%27%20depth%3A4), and [calls-from which shows all the files that file includes and transitively too](https://searchfox.org/mozilla-central/query/default?q=calls-from%3A%27dom%2Fworkers%2FJSSettings.h%27%20depth%3A4).

I think this should automatically work for diagrams if you add the target records, but it is possible that I may have put some guards in place somewhere that inhibit things... if that happens it should be easy enough to fix and the immediate ability to search for the inclusions of the file I think would be useful.

asutherland (approved):
This is awesome and will be a huge quality of life improvement for people in general, but me in particular, so I very much appreciate this! ;)

I think you should have landing access since you're already in the https://github.com/orgs/mozsearch/teams/core-developers group so feel free to land whenever you want.  As I mention, it would be nice to have the target records but you should feel free to land this as-is if you prefer.  (And then feel free to land the target records without additional review if you prefer.  I'm always happy to do another round of review/looking, if you like, but I don't think there's any need.)

arai-a:
Thank you for reviewing!
Yes, after adding `target`, I can see all references
https://dev.searchfox.org/mozilla-central/search?q=symbol:FILE_toolkit%2Fmodules%2FAppConstants%402Esys%402Emjs&redirect=false

The diagram doesn't show usage, but I'll leave it to followup patch.

asutherland:
> The diagram doesn't show usage, but I'll leave it to followup patch.

It is working, a limit is just hit because the file gets used so much.  In this case there are 344 different files that reference that file, and the graph heuristics know that is unlikely to result in a useful graph.  But if we [raise the path-limit to 350 we can see (a not super useful graph)](https://dev.searchfox.org/mozilla-central/query/default?q=calls-to%3A%27toolkit%2Fmodules%2FAppConstants.sys.mjs%27+depth%3A4+path-limit%3A350).

But this can still be useful if we turn the calls-to usage into a [calls-between usage and specify another path](https://dev.searchfox.org/mozilla-central/query/default?q=calls-between%3A%27toolkit%2Fmodules%2FAppConstants.sys.mjs%27+calls-between%3A%27toolkit%2Fcomponents%2Fextensions%2FExtensionScriptingStore.sys.mjs%27++depth%3A4+path-limit%3A350).

asutherland:
This seems to be working great on searchfox.org now!  Woo!

It definitely would be great to send a ["Searchfox Update:" to dev-platform](https://groups.google.com/a/mozilla.org/g/dev-platform/search?q=%22Searchfox%20Update%22) about this!  I imagine I am not the only person who will greatly appreciate this!  (Also the improvements to symbol link are probably worth mentioning too, since people may not know to check that or may be used to it not working, so knowing it works much better now will also be useful to them.)

PR #716, opened 2024-05-07
Use -v option for the date command on macOS in indexer-logs-fetch.sh

arai-a:
macOS's date command doesn't support `--date` option, but it has `-v` option to adjust the date


asutherland:
Also thanks!

PR #717, opened 2024-05-07
Use GNU grep on macOS in indexer-logs-analyze.sh

arai-a:
macOS's grep doesn't have -P option.
GNU grep is available via homebrew, as ggrep.


asutherland:
Thanks!

PR #718, opened 2024-05-07
Test stabilization, go to def improvements

asutherland:
See specific commit messages for more detail.


asutherland:
Given that we potentially have a bunch of things landing, I'm going to kick off a config1 indexer run right now and validate the results when it completes.  The additional go to def stuff is straightforward so I don't have major concerns, but I do want to make sure I validate that without multiple feature changes potentially being conflated in a single run.

asutherland:
The run ran fine, results were consistent with my local `make build-mozilla-repo` tests.

PR #721, opened 2024-05-11
Bug 1896165 - Use chrome-map.json files from searchfox jobs

arai-a:
depends on https://bugzilla.mozilla.org/show_bug.cgi?id=1895989 and https://github.com/mozsearch/mozsearch-mozilla/pull/234

bug 1895989 will add `chrome-map.json` to the artifact of searchfox(idx) job, for each OS.
This combines all `chrome-map.json` info into single `url-map.json`, with using array of strings instead of string for each mapping, so that single URL can be mapped to multiple files if it's conditional on OS.
e.g. `resource://passwordmgr/passwordstorage.sys.mjs` maps to `toolkit/components/passwordmgr/storage-desktop.sys.mjs` on desktop and `toolkit/components/passwordmgr/storage-geckoview.sys.mjs` on android.

The `js-analyze.js` is modified to use the array and generate source/target for each.

resource-link-per-os

### DIFF
--- a/scripts/js-analyze.js
+++ b/scripts/js-analyze.js
@@ -1488,13 +1488,15 @@ let Analyzer = {
     if (!(expr.value in urlMap)) {
       return;
     }
-    const targetPath = urlMap[expr.value];
+    const targetPaths = urlMap[expr.value];
 
     const name = "\"" + expr.value + "\"";
     const loc = expr.loc;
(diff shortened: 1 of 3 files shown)


asutherland:
I know this is a WIP but r=asuth for what's there right now (and I don't think I'd need to review any changes)

PR #722, opened 2024-05-12
Bug 1895876 - Index script and attributes in HTML files

arai-a:
This adds the following:
  * Handle script inside `<script>` and `on*` attributes in html and xhtml files
  * Linkify `chrome://` and `resource://` URLs  in HTML attributes

Now both JS files and HTML files are added to references, and they link each other.

This needs to be rebased onto https://github.com/mozsearch/mozsearch/pull/721


arai-a:
ID in script element
html-ID-in-script

Linkify URL in script element
html-URL-in-script

Linkify URL in attribute
html-URL-in-attribute

Link to HTML files from JS files
html-URL-in-jsfile

There's a problem with identifier in attributes vs identifier in script element, where they don't seem to point the same thing.
I'll look into this.
html-ID-in-attribute

arai-a:
> There's a problem with identifier in attributes vs identifier in script element, where they don't seem to point the same thing.
Oh, this is because the symbol in the `script` element gets `fileIndex + "-" + symId` style, which cannot be referred from separate `Analysis` instance.

I'll check if reusing symbols table across `Analysis` works

arai-a:
Putting the script element in the top-level scope and deferring the `on*` attribute analysis solved the issue.

html-script-and-attribute

asutherland:
Very exciting!  I looked at the individual commits (thanks for the clean factoring and excellent commit messages) and r=asuth for parts 1 through 14.

One thing that is probably worth thinking about is adding some [testing checks](https://github.com/mozsearch/mozsearch/blob/master/docs/testing-checks.md) to the tests repo.  As noted in the docs, these need not be exhaustive; it doesn't take much in the way of checks in order to help detect if something breaks or to help illustrate what data a part of the system produces.  It's also okay if you don't add any right now; it's only essential to add them if something regresses and we want to make sure it doesn't regress again.

For example, the [analysis/js/some_javascript.js/all__priv_field_num__json](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/inputs/analysis/js/some_javascript.js/all__priv_field_num__json) check produces [this snapshot output](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/snapshots/analysis/js/some_javascript.js/check_glob%40all__priv_field_num__json.snap) and then the [html-producing variation of the check](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/inputs/analysis/js/some_javascript.js/all__priv_field_num__html) produces [this snapshot](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/snapshots/analysis/js/some_javascript.js/check_glob%40all__priv_field_num__html.snap).

If you want coverage to include the url map, [tests/tests/metadata](https://github.com/mozsearch/mozsearch/tree/master/tests/tests/metadata) is where we put synthetic files like that and the [tests/tests/setup](https://github.com/mozsearch/mozsearch/blob/master/tests/tests/setup) script symlinks them into place.  As you may note from the fake files there, so far I've only put a small amount of fake data in there to get us enough coverage for specific test cases or manual testing and to avoid creating files that bit rot.  But sometimes the easiest thing is to just check in stuff from mozilla-central where it would be a hassle to craft fake data or there are complicated toolchain steps that we don't actually need or want to run in our tests tree.  For example, https://github.com/mozsearch/mozsearch/pull/528 checked m-c artifacts into [mc-analysis](https://github.com/mozsearch/mozsearch/tree/master/tests/tests/mc-analysis) and [mc-generated](https://github.com/mozsearch/mozsearch/tree/master/tests/tests/mc-generated).

One thing I would suggest is that if you're adding new files for test cases, it probably makes sense to put them in a subdirectory of [tests/tests/files](https://github.com/mozsearch/mozsearch/tree/master/tests/tests/files) because, as you know from having to update the snapshot before, the directory listing check of the root directory otherwise will otherwise tend to churn more than it probably strictly needs.  Also, I should call out that a current implementation limitation is that even though the check filenames live in a hierarchical path structure, the actual implementation does key them just by their name without their file path, so it is desirable to give the file names unique names.  Otherwise the test runner tries to deduplicate the names and that can complicate things if names stop being duplicated.

arai-a:
One thing I worry about is Part 21, where an intentional syntax error appears in [test_bug531176.html](https://searchfox.org/mozilla-central/rev/dd6e430c1bc2db90d9b3b1dd7e5215b4edc4d51a/dom/tests/mochitest/bugs/test_bug531176.html#59), which doesn't match the existing pattern for ignoring intentional error.
If there's better way to handle such case, let me know.

asutherland:
r=asuth through part 21.  Thanks for adding all the test coverage and that's a nice enhancement with `--symbol-prefix`!

> One thing I worry about is Part 21, where an intentional syntax error appears in [test_bug531176.html](https://searchfox.org/mozilla-central/rev/dd6e430c1bc2db90d9b3b1dd7e5215b4edc4d51a/dom/tests/mochitest/bugs/test_bug531176.html#59), which doesn't match the existing pattern for ignoring intentional error. If there's better way to handle such case, let me know.

The existing pattern really doesn't matter.  The high level goal has been to make sure that we the searchfox maintainers would notice when some kind of new thing was added to SpiderMonkey and we needed to update js-analyze.js to support it so searchfox would support everything that's getting used in tree.  As long as we keep the warning emails non-spammy, the warning mechanism works well for this.

So the general idea was to make sure that we downgrade all the problems that we run into and have decided don't need js-analyze enhancements get downgraded to info.  All the structure around that and the separate sections has been to try and be good about indicating why we added them, especially in case anyone ever wanted to improve things.  And then it was always easier to just add things to the lists as they came up than maybe try and do a big clean up.

Especially since it's possible much of the existing js-analyze.js stuff could be mooted through use of scip-typescript, I very much am not concerned if this mechanism is not beautiful.  Especially since you are the person actively enhancing js-analyze, I think it's really up to you how you want to handle things and to decide whether something is too hacky or anything like that :)

PR #723, opened 2024-05-15
Load sax.js only if necessary

arai-a:
sax is not used for JS files.

### DIFF
--- a/scripts/js-analyze.js
+++ b/scripts/js-analyze.js
@@ -2046,10 +2046,21 @@ class XBLParser extends XMLParser {
   }
 }
 
+// Prepare the `sax` global variable.
+// This function shouldn't be called multiple times.
+//
+// js-analyze.js is executed for single file, and the code path
+// "analyzeFile -> analyze* -> loadSax" is taken at most once.
+function loadSax()
+{
+  load(mozSearchRoot + "/sax/sax.js");
+}
+
 function analyzeXBL(filename)
 {
   let text = preprocess(filename, line => `<!--${line}-->`);
 
+  loadSax();
   let parser = sax.parser(false, {trim: false, normalize: false, xmlns: true, position: true});
 
   new XBLParser(filename, parser);
@@ -2102,6 +2113,7 @@ function analyzeXUL(filename)
     text = "<root>" + text + "</root>";
   }
 
+  loadSax();
   let parser = sax.parser(false, {trim: false, normalize: false, xmlns: true, position: true, noscript: true});
 
   let parser2 = new XULParser(filename, parser);
@@ -2120,6 +2132,7 @@ function analyzeHTML(filename)
     text = "<root>" + text + "</root>";
   }
 
+  loadSax();
   let parser = sax.parser(false, {trim: false, normalize: false, xmlns: true, position: true, noscript: false});
 
   let parser2 = new HTMLParser(filename, parser);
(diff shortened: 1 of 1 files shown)


scripts/js-analyze.js
```diff
@@ -2046,10 +2046,16 @@ class XBLParser extends XMLParser {
   }
 }
 
+function loadSax()
+{
+  load(mozSearchRoot + "/sax/sax.js");
```
asutherland, line 2051: I'm guessing the [isRunOnce](https://searchfox.org/mozilla-central/rev/7a8904165618818f73ab7fc692ace4a57ecd38c9/js/public/CompileOptions.h#481) mechanism and its [hookup to the JS cache](https://searchfox.org/mozilla-central/rev/7a8904165618818f73ab7fc692ace4a57ecd38c9/js/src/shell/js.cpp#2288-2290) and since (most usefully since there don't really seem to be clear comments about the field) [setModule sets it in setModule](https://searchfox.org/mozilla-central/rev/7a8904165618818f73ab7fc692ace4a57ecd38c9/js/public/CompileOptions.h#561), we expect this to evaluate the file at most once.  It would be good to add a comment here to this effect like "jsshell load() will run a given path at most once because of setIsRunOnce".  Because, from a DOM perspective, for something like `importScripts` we do expect the script to run multiple times (I think?).

It might also make sense to change the loadSax name to `ensureSaxLoaded` for additional clarity, but I think the comment is necessary either way and that would cover things.

asutherland (approved):

scripts/js-analyze.js
arai-a: There's no guard against multiple runs.

"RunOnce" is only for enabling an optimization for object literal etc ([[1]](https://searchfox.org/mozilla-central/rev/55b8c839700564409af5295286a1fd389410e19b/js/src/frontend/BytecodeEmitter.cpp#9731-9733), [[2]](https://searchfox.org/mozilla-central/rev/55b8c839700564409af5295286a1fd389410e19b/js/src/vm/Opcodes.h#875-896)), and it means "the bytecode doesn't support multiple execution of the top-level script object".  `load` compiles the file each time, which results in creating different top-level script object each time, and the "RunOnce" requirement doesn't apply across each `load` call.

Module evaluation is guarded by the module loader's module map ([[3]](https://searchfox.org/mozilla-central/rev/55b8c839700564409af5295286a1fd389410e19b/js/loader/ModuleLoaderBase.cpp#446-447,468-470)). "RunOnce" here is used for modules because it's guaranteed by module loader and we can apply the optimization.

Then, I don't put the guard here because `js-analyze.js` is called for single file, and `sax` is used at most once, and `loadSax` is also called at most once.
I'll add comment.

asutherland: > Then, I don't put the guard here because js-analyze.js is called for single file, and sax is used at most once, and loadSax is also called at most once.

This was my confusion.  When we were talking about the runtime of js-analyze.js previously I looked at the use of `parallel` in `js-analyze.sh` way too quickly and misread it, thinking it was [passing multiple args](https://www.gnu.org/software/parallel/parallel_tutorial.html#more-than-one-argument) so js-analyze.js was doing batch processing.  But, as you say, it is not.  (And the explicit redirection in the invocation for the output file should have made that more clear to me; I've been spending too much time with our batch-processing jobs!)

I'll update the [section on GNU parallel in the bash cheatsheet doc](https://github.com/mozsearch/mozsearch/blob/e83177fb40e6711484a254226929e595d5711f8f/docs/bash-scripting-cheatsheet.md#gnu-parallel-and-its-processing) to help me refresh my understanding of its operation in the future more quickly/accurately.

PR #726, opened 2024-05-17
Bug 1795177 - Add macosx64-aarch64 to document.

arai-a:
Depends on https://phabricator.services.mozilla.com/D210841 and https://github.com/mozsearch/mozsearch-mozilla/pull/236


asutherland (approved):
Thanks for taking the time to update this doc file too!

PR #728, opened 2024-05-19
Fix chrome-map override direction

arai-a:
The chrome-map's override should've been handled in the opposite way.


arai-a:
Actually, I'll put this into bug 1896441 patch stack.

PR #729, opened 2024-05-19
Bug 1896441 - Handle chrome/resource URL mapping as separate step.

arai-a:
Instead of handling the mapping in `js-analysis.sh` step, generate entries with `URL_` symbol,
and replace the item with `FILE_` symbol items as separate `replace-aliases.sh` step.
                                                                                                                
This allows using `URL_` symbol in C++ file.

The `replace-aliases.sh` step takes 20 seconds locally.


scripts/mkindex.sh
```diff
@@ -66,7 +66,7 @@ $MOZSEARCH_PATH/scripts/objdir-mkdirs.sh
 
 date
 
-$MOZSEARCH_PATH/scripts/process-chrome-map.py $GIT_ROOT $INDEX_ROOT/url-map.json $INDEX_ROOT/*.chrome-map.json || handle_tree_error "process-chrome-map.py"
+$MOZSEARCH_PATH/scripts/process-chrome-map.py $GIT_ROOT $INDEX_ROOT/aliases/url-map $INDEX_ROOT/*.chrome-map.json || handle_tree_error "process-chrome-map.py"
```
asutherland, line 69: Is there a reason to not have the file have a `.json` extension anymore?  It seems like the file is still legal JSON?

asutherland (approved):
Looks great!  I have some questions / comment addition suggestions, but there's no need to change anything in the patch.  In particular, the use of "parallel" could happen in a follow-up.

PR #730, opened 2024-05-20
Build fix: `cargo update` run to get time 0.3.36 update to fix nightly build failure.

asutherland:
As noted on the thread for the indexer failure, we failed to update because the new rustc nightly does not time 0.3.34.  So I did a `cargo update` and then did a fixup for tree-sitter-rust which had performed a divergent upgrade to 0.22.x which I manually downgraded to 0.20.9 to match everything else.

### DIFF
--- a/tools/Cargo.lock
+++ b/tools/Cargo.lock
@@ -19,12 +19,12 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "ahash"
-version = "0.8.7"
+version = "0.8.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


asutherland:
I've initiated re-provisioning as documented by the steps at https://github.com/mozsearch/mozsearch/blob/master/docs/aws.md#provisioning-and-cloud-init and we should get emails in 20-30 minutes confirming that it worked or not.  I'll re-trigger the indexer job assuming those don't fail.

asutherland:
Provisioning completed, I triggered the release1 indexer again.  I cleaned up the AMIs leaving the new and most recent AMI around from https://docs.google.com/document/d/1fHMTgxbXXu_hH4ileEWkJd2QHil_xZfn2lMAW9UvP3s/edit?usp=sharing and triggering deletion of all snapshots (which leaves those associated with those 4 AMIs still around inherently).

PR #732, opened 2024-05-22
Handle imported and exported symbols and related expressions

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1897787

This does the following:
  * Define imported symbols: `import { exportedName } from "...";`
  * Handle reference (exported name) inside import declarations:  `import { exportedName as localName } from "...";`
  * Define declarations inside exported declaration in more case: `export default function localName() {}`
  * Handle reference (local name) inside export declarations: `export { localName as exportedName };`

### DIFF
--- a/tests/tests/checks/snapshots/analysis/js/export.mjs/check_glob@exported_decls__json.snap
+++ b/tests/tests/checks/snapshots/analysis/js/export.mjs/check_glob@exported_decls__json.snap
@@ -1,0 +1,255 @@
+---
+source: tests/test_check_insta.rs
+expression: "&json_results"
+---
+[
+  {
+    "loc": "00001:0",
+    "target": 1,
+    "kind": "def",
+    "pretty": "file js/export.mjs",
+    "sym": "FILE_js/export@2Emjs"
+  },
+  {
+    "loc": "1:11-18",
+    "source": 1,
+    "syntax": "def,prop",
+    "pretty": "property letDecl",
+    "sym": "#letDecl"
+  },
+  {
+    "loc": "1:11-18",
+    "target": 1,
+    "kind": "def",
+    "pretty": "letDecl",
+    "sym": "#letDecl",
+    "context": ""
+  },
+  {
+    "loc": "2:13-22",
+    "source": 1,
+    "syntax": "def,prop",
+    "pretty": "property constDecl",
+    "sym": "#constDecl"
+  },
+  {
+    "loc": "2:13-22",
+    "target": 1,
+    "kind": "def",
+    "pretty": "constDecl",
+    "sym": "#constDecl",
+    "context": ""
+  },
+  {
+    "loc": "3:16-23",
+    "source": 1,
+    "syntax": "def,prop",
+    "pretty": "property funDecl",
+    "sym": "#funDecl",
+    "nestingRange": "3:26-3:27"
+  },
+  {
+    "loc": "3:16-23",
+    "target": 1,
+    "kind": "def",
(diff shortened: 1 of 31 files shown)


tests/tests/checks/snapshots/analysis/js/export.mjs/check_glob@exported_decls__json.snap
```diff
@@ -0,0 +1,255 @@
...
+    "sym": "#local2"
+  },
+  {
+    "loc": "6:17-23",
+    "target": 1,
+    "kind": "def",
+    "pretty": "local2",
+    "sym": "#local2",
+    "context": ""
+  },
```
asutherland, line 149: I wonder if we should either use the "alias" kind (introduced for [typedefs/using aliases](https://searchfox.org/mozilla-central/rev/6d437ba43e7752a85f5731133ee66152b390aa16/build/clang-plugin/mozsearch-plugin/MozsearchIndexer.cpp#1817)) or, potentially better, explicitly define import/export kinds.  Specifically, it's our policy to only generate "go to def" context menu links when there is a single def, so having exports treated as defs is suboptimal for that purpose, although we tend to have so many collisions in JS because of the single global namespace soup we put everything in, I don't know that it matters that much.

No need to change anything in this patch, but I think it's a potential interesting refinement.  (And arguably it's better to do in a different patch/stack since it involves a number of changes to things like router.py.  I imagine the patch that added "alias" probably would provide useful context.)

tests/tests/checks/snapshots/analysis/js/import.mjs/check_glob@imported_symbols__json.snap
```diff
@@ -0,0 +1,118 @@
...
+    "sym": "#ns"
+  },
+  {
+    "loc": "1:7-9",
+    "target": 1,
+    "kind": "def",
+    "pretty": "ns",
+    "sym": "#ns",
+    "context": ""
+  },
```
asutherland, line 27: So we could use a "decl" kind here, but again, probably having import/export might work better.  I guess a fun thing that could be done here is also that if we baked the file path into the symbols we could also have the jumpref mechanism (which I think I discussed somewhat recently with Botond in scrollback in the channel) then provide for "go to export".  That is, right now `#ns` is of course not unique, but `FILEPATH#ns` is and there should be a unique pair between the import and the export making it easy for jumpref and the context-menu to do the right thing.

It's also fine if we don't do that, as this potentially gets into overlap with scip-typescript, but we would likely end up modeling it in a similar fashion in that case.  (The main difference would just be the symbols that we assign in that case.)

I realize you may already have plans for something like this, in which case, consider this a confirmation of the wiseness of your plans :)

asutherland (approved):
Woooo!

tests/tests/checks/snapshots/analysis/js/import.mjs/check_glob@imported_symbols__json.snap
arai-a: Yeah, adding "go to export" would be really helpful.
I'll look into it after [bug 1895653](https://bugzilla.mozilla.org/show_bug.cgi?id=1895653) to also handle relative paths cases.

tests/tests/checks/snapshots/analysis/js/export.mjs/check_glob@exported_decls__json.snap
arai-a: I'll experiment with "alias" in followup

PR #733, opened 2024-05-23
Index custom properties and URLs in CSS

arai-a:
This adds the following:
  * Index CSS files
  * Index `style` elements in HTML files
  * Index `style` attributes in HTML files
  * Handle custom properties (`--some-thing`), both definitions and references, with `CSSPROP_*` prefix
  * Add `FILE_*` target for CSS files
  * Index `chrome://` and `resource://` URLs in style rules

Parsing is done with [tree-sitter-css](https://github.com/tree-sitter/tree-sitter-css/), via WASM binding.
WASM binding is chosen in order to integrate CSS parsing with HTML parsing.

Import for tree-sitter files are done manually, with steps described in `tree-sitter-js/README` file.

`css-analyze.sh` takes 28 seconds on mozilla-central locally


arai-a:
possible followup would be to handle ID and class names in the selector, both in HTML and CSS, and link them

tools/src/tokenize.rs
```diff
@@ -1256,6 +1271,18 @@ pub fn tokenize_tag_like(string: &str, script_spec: &LanguageSpec) -> Vec<Token>
                         kind: t.kind,
                     });
                     result.extend(script_toks);
+                } else if literal_is_css {
+                    literal_is_css = false;
+
+                    let css_start = token.start;
+                    let css = &string[token.start..token.end];
+                    let css_toks = tokenize_css(&css.to_owned());
```
emilio, line 1279: do you need this to_owned()?

tools/src/tokenize.rs
```diff
@@ -1267,7 +1294,7 @@ pub fn tokenize_tag_like(string: &str, script_spec: &LanguageSpec) -> Vec<Token>
                     tag_stack.pop();
                 }
 
-                let starting = peek(&tag_stack, 0, "script")
+                let starting_script = peek(&tag_stack, 0, "script")
                     || peek(&tag_stack, 0, "constructor")
```
emilio, line 1298: We should remove these XBL remnants btw.

scripts/js-analyze.js
```diff
@@ -2103,6 +2150,146 @@ function analyzeHTML(filename)
   parser2.processEventListeners();
 }
 
+class CSSAnalyzer {
```
emilio, line 2153: We already have the rust CSS parser which we use to tokenize the CSS. Maybe we should use it to analyze the CSS as well? Should be trivial-ish.

But not objecting to using js-analyze I guess, it's just probably less efficient.

emilio (approved):
r=me on the tokenizer changes, but I'd like andrew to review the js-analyze stuff. On one hand it seems unfortunate having to use yet another css parser. On the other... well, maybe it's not a big deal.

arai-a:
Yeah, we could instead do the following if necessary:
  * build indexer command based on the in-tree CSS parser
  * build WASM binding of the indexer command as well
  * for standalone CSS, use the native binary indexer command
  * for CSS inside HTML, use WASM binding from `js-analyze.js`

arai-a:
experimenting with `cssparser` crate, both with native binary and WASM binding:
https://github.com/arai-a/mozsearch/commits/index-css-shared-lib/
(not yet integrated to js-analysis nor mkindex tho)

tree-sitter-js/README
```diff
@@ -0,0 +1,13 @@
...
+tree-sitter.js and tree-sitter.wasm are imported from
+https://github.com/tree-sitter/tree-sitter/releases/tag/v0.22.6
+
+# tree-sitter-css 0.21.0
+
+tree-sitter-css.wasm is generated by the following commands in empty directory:
+```
+npm install --save-dev tree-sitter-cli tree-sitter-css
+npx tree-sitter build --wasm node_modules/tree-sitter-css
+```
```
asutherland, line 12: Given that we have a provisioning step ([where we already install node](https://github.com/mozsearch/mozsearch/blob/0f89a38f08c8cdc4375291feaf6ee11c5736961a/infrastructure/indexer-provision.sh#L23-L46)) and we have a build step, I don't understand why we're introducing manual steps here that result in us having to check in non-trivial binary blobs that would potentially beget additional manual work in the future to update.

This would potentially go with using cssparser as well.

If this is motivated by how long it can take to provision or the build times of the `make build-test-repo`, I think it probably makes more sense to try and improve those to do less re-work or otherwise be more ergonomic[1].  Right now I think the current idiom for experimenting with adding things to the provisioning process and being able to do try runs is to update https://github.com/mozsearch/mozsearch/blob/master/infrastructure/indexer-update.sh so that the indexer will run the steps at its startup.  Also one can obviously run the script from within the provisioned docker image.  (But if you reprovision, the script will even get run twice, so it will have effects and we will verify that it won't break when an indexer AMI updates.)

Aside: I do understand it's not unusual for JS projects to potentially vendor WASM, but I think that tends to be justified by environments where there is already aggressive vendoring and/or the inability/policy to not run additional build steps beyond webpack/esbuild or whatever.  (I know we have our weird/ugly sax.js thing, but that's definitely not best practice for anything! ;)

1: Like Nicolas had added a [nix flake](https://github.com/mozsearch/mozsearch/blob/master/flake.nix) that I don't know if that would be a better workflow for us to use in some cases?  It seems like the flake may only be for outside of docker, though, which makes sense since we don't install nix into the docker image.

scripts/js-analyze.js
asutherland: I think the idea of doing this from within js-analyze is that we also [use it for the HTML analysis too](https://github.com/mozsearch/mozsearch/blob/0f89a38f08c8cdc4375291feaf6ee11c5736961a/scripts/html-analyze.sh#L17) and there is synergy for HTML/JS/CSS analysis to all be using the same codebase because of how HTML can include CSS and JS inline and how an unfortunate reality is that JS may also potentially contain inline HTML and CSS as well.

In terms of using tree-sitter-css, that was my suggestion in a DM conversation where that topic didn't really need to be in DM, so I'll include my message below.  The way we end up using the parser here doesn't really result in any direct synergies that I was thinking of, though, and potentially using tree-sitter for CSS tokenization (for consistency with blame history and the potential for cheap position:sticky) is still a ways off, so I think the change to cssparser makes sense (and provides more synergy for the current tokenizer situation).

The info dump messages:

> for tree-sitter-css, the main reason I find it interesting is, and I think I mentioned something related this recently... the [tree-sitter query syntax (docs)](https://tree-sitter.github.io/tree-sitter/using-parsers#query-syntax) is quite powerful and enables the ability to treat different languages somewhat generically.  So for example for the hyperblame effort, the [C++ query support](https://github.com/mozsearch/mozsearch/blob/master/tools/languages/tokenizer_queries/cpp.scm) helps us map from the AST to providing nesting semantics for C++, but then we also are able to handle [python](https://github.com/mozsearch/mozsearch/blob/master/tools/languages/tokenizer_queries/python.scm), [rust](https://github.com/mozsearch/mozsearch/blob/master/tools/languages/tokenizer_queries/rust.scm), and [TS/JS](https://github.com/mozsearch/mozsearch/blob/master/tools/languages/tokenizer_queries/typescript.scm) without much per-language glue code.
>
> There is an interesting potential to use (different) queries and some additional configuration to let a more generic tool built on tree-sitter potentially provide the best-effort indexing in various cases.  Like how [the tree-sitter-cpp has a tags.scm](https://github.com/tree-sitter/tree-sitter-cpp/blob/master/queries/tags.scm) for tags indexing, and that same mechanism can then [be used for ocaml](https://github.com/tree-sitter/tree-sitter-ocaml/blob/master/queries/tags.scm) as one example of many. https://tree-sitter.github.io/tree-sitter/code-navigation-systems documents the tags used for the tags command.
>
> There is no tags.scm for tree-sitter-css right now, though, just [highlights](https://github.com/tree-sitter/tree-sitter-css/blob/master/queries/highlights.scm) for the [tree-sitter syntax highlighting support](https://tree-sitter.github.io/tree-sitter/syntax-highlighting).  I don't know how well / if the tags semantics would translate well for searchfox or not.  (In general it's nice when we can have precise indexing as opposed to the syntactically based best-effort like tree-sitter would provide.)

> But CSS is also such a fundamental part of the web with semantics very unlike normal program languages, so it could also make sense to not try and use the more generic tree-sitter-css at all for semantic indexing.  Especially since it does seem likely a lot of (existing) CSS tooling will need token position information anyways for responsible error reporting.
>
> (And we can still use tree-sitter-css for the hyperblame implementation.  Like we're using tree-sitter-mozcpp for hyperblame even though we don't use that for our actual indexer/analysis.)

tree-sitter-js/README
asutherland: Relatedly, I think it would be nice to avoid adding new top-level directories like this where it seems like without some kind of hierarchy.  That is, if you ended up adding another WASM dep for js-analyze, it seems like it would also end up as another top-level directory too.  Maybe we could add a subdir like `scripts/js-analyze-deps` or something to be the parent of js-analyze specific things which would also help make the relationship more clear?

Alternately, maybe given the significance and notability of js-analyze, maybe we should create a new root dir like `web-analyze` and `js-analyze.js` the script could move in there and maybe be renamed to `web-analyze.js` (maybe under a `src/` subdir?) since it covers all of js/css/html.  And then the dep subdirs could live there too.

asutherland (changes requested):

asutherland:
> experimenting with `cssparser` crate, both with native binary and WASM binding: https://github.com/arai-a/mozsearch/commits/index-css-shared-lib/

Interesting!  So the idea there is that we do create a entirely rust-based CSS analyzer but that we let `js-analyze.js` do its integration by loading that logic via WASM to be able to call into it!

One interesting question (which maybe @emilio can help answer) would be whether there's a way to have the rust logic live under tools/ in a way that the WASMed logic doesn't end up bloating the WASM binary too much.

In particular, one thing that would be very nice to have is for the CSS analyzer to use the rust structs from https://github.com/mozsearch/mozsearch/blob/master/tools/src/file_format/analysis.rs for the source and target records rather than using `json!` since that could potentially benefit from additional type-checking and potentially make refactorings easier.  (That said, I don't expect a lot of churn for source/target records, but if structured records were emitted, that does regularly get changed right now.)

tools/src/symbol_graph_edge_kind.rs
```diff
@@ -0,0 +1,12 @@
...
+    Default, // solid line, closed arrow ("normal")
+    // These value are meant to be UML-ish
+    Inheritance,    // solid line, open arrow ("onormal")
+    Implementation, // dashed line, open arrow ("onormal")
+    Composition,    // solid line, closed diamond ("diamond")
+    Aggregation,    // solid line, open diamond ("odiamond")
+    // These are more specific searchfox concepts
+    IPC,           // dotted line, weird vee arrow ("vee")
+    CrossLanguage, // JNI-like; solid line, left-half-closed arrow ("lnormal")
+}
```
arai-a, line 12: This is moved from `cmd_pipeline/symbol_graph.rs`,
in order to remove the dependency from `ontology_mapping.rs` to the remaining of `cmd_pipeline/symbol_graph.rs`.

asutherland: Makes sense, thanks!  That whole cluster of code has evolved in a somewhat haphazard way and likely could benefit from a good spring cleaning! ;)

tools/src/css_analyzer.rs
```diff
@@ -0,0 +1,163 @@
+use cssparser;
+use ustr::{Ustr, ustr};
+
+use crate::file_format::analysis::{
+    AnalysisKind, AnalysisSource, AnalysisTarget, LineRange, Location, SourceRange, SourceTag, TargetTag, WithLocation
+};
+
+// NOTE: This does the same as analysis_manglings::mangle_file without regex
+//       dependency.  regex increases the wasm file size by ~80kB.
```
arai-a, line 9: oh, this was wrong. it's ~800kB instead

tools/src/css_analyzer.rs
```diff
@@ -0,0 +1,163 @@
...
+                s.push(c as char);
+            }
+            _ => {
+                s.push_str(format!("@{:02X}", c as u8).as_str());
+            }
+        }
+    }
+
+    s
+}
```
arai-a, line 25: mangling is not shared because of the increase in the file size of wasm binary, caused by regex, which affects the load time.
  * without regex: 108kB
  * with regex: 890kB

arai-a:
`css-analyze.sh` takes 2s for 1,200 files, with rust-based native binary (it was 28s with the first WASM based patch).
The current binary processes single file.  we could make it batch in followup bug if it needs to be more optimized.

`html-analyze.sh` takes 4min for 42,000 files (it was 2m20s before this patch, and it was 10min with `analysis_manglings::mangle_file` which uses `regex`)

asutherland:
> `css-analyze.sh` takes 2s for 1,200 files, with rust-based native binary (it was 28s with the first WASM based patch). The current binary processes single file. we could make it batch in followup bug if it needs to be more optimized.
> 
> `html-analyze.sh` takes 4min for 42,000 files (it was 2m20s before this patch, and it was 10min with `analysis_manglings::mangle_file` which uses `regex`)

I'm assuming these are local run times; from the current dev channel run, the times seem to come out:
- `css-analyze.sh` is 5 seconds
- `html-analyze.sh` is 20m 17sec?

```
Trees
└── mozilla-central
    └──                                                                 
          script                  time since start   apparent duration  
         ────────────────────────────────────────────────────────────── 
          find-repo-files.py      0:00:00            0:02:41            
          build.sh                0:02:41            0:15:04            
          scip-analyze.sh         0:17:45            0:01:39            
          process-chrome-map.py   0:19:24            0:00:00            
          js-analyze.sh           0:19:24            0:04:30            
          html-analyze.sh         0:23:54            0:20:17            
          css-analyze.sh          0:44:11            0:00:05            
          idl-analyze.sh          0:44:16            0:00:15            
          ipdl-analyze.sh         0:44:31            0:00:18            
          crossref.sh             0:44:49            0:22:25            
          output.sh               1:07:14            0:08:18            
          build-codesearch.py     1:15:32            0:03:49            
          compress-outputs.sh     1:19:21            0:08:15            
          check-index.sh          1:27:36
```

This is against a current on-server run-time of 5min 7sec:
```
├── mozilla-central                                                                                        
│   └──                                                                                                    
│         script                  time since start   apparent duration                                     
│        ──────────────────────────────────────────────────────────────  
│         find-repo-files.py      0:00:00            0:02:35                                               
│         build.sh                0:02:35            0:17:00                                               
│         scip-analyze.sh         0:19:35            0:01:35                                                                                                                                                          
│         process-chrome-map.py   0:21:10            0:00:02                                                                                                                                                          
│         js-analyze.sh           0:21:12            0:04:22                                                                                                                                                          
│         html-analyze.sh         0:25:34            0:05:07                                                                                                                                                          
│         idl-analyze.sh          0:30:41            0:00:15                                                                                                                                                          
│         ipdl-analyze.sh         0:30:56            0:00:19                              
│         crossref.sh             0:31:15            0:21:37                                               
│         output.sh               0:52:52            0:08:32                                               
│         build-codesearch.py     1:01:24            0:03:43                                               
│         compress-outputs.sh     1:05:07            0:05:55            
│         check-index.sh          1:11:02        
```

To be clear, this is fine to land, but we should probably have a plan in mind of how to improve the performance in the medium term.  It's okay if that plan is also just to upgrade to a more powerful AWS instance type!  Although I think we could probably move js-analyze.js to perform some batch processing.  (We don't need to have it be one invocation per processor core if we're worried about the heap getting messy... we could limit invocations to tens or hundreds of files, etc.)

asutherland (approved):
This is really amazing!  It's very exciting to both have this new capability but to also know that you've laid the groundwork for additional magic for CSS and HTML!

Thank you so much for mechanizing the build process and restructuring the rust code to integrate with the existing rust codebase, including the labeling of all the `#[cfg(not(target_arch = "wasm32"))]` annotations.  It seems like a reasonable trade-off to try and avoid duplicative code. (Also, something that would go away if we move to something more rust-based like using https://crates.io/crates/lol_html/ for the HTML and if we farm out the JS indexing to scip-typescript (with the HTML indexer creating transformed JS files for it which then get merged in).)

### ustr digression

Re: the use of `ustr` in the CSS/wasm stuff, it would be okay to make AnalysisSource and AnalysisStructured generic over the string type that's used inside them if you ever think it would be helpful/simplify things.

I was pretty sure I had done that at some point, but I don't see it in tree, so maybe it was something I tried and removed in a patch.  My general rationale has been that it's very useful for `crossref` or any batch processing where we potentially have to hold most of the database in memory all at the same time, but it's potentially quite undesirable for our web-servers since it can inhibit parallelism and we don't want the server process memory use to increase over time.  I believe this is why I frequently have kept things in the `serde_json::Value` rep much of the time.  I think I may have forgotten about the side-effects of Ustr entraining its strings forever more recently though, so it's possible I've been creating a slow quasi-memory-leak in some of the diagram-related code.

arai-a:
I have a patch for batching JS/HTML analysis.
I'll post it shortly.

I agree for making analysis structs genetic over string type. I'll look into it.
To be clear, there was no notable effect on the WASM binary size between using plain serde_json vs `analysis.rs` with ustr.

PR #734, opened 2024-05-23
Ignore anonymous FunctionDeclaration in export default

arai-a:
Fixes "Invalid expression FunctionDeclaration" warnings happening on release


asutherland (approved):
Thanks!

PR #736, opened 2024-05-25
Add ignore_missing_path_prefixes field to repository config

arai-a:


tools/src/file_format/config.rs
```diff
@@ -67,6 +67,9 @@ pub struct TreeConfigPaths {
     /// be mapped into `"__GENERATED__"`.  This will usually be a sub-directory
     /// of the `index_path` but exceptions could be possible.
     pub objdir_path: String,
+    /// List of the path prefixes where files may be missing at the point of
+    /// gathering metadata.
+    pub ignore_missing_path_prefixes: Option<Vec<String>>,
```
asutherland, line 72: Note that it's also possible to use `#[serde(default)]` with Vec without needing to wrap it in an Option.  For [example](https://github.com/mozsearch/mozsearch/blob/4b273b03816dd181c06600e7bf501582ab188493/tools/src/file_format/repo_data_ingestion.rs#L53-L54):
```rust
    #[serde(default)]
    pub dir_prefixes: Vec<String>,
```

This can make it easier to write the consuming code since usually the semantics are the same because it would be weird if the semantics weren't the same.

asutherland (approved):
Thank you for cleaning up all these warnings!

tools/src/file_format/config.rs
asutherland: (to be clear, there's no need to change the existing code)

PR #738, opened 2024-05-26
Perform batch processing in js-analyze.js.

arai-a:
JS shell has `os.file.redirect` which can redirect the `print`/`console.log` call to specified file.

https://searchfox.org/mozilla-central/rev/9508d23ab6de8b0734be2c95303515b08da1ce28/js/src/shell/OSObject.cpp#837-841
```cpp
    JS_FN_HELP("redirect", osfile_redirectOutput, 1, 0,
"redirect([path-or-object])",
"  Redirect print() output to the named file.\n"
"   Return an opaque object representing the previous destination, which\n"
"   may be passed into redirect() later to restore the output."),
```

Also JS shell has `readline` which can read from stdin.

https://searchfox.org/mozilla-central/rev/9508d23ab6de8b0734be2c95303515b08da1ce28/js/src/shell/js.cpp#9266-9268
```cpp
    JS_FN_HELP("readline", ReadLine, 0, 0,
"readline()",
"  Read a single line from stdin."),
```

This patch does the following:
  * Let parallel command to pass the list of files via stdin, with line numbers added, where the line number is the file index for local variables (the file index conflicts between js-analyze vs html-analyze.  I'll address it in another followup)
  * Let js-analyze.js read the list of files from stdin, and write the analysis to specified file instead of stdout
  * rewrite `loadSax` to `ensureSax` to avoid duplicate load
  * reset the global state for each file (maybe we could simplify this by adding global state class or object)

With this, `js-analyze.sh` takes 42 seconds, and `html-analyze.sh` takes 60 seconds on dev indexer instance.

### DIFF
--- a/scripts/html-analyze.sh
+++ b/scripts/html-analyze.sh
@@ -13,7 +13,8 @@ fi
 CONFIG_FILE=$(realpath $1)
 TREE_NAME=$2
 
-cat $INDEX_ROOT/html-files | \
-    parallel --halt 2 js -f $MOZSEARCH_PATH/scripts/js-analyze.js -- {#} \
-    $MOZSEARCH_PATH $FILES_ROOT/{} {} ">" $INDEX_ROOT/analysis/{}
+cat $INDEX_ROOT/html-files | nl -w1 -s " " | \
+    parallel --jobs 8 --pipe --halt 2 \
+    js -f $MOZSEARCH_PATH/scripts/js-analyze.js -- \
+    $MOZSEARCH_PATH $FILES_ROOT $INDEX_ROOT/analysis
 echo $?
--- a/scripts/html-analyze.sh
+++ b/scripts/html-analyze.sh
@@ -13,6 +13,10 @@ fi
 CONFIG_FILE=$(realpath $1)
 TREE_NAME=$2
 
+# Add line number for the file list with `nl`, which is used as a global
+# fileIndex and used for local variable symbols.
+#
+# See the comment in js-analyze.sh for more details.
 cat $INDEX_ROOT/html-files | nl -w1 -s " " | \
     parallel --jobs 8 --pipe --halt 2 \
     js -f $MOZSEARCH_PATH/scripts/js-analyze.js -- \
--- a/scripts/js-analyze.js
+++ b/scripts/js-analyze.js
@@ -548,6 +548,15 @@ let Analyzer = {
   // with empty strings.
   _lines: [],
 
+  resetState() {
+    this.symbols = new SymbolTable();
+    this.symbolTableStack = [];
+    this.nameForThis = null;
+    this.className = null;
+    this.contextStack = [];
+    this._lines = [];
+  },
+
   /**
    * Given a position, find the first instance of the given string starting
    * after the (exclusive, end) position.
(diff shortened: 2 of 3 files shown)


scripts/html-analyze.sh
```diff
@@ -13,7 +13,8 @@ fi
 CONFIG_FILE=$(realpath $1)
 TREE_NAME=$2
 
-cat $INDEX_ROOT/html-files | \
-    parallel --halt 2 js -f $MOZSEARCH_PATH/scripts/js-analyze.js -- {#} \
-    $MOZSEARCH_PATH $FILES_ROOT/{} {} ">" $INDEX_ROOT/analysis/{}
+cat $INDEX_ROOT/html-files | nl -w1 -s " " | \
+    parallel --jobs 8 --pipe --halt 2 \
```
asutherland, line 17: I infer that we're using `nl` here to be able to provide a global fileIndex across all the parallel jobs.  (Aside: This would be a nice thing to have a comment for, but I'm requesting a change here, so it probably won't be needed in the medium term if we eventually end up going with my suggestion.)

At the current time, we don't really need a global fileIndex since they're for "no_crossref" locals.  Also, since we are using the same number space for html-analyze and js-analyze, we already would potentially have collisions if it mattered.

However, I think it probably is potentially useful to try and have unique values here if we were to start doing something that caused locals from different source files to be unified into a single presentation where the SYM_INFO would want to try and have metadata for them.  So I might suggest we try and do [what we do in the C++ indexer](https://searchfox.org/mozilla-central/rev/9508d23ab6de8b0734be2c95303515b08da1ce28/build/clang-plugin/mozsearch-plugin/MozsearchIndexer.cpp#552-553):
```cpp
return std::string("V_") + mangleLocation(Decl->getLocation()) +
       std::string("_") + hash(std::string(Decl->getName()));
```

Our [hash implementation is here](https://searchfox.org/mozilla-central/source/build/clang-plugin/mozsearch-plugin/StringOperations.cpp).

Then we could use `--pipepart -a $INDEX_ROOT/html-files` and avoid the use of `cat` and `nl` here, as [is done in output.sh](https://github.com/mozsearch/mozsearch/blob/2dc26be186f705108cd5e1c750d039d1a7707d60/scripts/output.sh#L64).

I believe in [Bug 1771620](https://bugzilla.mozilla.org/show_bug.cgi?id=1771620) when I had us start using pipepart I did find that there was a surprisingly measurable performance benefit to using pipepart over having the parallel perl script be the one to divvy files up.  Note that I did [subsequently pipe the files through `shuf`](https://github.com/mozsearch/mozsearch/blob/2dc26be186f705108cd5e1c750d039d1a7707d60/scripts/find-objdir-files.sh#L12-L16) as part of writing all-files to disk in [bug 1783761](https://bugzilla.mozilla.org/show_bug.cgi?id=1783761) because it turned out that some parallel tasks were running significantly longer because some parts of the tree had much larger files than other parts of the tree, so shuffling the list helped avoid that while still allowing for use of all-files.

Also, the other output-file stuff about the job logs may or may not be useful; since we're redirecting stdout in js-shell that won't be useful, but since it presumably would catch stderr that might be useful.

asutherland (approved):
Amazing performance improvement!  Thanks so much!

I do have a suggestion for how to further improve things, but it absolutely should not block this landing as-is; it's just something to think about for future improvements.

scripts/html-analyze.sh
arai-a: Thank you for letting me know about the shuffle.  Indeed I see the time vary between jobs.
and yes, using hash instead of index sounds reasonable.
I'll look into them

PR #739, opened 2024-05-26
Make analysis types generic over Ustr/String and use String variant from css_analyzer.

arai-a:
This does the following:
  * Make analysis structs and impl generic over string type (`StrT`), which can be `Ustr` or `String`
  * Make `css_analyzer` use `String` variant of analysis types
  * Move `OntologyPointerKind` into its own file, to avoid annotating `ontology_mapping.rs` much
  * Remove unnecessary dependencies from `Cargo.toml`, `lib.rs`, and `mod.rs`

This reduces the size of `wasm_css_analyzer.wasm` from 106kB to 88kB.

### DIFF
--- a/tools/Cargo.toml
+++ b/tools/Cargo.toml
@@ -9,15 +9,12 @@ tonic-build = "0.7.1"
 
 [dependencies]
 cssparser = "0.29"
-lazy_static = "1.1"
 itertools = "0.10"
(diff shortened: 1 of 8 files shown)


asutherland (approved):
Thanks so much!  And it's great to hear the net result helped decrease the WASM size!

PR #740, opened 2024-05-28
Bug 1899363 - Check LHS vs RHS for CSS custom property, and support @property as declaration

arai-a:
This does the following:
  * Treat custom property names in LHS of ":" as definition
  * Treat custom property names in RHS of ":" as use
  * Treat custom property name after `@property` as declaration


asutherland (approved):
A nice iterative improvement, and with nice new test coverage!

PR #741, opened 2024-05-29
Bug 1899642 - Add ld wrapper to mitigate libtool issue in graphviz

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1899642

The issue here is already explained in 
https://github.com/mozsearch/mozsearch-mozilla/blob/5fdf3904e04c39ad6957f42aeb918d19d742939d/graphviz/build#L12-L15
where libtool passes `-load` and `...*libclang-index-plugin.so` to linker command.

This patch does:
  * Add a wrapper command for ld, which filters these parameters out

graphviz config is going to use this.

### DIFF
--- a/scripts/indexer-setup.py
+++ b/scripts/indexer-setup.py
@@ -23,6 +23,11 @@ flags = [
 ]
 flags_str = " ".join([ '-Xclang {}'.format(flag) for flag in flags ])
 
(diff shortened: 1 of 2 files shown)


asutherland (approved):
Thank you so much for digging into this and coming up with such a clean and elegant fix!

PR #743, opened 2024-05-30
Bug 1895653 - Linkify relative path of import/export declaration

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1895653

This does the following:
  * Add support for dynamic `import()`
  * Use "file " prefix instead of "type " prefix for URL in `js-analysis.js` (this is going to be rewritten by `replace-aliases.py` tho
  * Make `replace-aliases.py` remove any unknown `URL_*` records, given the record is not going to be used by crossref, and it consumes unnecessary RAM
  * Add relative path support, by:
    * Make `js-analysis.js` generate a record with `sym: "RELPATH"` and `pretty` being the relative path
    * Calculate the reverse map of `url-map.json`, that is, file path to urls , in replace-aliases.py
    * Replace `RELPATH` records with corresponding `FILE_*` records, or remove it if there's no any file

I use `RELPATH` symbol instead of `RELPATH_*`, because the symbol is not actually used except for checking if it's relative path.

dynamic import

dynamic-import

relative path

relpath

### DIFF
--- a/scripts/replace-aliases.py
+++ b/scripts/replace-aliases.py
@@ -32,11 +32,10 @@ def replace_aliases(path, alias_map):
             datum = json.loads(line)
             sym = datum["sym"]
 
-            if sym not in alias_map:
-                lines.append(line)
-                continue
-
             has_alias = True
+
+            if sym not in alias_map:
+                continue
             
             for alias in alias_map[sym]:
                 datum["sym"] = alias["sym"]
--- a/scripts/replace-aliases.py
+++ b/scripts/replace-aliases.py
@@ -23,7 +23,7 @@ def resolve_relpath(url, relpath):
         prefix = "chrome:/"
         path = url[8:]
     elif url.startswith("resource://"):
-        prefix = "chrome:/"
+        prefix = "resource:/"
         path = url[10:]
 
     parent = os.path.dirname(path)
--- a/scripts/replace-aliases.py
+++ b/scripts/replace-aliases.py
@@ -3,15 +3,38 @@
 import json
 import glob
 import os
+import re
 import sys
 
 analysis_dir = sys.argv[1]
 alias_map_path = sys.argv[2]
 
 
-def replace_aliases(path, alias_map):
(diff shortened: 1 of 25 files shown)


scripts/replace-aliases.py
```diff
@@ -3,15 +3,38 @@
...
+def at_unescape(text):
+  return re.sub("@[0-9A-F][0-9A-F]", lambda m: chr(int(m.group(0)[1:], 16)), text)
+
+
+def resolve_relpath(url, relpath):
+    if url.startswith("chrome://"):
+        prefix = "chrome:/"
+        path = url[8:]
+    elif url.startswith("resource://"):
+        prefix = "chrome:/"
```
asutherland, line 26: If we're intentionally re-mapping resource:// to chrome:// here and this isn't a typo, then this probably merits a comment.

asutherland (approved):
Very exciting!  I've been having to dig into web platform tests written both in HTML files and JS files and finding myself very jealous of all of these improvements as I have to manually search for the paths of the loaded files[1]! ;)

1: Obviously that's frequently a more complicated scenario, especially for cases like ServiceWorker scripts where the string is a literal that does not start with "./" but is instead a [normal relative path](https://searchfox.org/mozilla-central/rev/ccb101be8f27d837b655ccee51a52c0337dd1eb8/testing/web-platform/tests/service-workers/service-worker/ServiceWorkerGlobalScope/extendable-message-event.https.html#9) and gets passed as an argument to a function in another file that then goes on to call `navigator.serviceWorker.register(scriptRelativePath)`.  Although the path generally does end up being relative to the file it's in, so the same type of heuristics could potentially go pretty far!

asutherland:
Also, thanks for the great contextualizing comments in the pull request and the concise screenshots that demonstrate the functionality at work!

scripts/replace-aliases.py
arai-a: Good catch!
it's just a wrong copy-paste.
I should've added test for resource URI as well

PR #745, opened 2024-06-01
Bug 1900120 - Add Symbol section with symbol selection mechanism, behind Fancy Bar setting.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1900120

This does the following, behind the "Fancy Bar" setting (the "Alpha" gate):
  * Add "Symbol" section inside the navigation bar, above the "Copy as Markdown" section [1]
    * this section shows the currently-selected symbol, which is used by the "Symbol Link" [2]
    * if no symbol is selected (no symbol inside selected line), this section shows "(no symbol clicked)" text
    * this section has copy button, which copies the raw text of the symbol's qualified name (the text part of the current "Symbol Link")
    * the selected symbol's namespace part (including class name etc) and the local name part are shown in separate line, in order to make sure the local name is visible
    * the symbol section has fixed-width (240px), in order to avoid changing the navigation box's size depending on the selected symbol's length
  * Add a new selection state "symbol is clicked"
    * In addition to selecting line numbers, clicking a symbol (non-local symbol) selects the symbol
    * ~~the clicked symbol has orange dotted outline~~
    * the clicked symbol has the same background color as hovered symbol [3]
    * the clicked symbol is kept selected while operating on the navigation box [4]
    * the clicked symbol is de-selected by clicking outside of the navigation box
    * the clicked symbol has higher priority than the current "symbol inside the selected lines".  If a symbol is clicked, it's used for the "Symbol" section and the "Copy as Markdown" section
    * the clicked symbol's line number is reflected to the "Copy as Markdown"'s URL and the "Code Block", even if the line is not selected

Currently available on the dev channel
https://dev.searchfox.org/mozilla-central/source/js/src/frontend/Parser.cpp#1083

[1] Symbol section (without selected symbol)
symbol-section

[2] Symbol section shows the selected symbol
symbol-section-selected

[3] the clicked symbol
clicked-symbol-3

[4] the clicked symbol is kept selected while operating on the navigation box
selected-symbol-3

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -60,7 +60,6 @@ table {
   --syntax-string-color: green;
   --syntax-regex-color: #6d7b8d;
   --syntax-symbol-highlight: yellow;
-  --syntax-symbol-selected-outline: orange;
 
   /* Our red-blue color scheme is from
    * https://colorbrewer2.org/#type=diverging&scheme=RdBu&n=11 and was chosen
@@ -126,7 +125,6 @@ table {
 
     --syntax-comment-color: GrayText;
     --syntax-symbol-highlight: #5d4d1d;
-    --syntax-symbol-selected-outline: orange;
     /* Shamelessly taken from source.chromium.org's dark theme */
     --syntax-type-color: #79a9c4;
     --syntax-reserved-color: #d8884b;
@@ -747,11 +745,9 @@ tr.after-context-line + tr.before-context-line {
 span[data-symbols]:hover {
   cursor: pointer;
 }
-span[data-symbols].hovered {
-  background-color: var(--syntax-symbol-highlight);
-}
+span[data-symbols].hovered,
 span[data-symbols].selected {
-  outline: 1px dotted var(--syntax-symbol-selected-outline);
+  background-color: var(--syntax-symbol-highlight);
 }
 
 /* Help screen */
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -60,6 +60,7 @@ table {
   --syntax-string-color: green;
   --syntax-regex-color: #6d7b8d;
   --syntax-symbol-highlight: yellow;
+  --syntax-symbol-selected-outline: orange;
 
   /* Our red-blue color scheme is from
    * https://colorbrewer2.org/#type=diverging&scheme=RdBu&n=11 and was chosen
@@ -125,6 +126,7 @@ table {
 
     --syntax-comment-color: GrayText;
     --syntax-symbol-highlight: #5d4d1d;
+    --syntax-symbol-selected-outline: orange;
     /* Shamelessly taken from source.chromium.org's dark theme */
     --syntax-type-color: #79a9c4;
     --syntax-reserved-color: #d8884b;
@@ -674,6 +676,37 @@ tr.after-context-line + tr.before-context-line {
   cursor: not-allowed;
 }
 
+.panel .selected-symbol-section {
+  width: 240px;
+  display: grid;
+  grid-template-columns: 200px 30px;
+  gap: 10px;
+}
+
+.panel .selected-symbol-section .selected-symbol-box {
+  position: relative;
+}
+
+.panel .selected-symbol-section .selected-symbol-box .selected-symbol-ns {
+  position: absolute;
+  top: -1em;
+  padding-inline-start: 26px;
+  height: 0.8em;
+  font-size: 0.8em;
+  white-space: pre;
+}
+.panel .selected-symbol-section .selected-symbol-box .selected-symbol-local {
+  padding-inline-start: 30px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: pre;
+}
+
+.panel .selected-symbol-section .copy-box.disabled {
+  opacity: 0.5;
+  cursor: not-allowed;
+}
+
 /* Info Boxes */
 .info-box {
   border: 1px solid rgba(0, 0, 0, .5);
@@ -716,6 +749,9 @@ span[data-symbols]:hover {
 }
 span[data-symbols].hovered {
   background-color: var(--syntax-symbol-highlight);
+}
+span[data-symbols].selected {
+  outline: 1px dotted var(--syntax-symbol-selected-outline);
 }
 
 /* Help screen */
--- a/static/js/code-highlighter.js
+++ b/static/js/code-highlighter.js
@@ -201,10 +201,7 @@ var DocumentTitler = new (class DocumentTitler {
     // The pretty will include a descriptor prefix like "function " which we
     // don't care about.
     if (bestPretty) {
-      let idxSpace = bestPretty.indexOf(" ");
-      if (idxSpace !== -1) {
-        bestPretty = bestPretty.substring(idxSpace + 1);
-      }
+      bestPretty = bestPretty.replace(/[A-Za-z0-9]+ /, "");
 
       // Shorten any namespaces.
       bestShortPretty = this._shortenNamespaces(bestPretty);
@@ -668,12 +665,18 @@ var Highlighter = new (class Highlighter {
     DocumentTitler.processLineSelection(this.selectedLines);
   }
 
-  toHash() {
-    if (!this.selectedLines.size) {
+  // Convert the list of selected lines into a hash string.
+  // Passing extraLine makes that line also selected.
+  toHash(extraLine = undefined) {
+    if (extraLine === undefined && !this.selectedLines.size) {
       return "";
     }
     // Try to create ranges out of the lines.
-    let lines = [...this.selectedLines].sort((a, b) => a - b);
+    const unsortedLines = [...this.selectedLines];
+    if (extraLine !== undefined && !unsortedLines.includes(extraLine)) {
+      unsortedLines.push(extraLine);
+    }
+    let lines = unsortedLines.sort((a, b) => a - b);
     let ranges = [];
     let current = { start: lines[0], end: lines[0] };
     for (let i = 1; i < lines.length; ++i) {
--- a/static/js/context-menu.js
+++ b/static/js/context-menu.js
@@ -6,6 +6,8 @@ var ContextMenu = new (class ContextMenu {
     this.menu.style.display = "none";
     document.body.appendChild(this.menu);
 
+    this.selectedToken = null;
+
     this.menu.addEventListener("mousedown", function (event) {
       // Prevent clicks on the menu to propagate
       // to the window, so that the menu is not
@@ -37,6 +39,16 @@ var ContextMenu = new (class ContextMenu {
   }
 
   tryShowOnClick(event) {
+    if (Settings.fancyBar.enabled) {
+      if (this.selectedToken) {
+        if (!Panel?.isOnPanel?.(event)) {
+          this.selectedToken.classList.remove("selected");
+          this.selectedToken = null;
+          Panel?.onSelectedTokenChanged?.();
+        }
+      }
+    }
+
     // Don't display the context menu if there's a selection.
     // User could be trying to select something and the context menu will undo it.
     if (!window.getSelection().isCollapsed) {
@@ -80,6 +92,12 @@ var ContextMenu = new (class ContextMenu {
 
     let symbolToken = event.target.closest("[data-symbols]");
     if (symbolToken) {
+      if (Settings.fancyBar.enabled) {
+        this.selectedToken = symbolToken;
+        this.selectedToken.classList.add("selected");
+        Panel?.onSelectedTokenChanged?.();
+      }
+
       const symbols = symbolToken.getAttribute("data-symbols").split(",");
 
       const seenSyms = new Set();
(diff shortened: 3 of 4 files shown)


arai-a:
@smaug---- can I have your feedback for the UX?

static/css/mozsearch.css
```diff
@@ -717,6 +750,9 @@ span[data-symbols]:hover {
 span[data-symbols].hovered {
   background-color: var(--syntax-symbol-highlight);
 }
+span[data-symbols].selected {
+  outline: 1px dotted var(--syntax-symbol-selected-outline);
```
asutherland, line 754: I would suggest upping this to 3px.  Especially with the current light mode yellow highlighted symbol, it's very hard to notice this at 1px and still not much better at 2px, but at 3px I think it becomes obvious.

asutherland (approved):
This is pretty great!

I really like the aesthetic of how you dealt with the namespaces.  I do think the font size may be on the small side as a result, but I think it makes sense given the current limitations on how much space we give to the navigation panel and it is important to both literally fit in to the panel as well as figuratively; having a larger font size there could be distracting.  If we overhaul things to be more like :kats' mock-up, I think we can potentially enlarge the font or potentially change it to a serif font or something.

Also I think it does help that the namespace is usually going to be boring / very predictable values so peoples' brains will already be able to limit the search space; it's not like it's an eye chart with effectively random values.

Given that this is feature gated, I think it's fine to land this whenever you want, although if @emilio or :smaug would like to take a deeper look before that, they should feel free to chime in briefly that this shouldn't land yet and then delve in more when they have available time.

asutherland:
Also, thanks again for the really great pull request descriptions that provide incisive explanations of what/how the feature works and screenshots of it at work!

static/css/mozsearch.css
arai-a: Good point that it's not obvious, but I'm slightly hesitant to increase the outline width, given it hides the enclosing context.
Maybe we can use background color instead of outline?
Possibly the same color as hover (so, the clicked one remains with the highlight background color while operating on the navigation box), or a different color if it needs distinction between hover vs "no longer hover but selected" (so, the clicked one becomes "hovered" color, and it turns into different color while operating on the navigation box)?

asutherland: Ah, yeah, using the same color as the hover helps eliminate a concern about it potentially standing out against the existing highlight.  Then it basically only needs to stand out against the normal background, so the width becomes less important.  I also don't have a particularly strong opinion here and people can easily experiment with tweaking the CSS styling locally in devtools if they have opinions.

arai-a: Okay, updated the PR and dev channel to use the same styling as hovered.

smaug----:
I would have expected the context menu to have 'copy' for the symbol.  Having it on the right side forces one to move eyes quite a bit. 
Profiler has similar-ish feature, and it is in the context menu.

arai-a:
Thank you for your feedback!

> I would have expected the context menu to have 'copy' for the symbol. Having it on the right side forces one to move eyes quite a bit. Profiler has similar-ish feature, and it is in the context menu.

Yeah, that was previous attempt.
https://github.com/mozsearch/mozsearch/pull/744

And it was abandoned because:
  * The context menu needs to stay minimal (https://github.com/mozsearch/mozsearch/pull/744#pullrequestreview-2091173924)
  * The navigation box already has the "Copy as Markdown" functionality, which is beneficial also for the clicked symbol
  * Navigation box can provide more functionality for copy, such as omitting namespace etc

Maybe we could resurrect it behind a setting, for the raw text copy purpose?

asutherland:
Quoting @smaug---- from https://github.com/mozsearch/mozsearch/pull/745#issuecomment-2143880878
> I would have expected the context menu to have 'copy' for the symbol. Having it on the right side forces one to move eyes quite a bit. Profiler has similar-ish feature, and it is in the context menu.

I'm not as concerned about the eyes situation as much as the [Fitts' law aspects](https://en.wikipedia.org/wiki/Fitts%27s_law) for mouse movement.  That said, the idiom for context menus appropriately does not assume that users will click the button and the need to search the rest of the screen for other actions they can take.

I think something worth noting is that the profiler generally only pops up context menus for a single symbol at a time and the options it provides are largely invariant so the user does not need to scan the menu.  In contrast, searchfox's context menus are data-dependent and need to be scanned, although we try and make this easier with icons, predictable section delineations, and predictable text patterns.  This is in contrast to something like source.chromium.org where all of the data we put in the context menu (except a single def) instead goes in the bottom panel.

Here's a screenshot of the profiler right-click menu from the call tree for a C++ symbol: 
image

Quoting @arai-a from https://github.com/mozsearch/mozsearch/pull/745#issuecomment-2143887106
> Yeah, that was previous attempt. #744
> 
> And it was abandoned because:

So I think this maybe conflates a few things.  Copying the pretty identifier without a link is, I think, significantly less useful than the symbol link affordance we already have.  So if we only add a single copy option to the context menu, it shouldn't be the raw symbol text, it should be along the lines of the markdown "symbol link" which works in all of our communication platforms (matrix, slack, bugzilla, github).

We can add more things to the context menu, but I think we do need to be careful; here's my VS code context-menu for working in m-c right now with 28 menu items and 7 separators, which I think many people might feel is a bit overwhelming: 
image

Something that I think I've proposed before that might be relevant here is if we were to add "copy" segments to the right side of the context menus.  So `[ Go to definition of Foo ]` would become `Go to definition of Foo | (Copy Icon) ]` where the copy icon is its own link/click target.  We could potentially have a setting to add more segments so like we'd have `| (copy icon)(markdown icon) | (copy icon)(raw icon)]`.

That would probably imply adding some kind of additional context menu item for the current token/line to make it possible to generate a link to that line.  That's a little weirder because having the large hit area be a hyperlink to the current position is maybe weird when the user's only desired action would be to copy it.  Also, currently we go out of our way to avoid emitting any jump links if the token is already on the line we would go to.

But maybe the main hit area for that line could be a concise example of what the markdown phrasing would be?  Like `[Copy: "Use of Foo in Bar" (markdown icon) | (copy icon)(raw icon) ]`.  When I'm explaining things in Bugzilla, I like to make it clear that I'm referencing a use and in what context.

Some other things that we could consider, including combinations of them:
- We currently do our context-menu on left-click, which is weird.  If we stick with that forever, we could also do something on right-click, and arguably that's a reasonable place to find a bunch of copy options.
- We don't currently have submenus, but we could have submenus.  So we could have "Copy..." which follows the normal context menu sub-menu idiom.
- Also we could do something less idiomatic and more custom, but "Copy..." could result in a pop-up that shows options with previews of what they look like.
- Add a horizontal row of buttons to the context menu like how right-clicking on web page content in Firefox has `[ (Back Icon) ][ (Forward Icon) ][ (Reload Icon) ][ (Bookmark Icon) ]`.  This might be a little jarring to muscle memory if we put it at the top, but if we put it in one of the lower sections that might actually help people scan the context menu, like if we put the diagram stuff below it.

asutherland:
One thing I should say is that, if we have settings, something that's more work that we could do longer term to help surface the settings is do something like gmail's action bubble popup things that show up after a user takes an action and that provides an interface for undoing the action.  Treeherder also does a version of this when UI actions are taken, but they usually don't have any affordance to take additional action and they tend to be large and stack up in a way that's more frustrating than helpful, but that may also be familiar.

We wouldn't need to undo any actions, but the idea would be the pop-up could provide effectively a preview of what was copied and provide a "change my settings for copy and paste" link which would then link to the settings UI and options to change what the buttons do or how they work or stuff like that.  So if someone sometimes wants to copy markdown and sometimes wants to copy raw links or raw symbol names, they wouldn't need to figure out that there is a setting and how to get to settings.  And we wouldn't need to surface all options in the UI by default.  Instead the pop-up thing would be like "this is what we just did, if you wanted something different, click this link to change stuff".

Again, that would just be a long term plan, but I think this is an area where people could have a bunch of personal preferences and where having settings to support this stuff wouldn't necessarily result in a combinatorial explosion that results in an implementation and testing nightmare.

asutherland:
I also do want to emphasize that we don't need to figure this all out right now.  Part of iterating can be landing things and living with them a bit.  As long as they're behind a feature gate that people can turn off, our bar to landing things can be low because they won't interfere with people's muscle memories unless they have opted in.

arai-a:
For Copy in context menu, DevTools Inspector has a real world example, which has Copy item and a sub menu to specify the styling and data:

devtools-copy2

And I agree that designing a context menu behavior needs more iterations and it will take long time,
I guess we could repurpose this PR sorely for selection and contextualization, and thinking about copy functionality as followup.
(to be clear, I'll leave the copy button there, given having it there won't harm)
The selected symbol is currently shown in the page title, but nowadays the page title is almost invisible (the tab only shows 2.5 characters for the current tab for me, when there are many tabs), and having dedicate space will be helpful anyway.

smaug----:
I'm just a bit worried making the all the time visible UI complicated. Simplicity is a key reason why searchfox is so good.

> As long as they're behind a feature gate that people can turn off,

Or do you mean feature gate which people can turn on? I wouldn't want to see "Symbol" on the right side all the time by default.)

arai-a:
All the changes (including the "clicked symbol" behavior) are behind "Fancy Bar" feature gate, which is enabled by default when the default gate is set to "Alpha".
iiuc, there's no other thing implemented behind Fancy Bar, so, setting "Fancy Bar" to "Release" or "Beta" will disable the feature.

fancy

In the future, for example we could make each section collapse-able with persistence across pageloads, so that sections such as "Symbol" and "Copy as Markdown" can be hidden for each, while showing the "Revision Control" etc.

asutherland:
Quoting https://github.com/mozsearch/mozsearch/pull/745#issuecomment-2144892651
> Or do you mean feature gate which people can turn on? I wouldn't want to see "Symbol" on the right side all the time by default.)

@arai-a addressed how it works, but to clarify my phrasing: Yes, I meant turn on, and then turn off again if they want.  In particular, since it's quite possible many people set the default feature gate to alpha to get diagrams, they may now have fancy bar enabled, in which case they need to turn it off if they want by setting the fancy bar gate to release or beta or change the default gate to release or beta and then turn the diagram feature to alpha.

asutherland:
Quoting @smaug---- in https://github.com/mozsearch/mozsearch/pull/745#issuecomment-2144892651
> I'm just a bit worried making the all the time visible UI complicated. Simplicity is a key reason why searchfox is so good.

Very much agreed.  (Which is also part of my concern about adding too many context menu items.)

Quoting @arai-a in https://github.com/mozsearch/mozsearch/pull/745#issuecomment-2144974497
> In the future, for example we could make each section collapse-able with persistence across pageloads, so that sections such as "Symbol" and "Copy as Markdown" can be hidden for each, while showing the "Revision Control" etc.

Agreed that making sections collapsible is an important capability for people to feel in control of the UI.  I think it it also makes sense for the fancy bar to support a mechanism akin to Eclipse Perspectives ([old brief video](https://www.youtube.com/watch?v=q6lqzKG63zA)).  I've mentally been referring to that capability as a "lens" since I think it could potentially also impact the source listing view and the gutters.

For example, we've had requests like in [Bug 1545151](https://bugzilla.mozilla.org/show_bug.cgi?id=1545151) to enable having the blame sidebar color vary based on the age of the changes.  While there's discussion in the bug on how to do that in a subtle fashion, the reality is that if someone is doing code archaeology, that's a different use case from code understanding, and it could be useful for the blame strip's visual encoding to be much more noticeable/distracting in that case.  But having that on all the time would be distracting, so only having it enabled when desired would be good.

Related to that, we wouldn't want users to constantly have to manually toggle features individually, especially since coordinating the settings between multiple tabs could be a bit of a nightmare.  So the lens/preset mechanism could make switching between modes of operation easier and make sure that toggles are specific to the lens in question.  (Although we could have a gear icon menu that would let the user potentially collapse or remove a widget they don't like across all modes.)

We could also do some tricks to build on SessionStorage to help with the multiple open tabs case that leverages the fact that when a new tab is opened from an existing tab, its SessionStorage should be cloned.  This would let the lens be inherited, or potentially if we create a session identifier that is what is inherited, it could let us then have all those tabs share state by keying them by that identifier.

The default lens could be a "classic" minimal lens that looks like the current (release) navigation panel.

asutherland:
> And I agree that designing a context menu behavior needs more iterations and it will take long time, I guess we could repurpose this PR sorely for selection and contextualization, and thinking about copy functionality as followup. (to be clear, I'll leave the copy button there, given having it there won't harm) The selected symbol is currently shown in the page title, but nowadays the page title is almost invisible (the tab only shows 2.5 characters for the current tab for me, when there are many tabs), and having dedicate space will be helpful anyway.

Yeah, again, I think it's fine to land this as-is behind the feature gate so at least you and I can try living with it for a while.  I should also note that while :smaug's opinions are deeply valued and his input is always considered, it's not the final word; for example, at least in the past he did not agree with the emphasis given to the test info boxes (at least in the past), filing a bug then-titled ["Warnings like "Test Info: Errors" clutter the main UI"](https://bugzilla.mozilla.org/show_bug.cgi?id=1655952).  Given our epidemic of disabled and flaky tests and how many people seem to completely ignore the navigation panel, I think the current in-content-stream positioning is correct.

arai-a:
Okay, I'll land this and leave the navigation box design and context menu design to followup bugs.

PR #746, opened 2024-06-03
Bug 1900176 - Show holes in class layout

arai-a:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1900176

This patch does:
  * Apply the new table design based on https://clicky.visophyte.org/files/screenshots/20200331-020445.png , but with minor differences
    * Show single table with borders
    * The column header is shown at the top, shared across all classes
    * Each class name is shown in cell, with colspan
  * Add a separate step to collect and process field layout data
    * Platforms are grouped with the entire set of fields across class hierarchy
      * This is necessary to show padding and holes between super/sub classes
    * Stable order is applied to the platform order
    * Fields are grouped/merged based on the field name, instead of line number
      * This solves the issue where fields defined inside a macro is merged
    * Fields are ordered based on the line number, average offset across platforms, and which platforms the field exists
      * This achieves the stable order even if fields are defined inside a macro
    * Calculate the holes between fields, the padding at the end of class, based on the previous field and the class size 
  * Add more info to the table
    * Show holes between fields
    * Show the padding at the end
      * Questions: Currently the order of classes remains the same, which shows the sub class above super class.  It's looks like pahole also does so.  But if we show hole and padding above/below fields, "the hole after super class" and "the padding at the end of super class" might be confusing, because the last field of super class is shown after the first field of sub class.  Do you think it's better ordering classes based on the offset?
    * Show all type variants for the field
    * Show bit field position

Example tables:

Class hierarchy with holes and end padding:
table-subclass

Show all type variants for the field, concatenating with "|"
table-types

Bit field position:
table-bitfields


arai-a:
@hotsphink This is the update for the class layout we talked in the previous meeting. Can I get your feedback on the formatting?

asutherland (approved):
Holy cow, this is so amazing!  You've really taken this feature and its quality to the next level!  Wooooo!

I've mainly just given this a skim, my thoughts are mainly:
- I very much appreciate the comments throughout!
- Really excellent test coverage!
- I'm glad that the general approach I landed on for the table generation in rust with a thin-ish layer of Liquid templating seems to have worked out!
  - In general, the liquid templating hasn't been as ergonomic as I hoped and I've really had to lean on the https://github.com/mozsearch/mozsearch/blob/master/docs/liquid-templating-cheatsheet.md I wrote up when it comes to the [truthy / falsy / empty / blank](https://github.com/mozsearch/mozsearch/blob/master/docs/liquid-templating-cheatsheet.md#truthy--falsy--empty--blank) bits.  But it still potentially provides a way for per-tree customization and I'm not sure that having the HTML translation in rust would necessarily be that much more ergonomic.  That said, I think there's definitely much more work happening in templating-in-rust for various web frameworks than there is on liquid.
    - Which is mainly to say, I'm not wedded to the use of liquid, although if we were to consider moving away, it's probably still worth considering whether the ability to use a liquid JS runtime to render some things to HTML on the client is useful or not.  (That said, we could always compile rust to WASM and run it in a SharedWorker.)

Re: Questions:
> Currently the order of classes remains the same, which shows the sub class above super class. It's looks like pahole also does so. But if we show hole and padding above/below fields, "the hole after super class" and "the padding at the end of super class" might be confusing, because the last field of super class is shown after the first field of sub class. Do you think it's better ordering classes based on the offset?

I think the current reverse ordering works out better for worst-case scenarios like [field-layout:'nsGlobalWindowInner'](https://searchfox.org/mozilla-central/query/default?q=field-layout%3A%27nsGlobalWindowInner%27).  But that's mainly because it gets us to the class of interest immediately.  If we could do something semi-clever like automatically scrolling down to the class of interest and it didn't seem too janky, that would probably be okay because mentally it does feel more right (and more useful) for things to be in absolute offset order.

Ideally this would also do something with the focus or otherwise help ensure screen reader users are able to start from the class of interest as well.  I wouldn't want to reverse the class order without making sure that the screen reader behavior is acceptable.  I believe we have a pretty active accessibility channel on slack who might be able to provide input; I also needinfo :jteh on bugs and/or try investigating stuff with NVDA myself.  Presumably table mode would be used for navigation here, just like for source listings.

hotsphink:
(Nightly just crashed on me when I tried to post this comment. Rewriting it now...)

My main feedback is "this is awesome! Land it quick!"

But I can give lots of bikeshedding-style feedback if you like:

- I usually prefer offsets to be in hex, with a "0x" prefix. (Though not always; sometimes decimal is more convenient.) But that does mean you have to agonize over leading zeroes (0x4 vs 0x004): don't use them? Look ahead and use the same number for all values?
- I think I prefer offset order instead of leaf class first, even without the mitigations that @asutherland suggested. I find myself looking for fields in base classes just as often as fields in the leaf class, and it makes padding easier to see and understand.
- It's a little confusing to understand what a class is (the leaf class vs a base class / superclass). Maybe consider appending " (base class)" or " (superclass)" to the class name row?
- (maybe?) consider giving the offset of a base class to the class name. It's redundant with the offset of the first field, but it kind of helps nudge the reader to remember that base classes are stored in memory before the subclass.
- consider putting padding and holes in a different font
- (maybe?) consider treating padding as an "unnamed field" so that its display is more similar to fields, with an offset and size
- I'm guessing the platform-varying types might be a little hard to spot. Consider adding a line break after the vertical bar.
- (maybe?) consider using @ as a unifying symbol for offsets, to keep from getting lost in a large table

I think I had a few other suggestions, but that's all I can remember now. None of these are necessary; your existing version looks very good. I mocked up many of these suggestions in HTML. I'll try to upload it here (not sure if that works), though I think that may be what crashed my browser last time. The HTML file has 3 variants: the first is with the leaf class first, the second is in offset order, and the third shows padding as an unnamed field. The third one also tries adding a line break between the types of a varying type field. (I don't know what's easy/hard/possible in Liquid, I'm just mocking with HTML.) Hm, I guess there would be another padding line in the macosx64 variant in that case...

[layout.html.gz](https://github.com/user-attachments/files/15538184/layout.html.gz)

Question: what happens when the class hierarchy varies by platform? (I think the layout should be optimized to assume that it doesn't, but I'm wondering what happens. Though if that never shows up in Gecko, maybe we don't care.)

hotsphink:
One other comment I just remembered: I wasn't sure whether "bits 3+4" meant 4 bits starting at bit 3, or bit 3 and bit 4. "bits 3..6" would make more sense to me, except that I don't know of any universal convention for inclusive vs exclusive ranges. Maybe "bits 3..6 (4 bits)"? Rust's "bits 3..=6" is too unfamiliar to help, sadly.

(I see from the "bits 3+2" example that you meant a starting bit and count.)

arai-a:
> I think the current reverse ordering works out better for worst-case scenarios like [field-layout:'nsGlobalWindowInner'](https://searchfox.org/mozilla-central/query/default?q=field-layout%3A%27nsGlobalWindowInner%27). But that's mainly because it gets us to the class of interest immediately. If we could do something semi-clever like automatically scrolling down to the class of interest and it didn't seem too janky, that would probably be okay because mentally it does feel more right (and more useful) for things to be in absolute offset order.
> 
> Ideally this would also do something with the focus or otherwise help ensure screen reader users are able to start from the class of interest as well. I wouldn't want to reverse the class order without making sure that the screen reader behavior is acceptable. I believe we have a pretty active accessibility channel on slack who might be able to provide input; I also needinfo :jteh on bugs and/or try investigating stuff with NVDA myself. Presumably table mode would be used for navigation here, just like for source listings. [these comments cover a bit how I previously tested source listing changes with NVDA](https://bugzilla.mozilla.org/show_bug.cgi?id=1511275#c9).

Yeah, that's good example.
I'll keep the current order for now.
Maybe we could change the order dynamically on client side,
along with other client-side feature like https://maraneshi.github.io/HLSL-ConstantBufferLayoutVisualizer/ mentioned in the bug comment.

> * I usually prefer offsets to be in hex, with a "0x" prefix. (Though not always; sometimes decimal is more convenient.) But that does mean you have to agonize over leading zeroes (0x4 vs 0x004): don't use them? Look ahead and use the same number for all values?

Yeah, forgot to mention that this PR changes the offset/size from hex to decimal, simply because the mock used decimal.
I'll change the offset back to hex.

> * I think I prefer offset order instead of leaf class first, even without the mitigations that @asutherland suggested. I find myself looking for fields in base classes just as often as fields in the leaf class, and it makes padding easier to see and understand.

Okay, I'll look into making it possible to change the order.

> * It's a little confusing to understand what a class is (the leaf class vs a base class / superclass). Maybe consider appending " (base class)" or " (superclass)" to the class name row?

Makes sense.
I'll add the suffix.

> * (maybe?) consider giving the offset of a base class to the class name. It's redundant with the offset of the first field, but it kind of helps nudge the reader to remember that base classes are stored in memory before the subclass.

Yeah, actually, I forgot to add the class size to the table (which pahole shows).
We could add the offset+size for class itself.
I'll look into it as followup, given the offset and size can vary across platforms, and appending as text won't work well.

> * consider putting padding and holes in a different font

Yeah, I was actually thinking using different color, but didn't implement it because relying on color might not work well.
The italic style looks good. I'll use that.

> * (maybe?) consider treating padding as an "unnamed field" so that its display is more similar to fields, with an offset and size

I feel making the hole/padding look more different than the regular field is better, especially when it can vary across platform.
having "3 bytes hole/padding" inline will make it clearer which platform group has it.

> * I'm guessing the platform-varying types might be a little hard to spot. Consider adding a line break after the vertical bar.

okay, I'll do that.

> * (maybe?) consider using @ as a unifying symbol for offsets, to keep from getting lost in a large table

Good point.
That will make it clearer which is size and which if offet.
I was thinking utilizing double border or dotted to emphasize the offset+size, but having prefix sounds clearer and also easy to implement.

> I think I had a few other suggestions, but that's all I can remember now. None of these are necessary; your existing version looks very good. I mocked up many of these suggestions in HTML. I'll try to upload it here (not sure if that works), though I think that may be what crashed my browser last time. The HTML file has 3 variants: the first is with the leaf class first, the second is in offset order, and the third shows padding as an unnamed field. The third one also tries adding a line break between the types of a varying type field. (I don't know what's easy/hard/possible in Liquid, I'm just mocking with HTML.) Hm, I guess there would be another padding line in the macosx64 variant in that case...
> 
> [layout.html.gz](https://github.com/user-attachments/files/15538184/layout.html.gz)

Thank you for the good example!

> Question: what happens when the class hierarchy varies by platform? (I think the layout should be optimized to assume that it doesn't, but I'm wondering what happens. Though if that never shows up in Gecko, maybe we don't care.)

Currently it's not supported, and we only traverse the main platform's hierarchy.
we could traverse variants' `supers` field maybe, but that's for followup.

> One other comment I just remembered: I wasn't sure whether "bits 3+4" meant 4 bits starting at bit 3, or bit 3 and bit 4. "bits 3..6" would make more sense to me, except that I don't know of any universal convention for inclusive vs exclusive ranges. Maybe "bits 3..6 (4 bits)"? Rust's "bits 3..=6" is too unfamiliar to help, sadly.
> 
> (I see from the "bits 3+2" example that you meant a starting bit and count.)

yeah, I've use "3+2" as, "starting from 3rd bit from MSB, 2 bits width",
but actually the "3" should be part of "Offset", not "Size".

I'll replace it with the following style.

| Offset          | Size   |
|-----------------|--------|
| @ 0x10 + 3 bits | 2 bits |

asutherland:
> Yeah, forgot to mention that this PR changes the offset/size from hex to decimal, simply because the mock used decimal. I'll change the offset back to hex.

Ah, sorry, it was mainly the improved column separation I was pointing out.  I also think hex is preferable (which is why it's what landed in searchfox), so we're all on the same page here.

> > Question: what happens when the class hierarchy varies by platform? (I think the layout should be optimized to assume that it doesn't, but I'm wondering what happens. Though if that never shows up in Gecko, maybe we don't care.)
> 
> Currently it's not supported, and we only traverse the main platform's hierarchy. we could traverse variants' `supers` field maybe, but that's for followup.

More detail on this in:
- the initial field layout implementation in https://github.com/mozsearch/mozsearch/pull/692 specifically:
> Something this absolutely reveals is that the structured "variants" mechanism where we hash on content is causing duplicates. Additionally, this raises the issue that we really need to the root structured rep to be a composite of all variants, not just one arbitrary one. Otherwise the diagram functionality and anything else that is trying to treat the root structured object as canonical is potentially going to be misleading.
- the latter part of https://bugzilla.mozilla.org/show_bug.cgi?id=1896441#c3 starting from "An interesting nuance"

I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1900457 since it's not a particularly trivial undertaking and this will likely be something that comes up multiple times in the future so it's handy to be able to reference it.  In the meantime I expect this falls under "don't do that!" ;)

arai-a:
While testing the updated behavior with complex tests, I noticed that the offset value for classes used in the multiple inheritance context can have wrong value, because the offset can differ for each consumer, and the offset value specified in the analysis record could be from different consumer's case.

in https://searchfox.org/mozilla-central/query/default?q=field-layout%3A%27nsGlobalWindowInner%27
`nsIGlobalObject::mHostObjectURIs` has 0ffset `0x08` on macosx64, but locally it's `0x110` for `nsGlobalWindowInner`.
and there are many fields starting from `0x08` in the table.

In my patch, this results in showing padding and hole in unexpected places, given the calculation doesn't work well if there's overlap between classes, which makes the field interleave each other in the list of all fields.

We'll need to encode the field offset in different way, for example:
  * class's `structured` record has a offset of the class itself for the case when it's the leaf
  * (optionally) each field has offset within the class, (or calculate the class-local offset in the consumer side)
  * `supers` item has the offset of each super class

for now, I'll try to make it detect such case and show warning.

arai-a:
Added warning message:

warning-for-multiple-inheritance

it's quick workaround until https://bugzilla.mozilla.org/show_bug.cgi?id=1900465 gets fixed.

> the offset value specified in the analysis record could be from different consumer's case.

this was slightly wrong.
it's simply the value of when the class is the leaf.
we need to add a offset to all fields when it's 2nd+ super class.

asutherland:
The warning seems like a great compromise to get this awesome improvement out to users sooner!

PR #748, opened 2024-06-05
Bug 1900692 - Support multiple inheritance in the class field layout table

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1900692

This does the following:
  * Sync clang plugin with m-c, including [bug 1900465](https://bugzilla.mozilla.org/show_bug.cgi?id=1900465)
  * Cleanup the variant handling not to rely in "main vs variants" groups
    * populate the list of all platforms appears in the class hierarchy as first step
    * perform the remaining steps based on the list of all platforms
  * Reflect super class's offset to the field offset, for multiple inheritance case
  * Instead of tracking the class size, track the end offset of the class, so that it reflects the class offset as well
  * Add `TraveralId` to handle a class that appears multiple times in the class hierarchy, for multiple instance case
  * Reflect super classes in all variants

Result with nsGlobalWindowInner

field-layout-nsGlobalWindowInner

### DIFF
--- a/clang-plugin/MozsearchIndexer.cpp
+++ b/clang-plugin/MozsearchIndexer.cpp
@@ -1113,6 +1113,14 @@ public:
 
         J.attribute("sym", getMangledName(CurMangleContext, BaseDecl));
 
+        if (Base.isVirtual()) {
+          CharUnits superOffsetBytes = Layout.getVBaseClassOffset(BaseDecl);
+          J.attribute("offsetBytes", superOffsetBytes.getQuantity());
+        } else {
+          CharUnits superOffsetBytes = Layout.getBaseClassOffset(BaseDecl);
+          J.attribute("offsetBytes", superOffsetBytes.getQuantity());
+        }
+
         J.attributeBegin("props");
         J.arrayBegin();
         if (Base.isVirtual()) {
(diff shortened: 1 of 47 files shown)


asutherland (approved):
Super nice!  Thanks for the amazing patch stack factoring and the introduction of ClassId and FileId to improve readability and clarity!  The well structured stack really makes it easy to review the changes (although I should emphasize I gave this a deep skim but did not do a full walkthrough of all logic).  This definitely gives me a quality of formulation of patches to aspire to for when I'm next able to actively do some enhancement work!

Other thoughts:
- Oh, huh, I hadn't realized the bug 1897647 changes weren't in this tree.  Whoops!

asutherland:
One string localization nit from the existing hole description logic that I think was disabled for the multiple inheritance case so I didn't notice it is that I think normally "4 bytes hole" would probably instead be phrased as "4 byte hole" and that works for all plurals, but it also doesn't matter.

arai-a:
Thank you for reviewing!

pahole uses "bytes" for plurals, so I'll leave it to followup

```
struct Sub : Base {
        /* struct Base                <ancestor>; */     /*     0    12 */

        /* XXX last struct has 3 bytes of padding */
        /* XXX 4 bytes hole, try to pack */

        uint64_t                   d;                    /*    16     8 */
...
};
struct Base {
        uint8_t                    a;                    /*     0     1 */

        /* XXX 3 bytes hole, try to pack */

        uint32_t                   b;                    /*     4     4 */
        uint8_t                    c;                    /*     8     1 */
...
        /* padding: 3 */
...
};


```

PR #749, opened 2024-06-05
Bug 1900823 - Linkify resource:// and chrome:// URLs in textual occurrences

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1900823

This does the following:
  * Fix the chrome-map handling (basically this reverts https://github.com/arai-a/mozsearch/commit/98654da28d200dd7dbf106c76d29602b3fc5033b .  the code was correct but the test data was wrong)
  * Make the output step to generate a `FILE_` symbols span for `chrome://` and `resource://` URLs inside comments, without reflecting it to `SYM_INFO`
  * Make the context menu handling to generate pseudo symbol info for unknown `FILE_` symbols

This linkifies URLs in `jar.mn` , and it discovered the chrome-map handling bug.

link-chrome

### DIFF
--- a/tools/Cargo.toml
+++ b/tools/Cargo.toml
@@ -56,6 +56,8 @@ reqwest = "0.11.3"
 rls-analysis = "0.18.1"
 rls-data = "0.19.1"
 scip = "0.3.3"
+# NOTE: serde_json dependency is also defined above, without "std" feature.
+#       The "std" feature should be enabled only for non-wasm case.
 serde_json = { version = "1.0.113", features = ["preserve_order", "std"] }
 shell-words = "1.0.0"
 tokio = { version = "1.6.0", features = ["rt-multi-thread", "net", "macros", "fs", "io-util", "signal"] }
--- a/tools/Cargo.toml
+++ b/tools/Cargo.toml
@@ -56,6 +56,7 @@ reqwest = "0.11.3"
 rls-analysis = "0.18.1"
 rls-data = "0.19.1"
 scip = "0.3.3"
+serde_json = { version = "1.0.113", features = ["preserve_order", "std"] }
 shell-words = "1.0.0"
 tokio = { version = "1.6.0", features = ["rt-multi-thread", "net", "macros", "fs", "io-util", "signal"] }
 tokio-stream = "0.1.8"
--- a/tools/src/url_map_handler.rs
+++ b/tools/src/url_map_handler.rs
@@ -12,7 +12,7 @@ pub fn set_url_map_path(path: &str) {
 pub fn get_file_paths_for_url(url: &str) -> Option<Vec<URLMapItem>> {
     lazy_static! {
         static ref URL_MAP: URLMap = unsafe {
-            read_url_map(URL_MAP_PATH.as_ref().unwrap().as_str())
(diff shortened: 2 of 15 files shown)


tools/Cargo.toml
```diff
@@ -56,6 +56,7 @@ reqwest = "0.11.3"
 rls-analysis = "0.18.1"
 rls-data = "0.19.1"
 scip = "0.3.3"
+serde_json = { version = "1.0.113", features = ["preserve_order", "std"] }
```
asutherland, line 59: what about the copy of this [up on line 16](https://github.com/mozsearch/mozsearch/blob/c2a2e4a6476a4fd278953876072631eac3cc23e4/tools/Cargo.toml#L16)  that doesn't include "std" in the features?

tools/src/url_map_handler.rs
```diff
@@ -0,0 +1,21 @@
...
+}
+
+pub fn get_file_paths_for_url(url: &str) -> Option<Vec<URLMapItem>> {
+    lazy_static! {
+        static ref URL_MAP: URLMap = unsafe {
+            read_url_map(URL_MAP_PATH.as_ref().unwrap().as_str())
+        };
+    }
+
+    let sym = format!("URL_{}", mangle_file(url));
```
asutherland, line 19: should this be `FILE_`?

asutherland (approved):
> without reflecting it to `SYM_INFO`

So I guess this potentially works for non-HEAD revisions too since we're not depending on having access to jumprefs?

I do wonder if it makes sense to add a comment on `get_file_paths_for_url` or to the linkify call about how we could add the jumprefs resulting from linkified text with extra plumbing.  We'd want this in the case we ever start wanting to traverse edges from the FILE in the [crossref_converter.rs](https://github.com/mozsearch/mozsearch/blob/c2a2e4a6476a4fd278953876072631eac3cc23e4/tools/src/file_format/crossref_converter.rs) logic to offer extra options on the context-menu.

tools/src/url_map_handler.rs
arai-a: This is the key in the `url-map.json` data, which is a map from `URL_*` symbol to file data (pretty string and `FILE_*` symbol).
I'll add comment.

asutherland: Oh, right, I totally missed the next line was doing the lookup transform.  A comment certainly wouldn't hurt but this was just a case of me going through things too fast.

arai-a:
> So I guess this potentially works for non-HEAD revisions too since we're not depending on having access to jumprefs?

Good point about non-HEAD revisions.
actually this patch broke that feature, given the `URL_MAP_PATH` is not set.
I'll push fix shortly.

> I do wonder if it makes sense to add a comment on `get_file_paths_for_url` or to the linkify call about how we could add the jumprefs resulting from linkified text with extra plumbing. We'd want this in the case we ever start wanting to traverse edges from the FILE in the [crossref_converter.rs](https://github.com/mozsearch/mozsearch/blob/c2a2e4a6476a4fd278953876072631eac3cc23e4/tools/src/file_format/crossref_converter.rs) logic to offer extra options on the context-menu.

okay, I'll add comments.

PR #750, opened 2024-06-06
Bug 1900926 - Show vtables in the field layout.

arai-a:
https://bugzilla.mozilla.org/show_bug.cgi?id=1900926

This does the following:
  * ~~when a class has no super class and has virtual method, add a pseudo field for vtable, for all platform~~
  * ~~the vtable size is currently hardcoded either 4 or 8, depending on the platform name.  we could make the clang plugin to generate the vtable data or just a pointer size, in followup bug~~
  * when a class has virtual function table pointer, add a pseudo field
  * the vtable size is extracted from the analysis file

`nsISupports` for `nsGlobalWindowInner`:
vtable


arai-a:
actually, the condition used in the patch is wrong.
vtable pointer can be part of subclass if the super class doesn't have virtual.
I think it's better just integrating this with clang plugin to generate the data.

arai-a:
Fixed the patch to use the data generated by clang plugin from https://bugzilla.mozilla.org/show_bug.cgi?id=1901018

asutherland (approved):
Nice!  This is super interesting!

PR #751, opened 2024-06-06
Bug 1901146 - Add space between tables in field layout view for each class hierarchy

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1901146

This adds spacing (margin) before 2nd+ tables for field layout view.

table-space

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -1315,6 +1315,10 @@ g.true-diag g.node polygon {
   border-collapse: separate;
 }
 
+.symbol-tree-table + .symbol-tree-table {
+  margin-top: 2em;
+}
+
 .symbol-tree-table h3 {
   margin: 0;
 }


asutherland (approved):
Makes sense!  Thanks!

asutherland:
This is an interesting case that brings up a few questions (for other bugs):
- `js::proxy` seems to [be a function](https://searchfox.org/mozilla-central/rev/d3fea1aa852bb3a353a0a4a875c3685da11ce39b/js/src/proxy/ScriptedProxyHandler.cpp#1545) so maybe we should be filtering to just classes/structs as part of the symbol resolution stage.  We do apply [meta.kind checks in the context-menu](https://github.com/mozsearch/mozsearch/blob/c2a2e4a6476a4fd278953876072631eac3cc23e4/static/js/context-menu.js#L384), but they of course are not propagated to the query mapping.  [Here is where we invoke cmd_crossref_lookup](https://github.com/mozsearch/mozsearch/blob/master/tools/src/query/query_core.toml#L187-L189) but [we don't have an argument that would do the meta.kind filtering yet](https://github.com/mozsearch/mozsearch/blob/master/tools/src/query/query_core.toml#L187-L189).  Presumably if we applied that, we would want to treat lack of a meta.kind as a failure to match since we definitely can't do anything useful for things without meta.
- I wonder if we should have a heuristic that makes a situation like this become a case-sensitive match?  There have been some bugs filed about potentially treating any search with mixed-case as case-sensitive since that presumes intent, but it creates a interesting situation for `js::proxy` which then is ambiguous but `js::Proxy` is not.

arai-a:
Yes, I agree that non-classes/non-structs should be filtered out.

Also, treating this query as case-sensitive makes sense, as long as the primary entry point for this feature is the context menu, and the query is not much supposed to be hand-written.

asutherland:
> Also, treating this query as case-sensitive makes sense, as long as the primary entry point for this feature is the context menu, and the query is not much supposed to be hand-written.

I think that's definitely where we are right now.  There's some discussion on https://github.com/staktrace/query-parser/issues/1 that's interesting.  I think there are a lot of benefits to the query mechanism having a well-defined grammar, but I'm not sure it's sufficiently ergonomic.  If we are able to migrate to using the pipeline-server for everything, I think we may effectively need to have an "initial search grammar" that's more heuristic based and used for initial raw input like from keyword searches or the search field at the top of the screen, and then when it hits the server we could re-write it into the "normalized search grammar" that the query server fully operates in.  We could potentially have some kind of autocomplete UI or other client-side mechanism that would help ensure the queries are normalized search grammar, although that might be something people don't want / would be too un-searchfoxy.

arai-a:
yeah, making the search field also compatible with queries would be nice (actually, I first tried putting the query copied from different tab in the search field and it of course didn't work), but indeed there are conflicting cases and the grammar needs to be something special.

PR #752, opened 2024-06-07
Bug 1901279 - Show only classes/structs in field layout view

arai-a:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1901279

This does the following:
  * While traversing, filter out structured record that doesn't correspond to struct-ish data (https://clang.llvm.org/doxygen/namespaceclang.html#a9237bdb3cf715b9bff8bcb3172635548 suggests there can be some more than struct/class)
  * Do not add a table if the table is empty (the table is non-empty even if the struct has no field, because the table has the class name row)


asutherland (approved):
Ah, yeah, doing the filtering in field-layout also works (as opposed to in the crossref lookup pipeline stage).

PR #754, opened 2024-06-09
Bug 1416899 - Index WebIDL files

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1416899
depends on https://github.com/mozsearch/mozsearch-mozilla/pull/245
also depends on https://bugzilla.mozilla.org/show_bug.cgi?id=1901445

This does the following:
  * Collect WebIDL files, based on the filter from the config
  * Let `idl-analyze.sh` copy/download `WebIDL.py` in addition to `xpidl.py`
  * Generate analysis files for each WebIDL file, in the same way as XPIDL
  * Add slots to associate with C++ code
    * IDL interface => C++ namespace
      * e.g. `Performance` => `mozilla::dom::Performance_Binding`
    * IDL namespace => C++ namespace
      * e.g. `ChromeUtils` => `mozilla::dom::ChromeUtils_Binding`
    * IDL methods => C++ function
      * e.g. `ChromeUtils.getObjectNodeId` => `mozilla::dom::ChromeUtils_Binding::getObjectNodeId`
    * IDL attributes => C++ getter/setter functions
      * e.g. `Performance.timing` => `mozilla::dom::Performance_Binding::get_timing`
    * IDL const => C++ constant
      * e.g. `PerformanceNavigation.TYPE_NAVIGATE` => `mozilla::dom::PerformanceNavigation_Binding::TYPE_NAVIGATE`
    * IDL maplike/setlike/iterable/async iterable => C++ namespace for helpers
      * e.g. `HighlightRegistry.maplike` => `mozilla::dom::HighlightRegistry_Binding::MaplikeHelpers`
      * e.g. `ReadableStream::asynciterable` => `mozilla::dom::ReadableStreamAsyncIterator_Binding`
    * IDL dictionary => C++ class
      * e.g. `ImportESModuleOptionsDictionary` => `mozilla::dom::ImportESModuleOptionsDictionary`
    * IDL dictionary field => C++ class member
      * e.g. `ImportESModuleOptionsDictionary.global` => `mozilla::dom::ImportESModuleOptionsDictionary::mGlobal`
    * IDL enum => C++ enum class
      * e.g. `WebIDLProcType` => `mozilla::dom::WebIDLProcType`
  * Add slots to associate with JS symbols

TODO:
  * ~~remove the workaround once https://bugzilla.mozilla.org/show_bug.cgi?id=1901445 fix lands~~
  * ~~add testcases~~
  * ~~test on dev channel~~

Now available on dev channel, with importing `WebIDL.py` fix: https://dev.searchfox.org/mozilla-central/search?q=&path=.webidl&case=false&regexp=false

### DIFF
--- a/scripts/webidl-analyze.py
+++ b/scripts/webidl-analyze.py
@@ -411,9 +411,8 @@ def handle_maplike_or_setlike_or_iterable(records, iface_name, target):
     elif target.maplikeOrSetlikeOrIterableType == 'asynciterable':
         cpp_sym = f'NS_mozilla::dom::{iface_name}AsyncIterator_Binding'
     else:
-        print(f'WebIDL: Unknown maplikeOrSetlikeOrIterableType: {target.maplikeOrSetlikeOrIterableType}',
+        print(f'warning: WebIDL: Unknown maplikeOrSetlikeOrIterableType: {target.maplikeOrSetlikeOrIterableType}',
               file=sys.stderr)
-        sys.exit(1)
 
     emit_source(records, loc, 'idl', 'method', pretty, idl_sym)
     emit_target(records, loc, 'idl', pretty, idl_sym)
@@ -482,9 +481,8 @@ def handle_interface_or_namespace(records, target):
         elif isinstance(member, WebIDL.IDLAsyncIterable):
             handle_maplike_or_setlike_or_iterable(records, name, member)
         else:
-            print(f'WebIDL: Unknown member production: {member.__class__.__name__}',
+            print(f'warning: WebIDL: Unknown member production: {member.__class__.__name__}',
                   file=sys.stderr)
-            sys.exit(1)
 
     emit_structured(records, loc, 'class', pretty, idl_sym,
                     slots=slots, supers=supers,
@@ -548,9 +546,8 @@ def handle_dictionary(records, target):
             handle_dictionary_field(records, name, cpp_sym,
                                     fields, member)
         else:
-            print(f'WebIDL: Unknown member production: {member.__class__.__name__}',
+            print(f'warning: WebIDL: Unknown member production: {member.__class__.__name__}',
                   file=sys.stderr)
-            sys.exit(1)
 
     emit_structured(records, loc, 'class', pretty, idl_sym,
                     slots=slots, supers=supers,
@@ -713,9 +710,8 @@ def handle_productions(productions):
             records = get_records(target)
             handle_external_interface(records, target);
         else:
-            print(f'WebIDL: Unknown top-level production: {target.__class__.__name__}',
+            print(f'warning: WebIDL: Unknown top-level production: {target.__class__.__name__}',
                   file=sys.stderr)
-            sys.exit(1)
 
 
 def write_files(analysis_root):
--- a/scripts/webidl-analyze.py
+++ b/scripts/webidl-analyze.py
@@ -408,10 +408,6 @@ def handle_maplike_or_setlike_or_iterable(records, iface_name, target):
     elif target.maplikeOrSetlikeOrIterableType == 'setlike':
         cpp_sym = f'NS_mozilla::dom::{iface_name}_Binding::SetlikeHelpers'
     elif target.maplikeOrSetlikeOrIterableType == 'iterable':
-        # NOTE: loc is wrong. it points '<' after 'iterable'.
-        # TODO: Fix the WebIDL parser.
-        location = target.identifier.location
-        loc = to_loc_with(location._lineno, location._colno - len(name), len(name))
         cpp_sym = f'NS_mozilla::dom::{iface_name}Iterator_Binding'
     elif target.maplikeOrSetlikeOrIterableType == 'asynciterable':
         cpp_sym = f'NS_mozilla::dom::{iface_name}AsyncIterator_Binding'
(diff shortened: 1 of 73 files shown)


scripts/webidl-analyze.py
```diff
@@ -0,0 +1,749 @@
...
+
+    fields = []
+    for member in target.members:
+        if isinstance(member, WebIDL.IDLArgument):
+            handle_dictionary_field(records, name, cpp_sym,
+                                    fields, member)
+        else:
+            print(f'WebIDL: Unknown member production: {member.__class__.__name__}',
+                  file=sys.stderr)
+            sys.exit(1)
```
asutherland, line 558: Meta: Since this is working, should this and the other similar sys.exit() calls just be skipping the top-level thing going on and give a warning that our warning emailing script will pick up instead of exiting from the batch job?

No need to change anything in the PR.

asutherland (approved):
Wooo! This is so awesome!

And it seems like the work on https://bugzilla.mozilla.org/show_bug.cgi?id=1895418 could make it feasible to have a high quality linkage to this on the JS side so that searchfox diagrams/etc. could make the transition between JS uses of WebIDL and the underlying C++ implementations!

asutherland (approved):
Really nice test coverage with a nice trade-off on the C++ symbols.

arai-a:
Method: binding and impl
method

Method with overload: binding and 2 impls
(one of the impl corresponds to the other overload, but so far both are shown there, given it's hard to distinguish at the point of webidl analysis)
method-overload

Readonly attribute: getter binding and getter impl
attr-getter

Attribute: getter binding, getter impl, setter binding, and setter impl
attr-getter-setter

tools/src/file_format/analysis.rs
```diff
@@ -316,6 +316,7 @@ pub enum BindingSlotKind {
     InterfaceName,
     /// Callable.
     Method,
+    MethodImpl,
```
asutherland, line 319: Can you elaborate on the rationale for creating `Impl` suffixed variants of these types instead of doing something like [adding an implKind property to the slots](https://matrix.to/#/!vDKxYNxlsZYvjSWGBh:mozilla.org/$BFVpunP0N71qh_ChvkD3qFW-wcwvLj_YORjjg1eN-00?via=mozilla.org&via=matrix.org&via=guichard.eu)?

I'm concerned that we potentially end up needing Impl variants for pretty much everything in this list.  (It does get a bit interesting where for some bindings we're talking about a sub-section of an existing function, which is not a concept searchfox currently has, but I think is something we're likely to want longer term.)  And then match arms get a bit ugly, although that isn't experienced in the JS code here because it's using `endsWith`.

If we stick with this, I think it would be good to have explicit comments that identify the difference between the types.  I often realize I probably need to change how I'm modeling something when I go to write up the comments and I find out it's difficult to explain.  I think this case is probably along those lines, as I think our existing binding slot generators are probably all largely already emitting what would be Impl-suffixed here.  The existence of `formatBindingSlotKind` having first figured out if there are 2 paired slots allows us to compensate for this in the UI.

Of course, this isn't something `implKind` on its own fixes.  If it were an `Option` over an enum, that lets the analysis express that it doesn't care, but we do potentially still have to pair things up.  That's why the other [option I was proposing in chat was to put multiple symbols in the same slot](https://matrix.to/#/!vDKxYNxlsZYvjSWGBh:mozilla.org/$pWbNHvLazoNl_gDzeQfOPF7oBFAaHhi8pJoGviQVxX4?via=mozilla.org&via=matrix.org&via=guichard.eu).  That ends up making the concept of binding slots a bit more complicated to deal with for the simple cases, but avoids the need to take extra-steps to perform the aggregation.  To elaborate, that might look like each binding slot having `syms: [{ sym: "SYM", kind: "binding"}, { sym: "OTHERSYM", kind: "wrapper" }]` instead of a `sym`.

arai-a: Oh, I completely forgot that.
Will fix shortly with optional `implKind`

The `syms` would need some more consideration related to overloads, and I'll leave it to other bug, which possibly need to take other language (Java etc) into account.

asutherland (approved):
The revised approach looks great and I appreciate the comments, thanks!

asutherland:
Oh wow, and I had missed appreciating the screenshot of "Attribute: getter binding, getter impl, setter binding, and setter impl" as a great worst-case situation!  Truly an excellent test scenario for future potential context-menu overhauls!  Thanks for taking the time to find and generate all of those!

asutherland:
Seems like https://bugzilla.mozilla.org/show_bug.cgi?id=1895418 landing https://searchfox.org/mozilla-central/source/tools/ts/config/builtin.webidl snuck in and caused breakage:

```
+ /home/ubuntu/mozsearch/scripts/webidl-analyze.py /mnt/index-scratch/mozilla-central /mnt/index-scratch/mozilla-central/git /mnt/index-scratch/mozilla-central/analysis /tmp
WebIDL: Syntax error in IDL /mnt/index-scratch/mozilla-central/git/tools/ts/config/builtin.webidl
Traceback (most recent call last):
  File "/home/ubuntu/mozsearch/scripts/webidl-analyze.py", line 899, in <module>
    productions = parse_files(index_root, files_root, cache_dir)
  File "/home/ubuntu/mozsearch/scripts/webidl-analyze.py", line 836, in parse_files
    raise e
  File "/home/ubuntu/mozsearch/scripts/webidl-analyze.py", line 833, in parse_files
    parser.parse(text, local_path)
  File "/tmp/pymodules-mozilla-central/WebIDL.py", line 9218, in parse
    self._productions.extend(self.parser.parse(lexer=self.lexer, tracking=True))
  File "/tmp/pymodules-mozilla-central/ply/yacc.py", line 329, in parse
    return self.parseopt(input, lexer, debug, tracking, tokenfunc)
  File "/tmp/pymodules-mozilla-central/ply/yacc.py", line 821, in parseopt
    p.callable(pslice)
  File "/tmp/pymodules-mozilla-central/WebIDL.py", line 7495, in p_InterfaceRest
    p[0] = self.handleNonPartialObject(
  File "/tmp/pymodules-mozilla-central/WebIDL.py", line 7467, in handleNonPartialObject
    raise ex
  File "/tmp/pymodules-mozilla-central/WebIDL.py", line 7458, in handleNonPartialObject
    raise WebIDLError(
WebIDL.WebIDLError: error: Interface has the same name as non-interface object, tools/ts/config/builtin.webidl line 5:46
[LegacyNoInterfaceObject,Exposed=*] interface BufferSource {};
                                              ^
<unknown> line 2:8
        typedef (ArrayBufferView or ArrayBuffer) BufferSource;
        ^
```

asutherland:
Maybe we can just add a path-filter for `tools/ts/` for now.  I need to re-run config1 and config4 because of my cypress patch, so I'll do that quickly.  (Also, I think my cypress patch probably needs me to propagate the webidl filter too maybe?)

asutherland:
Okay, I did that in https://github.com/mozsearch/mozsearch-mozilla/pull/253 and am re-triggering config1 and then will re-trigger config4 once it seems like config1 has updated the m-c blame tarball.

PR #755, opened 2024-06-10
Bug 1901189 - Support Kotlin style triple double quote literals in syntax highlighter

makotokato:
Kotlin supports triple [double quote literals](https://kotlinlang.org/docs/strings.html#multiline-strings),  but it doesn't support single quote like Python.

### DIFF
--- a/tools/src/languages.rs
+++ b/tools/src/languages.rs
@@ -10,11 +10,11 @@ pub struct LanguageSpec {
     pub c_style_comments: bool,
     pub backtick_strings: bool,
     pub regexp_literals: bool,
-    pub triple_quote_literals: bool,
     pub c_preprocessor: bool,
     // Rust is mostly C-like, with a couple of differences.
     pub rust_tweaks: bool,
     pub cxx14_digit_separators: bool,
+    pub triple_quote_literals: &'static str,
     pub markdown_slug: &'static str,
 }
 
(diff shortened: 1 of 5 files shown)


tools/src/languages.rs
```diff
@@ -10,11 +10,11 @@ pub struct LanguageSpec {
     pub c_style_comments: bool,
     pub backtick_strings: bool,
     pub regexp_literals: bool,
-    pub triple_quote_literals: bool,
     pub c_preprocessor: bool,
     // Rust is mostly C-like, with a couple of differences.
     pub rust_tweaks: bool,
     pub cxx14_digit_separators: bool,
+    pub triple_quote_literals: &'static str,
```
asutherland, line 17: Is there a reason you went with a static str instead of creating an enum type?  No need to change this[1], more of an idle curiosity.

1: In particular, when I did the WebIDL review, I was struck by how the github PR diff had basic definition/use magic which I presume is based on their tree-sitter stack-graph technology they use for general code search/indexing.  I hadn't realized that existed, and it's a very neat feature, and I think probably further confirms that we do want to use tree-sitter tokenizers (which we'll effectively already have to be using anyways for hyperblame to reconstitute the blame-per-line).

asutherland (approved):

tools/src/languages.rs
makotokato: No reason to use str. I can/will update this using enum.

PR #756, opened 2024-06-12
Bug 1896336 - Handle UsingShadowDecl in visitHeuristicResult

theres-waldo:


asutherland (approved):
Thanks for the minimization and all the analysis in the bug!  Combined with that and https://clang.llvm.org/doxygen/classclang_1_1UsingShadowDecl.html I think I understand what's going on here!

theres-waldo:
I deployed this patch to botond.searchfox.org and confirmed that it fixes the original m-c testcase reported in the bug.

PR #757, opened 2024-06-12
[WIP] Bug 1781179 - Improve C++ syntax highlighting

theres-waldo:


theres-waldo:
This is a very early WIP patch for [bug 1781179](https://bugzilla.mozilla.org/show_bug.cgi?id=1781179), which intends to annotate C++ source tokens with a richer set of semantic token kinds for highlighting purposes.

So far, only a subset of the semantic token kinds are computed (those which I could easily infer from the `kind` field of `AnalysisStructured`). The idea is to expand these over time by looking at other fields of `AnalysisStructured` and/or `AnalysisSource` (and if needed putting more information into those structs).

Currently, what I'm seeing during local testing is that even this subset is only being populated for some symbols -- for many symbols, there is no `meta` field present in the jumpref record. I'm not sure why that is; I will continue to investigate, but I wanted to post what I had in case I was doing something obviously wrong.

asutherland:
`meta` come from structured records which we only emit when we see defs (or decls we decide to treat as defs).  If you aren't seeing a meta for something but you see a structured, there's something wrong.

One potentially notable thing since you spend a lot of time around templates, I believe I put [this bailout](https://searchfox.org/mozilla-central/rev/346ac9335cbc33e3124a4541c938a0dc455e785e/build/clang-plugin/mozsearch-plugin/MozsearchIndexer.cpp#1924-1931) in because clang straight up crashes if you try and get the AST record layout of something that doesn't have an AST record layout.  We definitely could emit structured records in those cases and just not claim to know the layout of stuff:
```cpp
if (D2->isThisDeclarationADefinition() &&
    // XXX getASTRecordLayout doesn't work for dependent types, so we
    // avoid calling into emitStructuredInfo for now if there's a
    // dependent type or if we're in any kind of template context.  This
    // should be re-evaluated once this is working for normal classes and
    // we can better evaluate what is useful.
    !D2->isDependentType() &&
    !TemplateStack) {
```

If there's specific locations you can point to, I can help drill down more.

Note that because the "dev" channel is currently spoken for I've just created a "botond" channel via `channel-tool.py` so you should feel free to trigger indexer runs for "botond" instead of "dev" for the [indexer run steps](https://github.com/mozsearch/mozsearch?tab=readme-ov-file#triggering-an-aws-indexer-run), although of course if you have clang indexer changes, you'll need to [do an m-c try build first](https://github.com/mozsearch/mozsearch?tab=readme-ov-file#running-mozilla-central-try-builds-for-changes-to-in-tree-indexing) and then do the `TRYPUSH_REV` thing the indexer run steps call for.

theres-waldo:
> One potentially notable thing since you spend a lot of time around templates, I believe I put [this bailout](https://searchfox.org/mozilla-central/rev/346ac9335cbc33e3124a4541c938a0dc455e785e/build/clang-plugin/mozsearch-plugin/MozsearchIndexer.cpp#1924-1931) in [...]

Yeah, that seems to be it. I was using bug1781178.cpp as my testcase, but of course that's mostly templates :)

So I think what I'll do as a next step is write some patches (probably in separate PRs) to emit structured records for more types of C++ symbols. In addition to method and class templates, some other categories that have come up so far are namespaces, enums, and namespace-scope variables. Then I'll continue extending this patch to start producing semantic tokens for those categories of symbols.

asutherland:
That sounds great!

I think the namespace case may be an interesting edge-case because in general we expect to only have a single structured record whose loc is the same loc as the "def" for the symbol which we pretty assume there is only one of.  But we do have like thousands of "def" records for the "mozilla" namespace.  Crossref will currently [only record the first structured record it sees for a symbol](https://github.com/mozsearch/mozsearch/blob/fc07f9c6525f10f51c950a249d110d9595236c5d/tools/src/bin/crossref.rs#L176) so as long as the structured records are all identical for the same symbol, that should be fine.

It seems like we [don't currently generate](https://searchfox.org/mozilla-central/rev/a18d79db1695fc6019cac47874113fd3bb65a821/build/clang-plugin/mozsearch-plugin/MozsearchIndexer.cpp#409-411) records for [unnamed (anonymous for gdb) namespaces](https://en.cppreference.com/w/cpp/language/namespace#Unnamed_namespaces) or [inline namespaces](https://en.cppreference.com/w/cpp/language/namespace#Inline_namespaces).  Because those can be opened multiple times in the file I guess we'd mangle the path into the symbol identifier like for `FILE_` and/or do our hashing thing but obviously excluding the line number from the hash.  (Noting that I think the deterministic hashing was probably a performance optimization we're not really as concerned about right now, although it could still matter for URL length reasons I suppose.)

PR #758, opened 2024-06-14
Add symbol info to field types in field layout table

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1900830

This does the following:
  * Refactor the symbol tree table:
    * Add `BasicMarkup` variants for text/heading with `data-symbols`
    * Convert the field name cell into general `SymbolTreeTableCell`
  * Make each field type into span with `data-symbols`, for the field type and `pointerInfo` types
    * given we don't have tokenization for the pretty string, the entire pretty string is marked with `span` with `data-symbols`, for all types appears inside it
  * Add more JS types to ontology mapping, to make them available in field layout table

Single type:
single-type

Wrapped type:
wrapped-type

### DIFF
--- a/config_defaults/ontology-mapping.toml
+++ b/config_defaults/ontology-mapping.toml
@@ -562,10 +562,44 @@ kind = "weak"
 [types."nsRefPtrHashKey".pointer]
 kind = "strong"
 
+# ### JS Wrapper Types ###
+
+[types."JS::Result".value]
+
+# ### JS Container Types ###
+
+[types."js::HashMap".container]
+
+[types."js::HashSet".container]
+
+[types."js::ProtectedData".container]
+
+[types."js::ProtectedDataNoCheckArgs".container]
+
+[types."js::ProtectedDataWriteOnce".container]
+
+[types."js::Vector".container]
+
+# ### JS Pointer Types ###
+
+[types."js::AtomicRefCounted".pointer]
+kind = "strong"
+
+[types."js::RefCounted".pointer]
+kind = "strong"
+
+# ### JS GC Wrapper Types ###
+
+[types."js::GCPtr".pointer]
+kind = "gcref"
+
 [types."JS::Handle".pointer]
 kind = "gcref"
 
 [types."JS::Heap".pointer]
+kind = "gcref"
+
+[types."js::HeapPtr".pointer]
 kind = "gcref"
 
 [types."JS::TenuredHeap".pointer]
@@ -574,8 +608,25 @@ kind = "gcref"
 [types."JS::MutableHandle".pointer]
 kind = "gcref"
 
+[types."JS::PersistentRooted".pointer]
+kind = "gcref"
+
 [types."JS::Rooted".pointer]
 kind = "gcref"
+
+# ### JS GC Collection Types ###
+
+[types."JS::GCHashMap".container]
+
+[types."js::GCRekeyableHashMap".container]
+
+[types."JS::GCVector".container]
+
+[types."JS::PersistentRootedVector".container]
+
+[types."JS::RootedVector".container]
+
+[types."JS::StackGCVector".container]
 
 # #### Webkit
 #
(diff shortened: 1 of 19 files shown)


asutherland:
Is it possible to kick off a dev channel run of this if you have not?

arai-a:
okay, triggered. it will become available in 1-2 hours

arai-a:
Here's dev channel
https://dev.searchfox.org/mozilla-central/query/default?q=field-layout%3A%27js%3A%3Agc%3A%3AGCRuntime%27

arai-a:
Just realized that the field type pretty string uses qualified type name, we can just search each pointer_type pretty string and markup each occurrence (removing the `T_` prefix).
I'll look into it in followup

asutherland:
> Just realized that the field type pretty string uses qualified type name, we can just search each pointer_type pretty string and markup each occurrence (removing the `T_` prefix). I'll look into it in followup

Yeah, to get the hack to work reliably I had to make sure the types were qualified.

Two thoughts:
- We could potentially have [maybe_parse_type_as_pointer](https://github.com/mozsearch/mozsearch/blob/b58a921c69986d85e5e963e5a5068923e8116598/tools/src/file_format/ontology_mapping.rs#L281) emit the span info into the type string provided to it.
- Although I think it's essential for the class diagram mechanism to be able to ignore the intermediary types that don't contribute to understanding (other than via labels that may be turned into icons), I think it could be okay for our parsing of the type to emit spans and symbol info for the types we don't actually care about as long as the diagram logic doesn't get confused.  (Like the diagram could just skip boring subtypes probably.)  Trying to reubild the info separately via searching seems error prone.  Plus, when we have the clang indexer help emit the types, we can ideally have it emit the spans too (although potentially against non-fully-qualified identifiers to maybe make things more terse?).
  - Aside: on trunk the [diagram for NativeIterator is boring](https://searchfox.org/mozilla-central/query/default?q=class-diagram%3A%27js%3A%3ANativeIterator%27%20depth%3A4) but [with the improved ontology mapping it's much more interesting!](https://dev.searchfox.org/mozilla-central/query/default?q=class-diagram%3A%27js%3A%3ANativeIterator%27%20depth%3A4) woo!

hotsphink:
Thank you arai! Looking at the example really drove home two things (that I more or less already knew, but this highlighted them):
1. Templates are a pain in the butt.
2. SpiderMonkey uses a lot of cryptic wrapper types that makes it really hard to understand what's actually going on.

I noticed a couple of things as a result of looking at the output here. My apologies if the resulting bugs are noise.

arai-a:
>     * We could potentially have [maybe_parse_type_as_pointer](https://github.com/mozsearch/mozsearch/blob/b58a921c69986d85e5e963e5a5068923e8116598/tools/src/file_format/ontology_mapping.rs#L281) emit the span info into the type string provided to it.

Sounds good!
I'll look into it.

Other option is to somehow use the actual field definition's type tokens with symbols already specified in the analysis file.
For the template usage in SpiderMonkey, expanding the typedef/using might be too verbose.

e.g. put `MainThreadData<JS::Zone*>` instead of `class js::ProtectedDataNoCheckArgs<class js::CheckMainThread<js::AllowedHelperThread::None>, class JS::Zone *>` in the table

asutherland:
> Other option is to somehow use the actual field definition's type tokens with symbols already specified in the analysis file. For the template usage in SpiderMonkey, expanding the typedef/using might be too verbose.
> 
> e.g. put `MainThreadData<JS::Zone*>` instead of `class js::ProtectedDataNoCheckArgs<class js::CheckMainThread<js::AllowedHelperThread::None>, class JS::Zone *>` in the table

Yes, I think that would definitely be something to aim for if/when updating the clang indexer, as it seems like the thing people would most expect is what they'd read in the source.  (And again, the current expansion was just an evolutionary series of hacks due to time limitations.)  For letting people understand what that expands to, I think what Nicolas had prototyped so far for macro expansion in https://github.com/mozsearch/mozsearch/pull/679 could potentially make that work[1].  I don't know that we took any screenshots unfortunately, but the basic idea was to show the expansion inline and let it overflow, and then use a slightly modified blame popup for the full expansion.

And if there were any particularly unpleasant types that somehow weren't feasible to clean up in m-c itself, we could always support having a "pretty print" variant of them defined.

1: Well, there's always the issue that very long template types are hard to read no matter what.

PR #759, opened 2024-06-17
WIP: Bug 1297090 - Support contextual keyword.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1297090

This does:
  * Modify the class name handling to be based on the class name itself, instead of `class="..."` attribute
  * Add `syn_contextual` which is temporary class for internal use
  * If a token with `syn_contextual` has corresponding analysis record, convert it into symbol for the record
  * If a token with `syn_contextual` has no corresponding analysis record, convert it into `syn_reserved`

This patch is should be rebased onto [bug 1781179](https://bugzilla.mozilla.org/show_bug.cgi?id=1781179), and also needs to be split into smaller patch after that.


arai-a:
I'll go with making keyword always contextual.

PR #760, opened 2024-06-19
Bug 1903464 - Move field-layout table formatting to liquid template

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1903464

This does the following:
  * Refactor `SymbolTreeTable*` structs to directly map the source data
    * `SymbolTreeTableNode` for class
    * `SymbolTreeTableItem` for field, hole, and padding
    * `SymbolTreeTableFieldType` for field's type
    * `SymbolTreeTableFieldOffsetAndSize` for field's offset and size
  * Move all styling and table row/cell handlings to liquid template

This will simplify https://bugzilla.mozilla.org/show_bug.cgi?id=1902875 patch, which is going to add more cell and also more styling


asutherland (approved):
Having the JSON rep be more semantic seems much nicer!  I very much appreciate how the resulting liquid templates feel much more idiomatic since then they're transforming the semantic rep to HTML rather than converting an ad-hoc markup abstraction that didn't have great ergonomics.  Thanks!

PR #762, opened 2024-06-23
Bug 1768480 - Allow more characters to be followed by RegExp in tokenizer

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1768480

This does the following:
  * allow `[` to be followed by RegExp, which can be e.g. `[/foo/]`
  * allow `}` to be followed by RegExp, which can be the first token after the last block statement
  * allow `;` to be followed by RegExp, which can be the first token after last statement
  * add tests
  * add some comments about overlapping cases


asutherland (approved):
Really nice test comments and great test coverage!  Thanks!

PR #763, opened 2024-06-23
Bug 1902875 - Show the field definition's lines in the field layout

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1902875

This does the following:
  * Add server method to fetch list of HTML-formatted lines separately
  * Add field's range as `lineRange` property to `StructuredFieldInfo`
    * this uses `PATH#line` or `PATH#line-line` format
    * `PATH` part is omitted if the field and struct exist in the same file
    * `PATH` is the full path in the repository
      * this could be relative path from the struct's file, but given it doesn't often happen in the current implementation (where the situation can happen only when the field is defined by `#include`), I'll leave it to other bug
      * once we support macro expansion, the situation will happen more, and we'll need to support relative path, in order to shorten the value
  * Rewrite the field sorting algorithm to use the `lineRange`, including:
    * use the start line for sorting
    * do not use line number when fields come from different file
  * Show the field definition lines in the table
    * This is same thing as source listing
  * Hide the field type column by default
  * Add column selector
    * controllable by both the query parameter and the checkbox
    * uses `show-cols` and `hide-cols` query parameter, with comma-separated values, instead of multiple `show-col`/`hide-col` parameters, given the named parameter doesn't support multiple occurrences in the query parser
    * show/hide is controlled by CSS classes, applied to the `#symbol-tree-table-list` div
  * Tweak the table styling:
    * use monospace font for field name and type
    * avoid line-wrapping for offset/size
    * make the border rendering compatible with showing/hiding columns

currently available on dev channel
https://dev.searchfox.org/mozilla-central/query/default?q=field-layout%3A%27mozilla%3A%3Adom%3A%3ABrowsingContext%27

### DIFF
--- a/static/js/search.js
+++ b/static/js/search.js
@@ -32,6 +32,8 @@ var Dxr = new (class Dxr {
       path: document.getElementById("path-bubble"),
     };
 
+    this.setupColSelector();
+
     this.startSearchTimer = null;
     // The timer to move to the next url.
     this.historyTimer = null;
@@ -107,11 +109,8 @@ var Dxr = new (class Dxr {
 
     this.hideBubbles();
 
-    let url = new URL(this.searchUrl, window.location);
-    url.searchParams.set("q", this.fields.query.value);
-    url.searchParams.set("path", this.fields.path.value.trim());
-    url.searchParams.set("case", this.fields.caseSensitive.checked);
-    url.searchParams.set("regexp", this.fields.regexp.checked);
+    let url = this.constructURL();
+
     let controller = new AbortController();
 
     this.fetchController = controller;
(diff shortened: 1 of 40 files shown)


static/js/search.js
```diff
@@ -141,6 +140,26 @@ var Dxr = new (class Dxr {
     }
 
     populateResults(results, false, false);
+
+    this.updateHistory(url);
+  }
+
+  constrcutURL() {
```
asutherland, line 147: nit: there is a transposition typo here `tcut` should be `tuct`

asutherland (approved):
Having the excerpted line there with all of the symbols clickable is really amazing!  And the ability to instantaneously add/remove columns to get more information or simplify the table is really great!

PR #765, opened 2024-06-24
Bug 1904373 - pipeline-server ulimit should be aware of mmap data sizes

asutherland:
Initially we used a hard-coded ulimit of 24G based on observed values on config1 out of a combination of erring on the side of safety and because I worried the shell scripting would be a headache and likely to break.  But it turns out the arithmetic context offered by bash via `$(( $a + $b ))` is easy to use in a way I honestly did not expect.

This patch translates the original rationale for arriving at 24G to be done based on invocations of du.  This works as expected on config4 (where I ran an updated version of the script) as well as config1 (where I ran the shell scripts manually).

### DIFF
--- a/infrastructure/web-server-run.sh
+++ b/infrastructure/web-server-run.sh
@@ -31,24 +31,31 @@ nohup $MOZSEARCH_PATH/router/router.py $CONFIG_FILE $STATUS_FILE > $SERVER_ROOT/
 export RUST_BACKTRACE=1
 nohup $MOZSEARCH_PATH/tools/target/release/web-server $CONFIG_FILE $STATUS_FILE > $SERVER_ROOT/rust-server.log 2> $SERVER_ROOT/rust-server.err < /dev/null &
 
-# let's try and stop the pipeline-server from causing problems by setting a ulimit
-# on virtual memory usage.  Currently, the worst-case scenario is mozilla-central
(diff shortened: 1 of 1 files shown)


infrastructure/web-server-run.sh
```diff
@@ -31,24 +31,31 @@ nohup $MOZSEARCH_PATH/router/router.py $CONFIG_FILE $STATUS_FILE > $SERVER_ROOT/
...
+# When first adding the ulimit, our VM size was 13.7G with resident usage of
+# 390M.  When writing this on the spare config1 I'm seeing 13.5G VM with 668M
+# resident with the MAPPED_FILES_USAGE above reporting ~12.7G which gives 800M
+STEADY_STATE_ASSUMED_K=$((800 * 1024))
+# Allowed growth.  When first adding the ulimit, we allowed 10.3G of VM usage
+# which paired with a 10.7G of resident usage.  I'm going to round this down to
+# 10G since we already grew the steady state above.
+ALLOWED_GROWTH_K=$((10 * 1024 * 1024))
+
+# I've also jmust manually confirmed that this works as expected for config4
```
arai-a, line 51: "just"

arai-a (approved):
Thank you for adding comments about the reasoning behind the limits! it's helpful!

Also it's good to know that bash has such feature :)

PR #766, opened 2024-06-24
Reuse more code between the codepaths that emit structured info for different Decl types

theres-waldo:


theres-waldo:
Technically, there is a small functional change in this patch: for `ObjCIvarDecl`, we used to bail on emitting a structured record at all, but now we emit one but omit the `parentsym` field. This seems harmless to me, but if you'd prefer I can try harder to preserve the old behaviour.

asutherland (approved):
Not having a parentsym for objective C seems fine.  We currently don't seem to use the parentsym for anything, I think, so there isn't actually anything to degrade.  (Right now the hierarchical clustering for the diagrams is just splitting on the pretty identifiers and then we transform back into symbol-space by performing a lookup on the identifiers.)  Conceptually it is nice to have though, especially since we maybe should try and use it.

asutherland:
Note that I am assuming you ran the check tests and this isn't changing anything in our tests right now.  Please let me know if that's not the case.  (We probably should try and get some CI going... given how long our provisioning process takes to run I've been hesitant to do the naive thing for that though for cost/environmental reasons.)

theres-waldo:
> Note that I am assuming you ran the check tests and this isn't changing anything in our tests right now.

I've run `make build-test-repo` and that passes, is that sufficient?

asutherland:
> > Note that I am assuming you ran the check tests and this isn't changing anything in our tests right now.
> 
> I've run `make build-test-repo` and that passes, is that sufficient?

Yes, 100%.

asutherland:
It looks like this got slightly bit-rotted by https://github.com/mozsearch/mozsearch/commit/34f681c15ff2ac7d06a2d14a892e27d6d9fcad3c in https://github.com/mozsearch/mozsearch/pull/763

I'm now seeing:
```
vagrant@ad5b78a6cd2b:/vagrant$ make build-test-repo
make -C clang-plugin
make[1]: Entering directory '/vagrant/clang-plugin'
clang++ -I/usr/lib/llvm-17/include -std=c++17   -fno-exceptions -funwind-tables -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -fPIC -Wall -Wno-strict-aliasing  -c MozsearchIndexer.cpp -o MozsearchIndexer.o
MozsearchIndexer.cpp:1208:40: error: use of undeclared identifier 'Loc'
 1208 |     FileID structFileID = SM.getFileID(Loc);
      |                                        ^
1 error generated.
make[1]: *** [Makefile:12: MozsearchIndexer.o] Error 1
make[1]: Leaving directory '/vagrant/clang-plugin'
make: *** [Makefile:17: build-clang-plugin] Error 2
vagrant@ad5b78a6cd2b:/vagrant$ 
```

PR #767, opened 2024-06-25
Bug 1904474 - Miscellaneous fixes

staktrace:
A few fixes towards bug 1904474. I don't have it fully working yet though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,12 @@ are a hassle.
 Can you already run "docker" on your command line and have it work?  Then you
 can skip this section!
 
+> [!WARNING]
+> If you're running docker on macOS, make sure the Rosetta option is disabled
+> in the Docker Desktop settings, otherwise you may run into segfaults on
+> installing rust components. See https://github.com/rust-lang/rustup/issues/3902
+> and friends.
+
 #### Important Docker Licensing Notes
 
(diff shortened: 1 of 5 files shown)


asutherland (approved):
There's no `&>`?  Noooooo

staktrace:
> There's no `&>`? Noooooo

The new thing is like a superset of `&>`

asutherland:
Sorry, I meant `|&` but 

> > There's no `&>`? Noooooo
> 
> The new thing is like a superset of `&>`

Sorry, I meant `|&` but mainly was assuming the mac didn't have it.

staktrace:
I'm not sure what `|&` does. I just tried it in a bash script and got a syntax error. That being said if it's a bash thing it shouldn't matter if it's linux or mac.

asutherland:
https://www.gnu.org/software/bash/manual/html_node/Pipelines.html says:

> If ‘|&’ is used, command1’s standard error, in addition to its standard output, is connected to command2’s standard input through the pipe; it is shorthand for 2>&1 |. This implicit redirection of the standard error to the standard output is performed after any redirections specified by command1.

asutherland:
Note I'm not saying to change anything, just I always struggle to find actual docs on bash that are linkable and so that was nice to find and so I'm linking it here.

PR #771, opened 2024-06-26
Test result changes experienced on full reprovision.

asutherland:
I just did a full local reprovision which included purging all docker containers and images as part of trying to make sure our Cargo.lock is up-to-date, and I ended up with these test changes which I'm going to merge.

I'm not quite sure how the xpctest attributes weren't up-to-date, but I think the changes make sense and could involve the difficulty of dealing with/juggling large patch-stacks with our snapshot mechanism and our current lack of CI.  When I was landing my hobby stack, I ended up basically doing an `edit` during a rebase for every single patch and re-running all tests to make sure the snapshots were updated at the right point.  I think this would be more painful now that we have Java build-steps that are (almost?) always re-run.

I believe the change in java is because a new release of scip-java has been made (https://bugzilla.mozilla.org/show_bug.cgi?id=1904166 is where :ryanvm will land it in m-c) and our [install steps don't mention a version](https://github.com/mozsearch/mozsearch/blob/c1afad3fac45665f803dc0c4fa551c635da96f6c/infrastructure/indexer-provision.sh#L56-L57) so the docker build process won't re-run that step unless we touch the file.  Ideally we would encode the version into the provisioning file to address that, although I also think we probably don't all regularly re-provision our docker images unless we need to (which I think is generally fine).

I will kick off a re-provisioning of the AWS images after this so they also end up on the same version of scip-java as us since https://phabricator.services.mozilla.com/D214660 is now approved and likely to land soon.

### DIFF
--- a/tests/tests/checks/snapshots/crossref/crossref/xpctest_attributes.idl/check_glob@booleanProperty_getter__json.snap
+++ b/tests/tests/checks/snapshots/crossref/crossref/xpctest_attributes.idl/check_glob@booleanProperty_getter__json.snap
@@ -55,6 +55,7 @@ expression: "&to_value(scil).unwrap()"
           "slotOwner": {
             "slotKind": "getter",
             "slotLang": "cpp",
+            "implKind": null,
             "ownerLang": "idl",
             "sym": "XPIDL_nsIXPCTestObjectReadWrite_booleanProperty"
           },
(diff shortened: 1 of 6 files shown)


asutherland:
The provisioning processes both succeeded per the emails to the list.

PR #772, opened 2024-06-26
Bug 1904890 - Make build-codesearch.py resilient to previous failures.

asutherland:
If /tmp/dummy already exists, the script fails otherwise.  For hygiene / simplicity reasons I add a call to `shutil.rmtree` to clear out the dir before recreating it anew.  While we do include the tree name in all the paths we generate (except /tmp/livegrep.json), the invocations of `ln -s` will error if the link file already exist, so it is necessary to remove those.

Idiomatically, we use the `run` command to remove the tree at the end of the file, so that also could have made sense, but since we were using `os.mkdir` in the same location to make the directory, I went with using the relevant python command.

### DIFF
--- a/scripts/build-codesearch.py
+++ b/scripts/build-codesearch.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import
 import os
 import sys
 import json
+import shutil
 
 from lib import run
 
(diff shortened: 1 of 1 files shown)


arai-a (approved):
Looks good!

asutherland:
Whoops, apparently `shutil.rmtree` gets sad if the dir doesn't exist.  Changing it to `rm -rf` in a quick fixup.  I'll retrigger config1 after landing.

PR #774, opened 2024-06-26
Bug 1904474 - Drop usage of six in favor of python3 native

staktrace:
Now that these scripts are fully on python3 there's no need to use the migration shims from the `six` package

### DIFF
--- a/router/router.py
+++ b/router/router.py
@@ -1,11 +1,9 @@
 #!/usr/bin/env python3
 
 from __future__ import absolute_import
-import six.moves.SimpleHTTPServer
-from six.moves.BaseHTTPServer import HTTPServer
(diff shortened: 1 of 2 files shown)


staktrace:
Note that I'm not taking it out of the provisioner stuff because it looks like cinnabar might still need python2.7 and six (not sure if that's still true - I'm running without cinnabar so I haven't tested that codepath)

asutherland (approved):
Thanks!

asutherland:
> Note that I'm not taking it out of the provisioner stuff because it looks like cinnabar might still need python2.7 and six (not sure if that's still true - I'm running without cinnabar so I haven't tested that codepath)

I don't think cinnabar needs it anymore but that definitely can be addressed separately.

PR #775, opened 2024-06-26
Bug 1904890 - Whoops, shutil.rmtree complains, just use rm -rf.

asutherland:
So I should have passed `ignore_errors=True` in the prior patch but I think, on balance, just doing the shell thing that most contributors will have awareness of is probably best after all.

### DIFF
--- a/scripts/build-codesearch.py
+++ b/scripts/build-codesearch.py
@@ -4,7 +4,6 @@ from __future__ import absolute_import
 import os
 import sys
 import json
-import shutil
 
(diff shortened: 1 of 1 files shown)


asutherland:
This corrects https://github.com/mozsearch/mozsearch/pull/772 and I copied and pasted the rm -rf line from the bottom of the file so ideally I can't have messed it up :sweat_smile:

asutherland:
re-triggering config1 now

PR #777, opened 2024-06-28
Bug 1880895 - Add webtest command and some JS-based tests for UI

arai-a:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1880895

This does the following:
  * Add `webtest` command for `searchfox-tool` for searchfox UI
     * The communication between browser is done by WebDriver, with https://crates.io/crates/fantoccini
     * Tests and test most of the harness exists in the web-content side, with the mochitest-browser style
     * WebDriver is used for initiating tests and logging
     * The same test can be done from browser, by the following:
       * Load `http://localhost:16995/tests/webtest/webtest.html`
       * Run `TestHarness.loadTest("tests/webtest/test_Search.js")` in console
  * Add a wrapper script to setup the environment and run webtest
    * installs geckodriver
    * downloads firefox
    * starts geckodriver
    * run webtest
    * terminates geckodriver
  * Added some simple tests for search and field layout

TODO:
  * ~~Add substring-match filter for tests~~
  * ~~Somehow hide test frameworks from production~~
  * Maybe import existing utilities from m-c, for event handling etc
  * ~~Add more helpers to test with multiple gate settings~~
  * ~~Add make command, with dependency to test repository~~
  * hide webtest command from web query interface (the `CommandSafetyLevel` in https://github.com/mozsearch/mozsearch/compare/master...asutherland:mozsearch:webdriver-tests?expand=1 I think?). not sure if it's actually exposed there, or maybe not listing it in `query_core.toml` is okay?
  * ~~Add a field to show the iframe's current location, to make it easier to investigate the test progress when manually running test on browser~~
  * ~~maybe rename `web-test` (`WebTest`) to `webtest` (`Webtest`)?~~

Command and example output (done inside docker image)
```
$ make webtest
...
Starting geckodriver
Running tests
INFO - Navigate to http://localhost/static/tests/webtest.html
INFO - Loading static/tests/test_Search.js
TEST_START - static/tests/test_Search.js
INFO - Entering test test_Search
INFO - Waiting for load
INFO - Loading /
INFO - Observed load event
INFO - Observed pageshow event
INFO - Setting text SimpleSearch
INFO - Waiting for condition: 1 class matches
PASS - 1 class matches
INFO - Waiting for condition: URL is updated
PASS - URL is updated
INFO - Leaving test test_Search
...

Overall Summary
===============
Ran 3 tests and 6 subtests in 2.166s.
Passed: 3 tests
Failed: 0 tests

OK
Void result.
Stopping geckodriver: PID=3119359
```


arai-a:
I assume the `webtest` command is not exposed to the web query UI, but I'm not 100% sure.
Can you help me verifying it?
At least it's not listed in `query_core.toml` file.

asutherland:
> I assume the `webtest` command is not exposed to the web query UI, but I'm not 100% sure. Can you help me verifying it? At least it's not listed in `query_core.toml` file.

It doesn't look like your stack has the `CommandSafetyLevel` enum mechanism I added in [the WIP I referenced in the bug](https://github.com/mozsearch/mozsearch/compare/master...asutherland:mozsearch:webdriver-tests?expand=1), but that was intended more as a defense-in-depth measure and assumed webtest would be more dangerous than it is.  If you'd like to include something like that, I think that would be great.  It's a fairly minor and straightforward thing to add; feel free to copy-and-paste parts of if you want to add it in your stack, but it's not necessary.

But as you're inferring, it is the case that a mapping is required in `query_core.toml` to expose it to the web, so that already does provide a level of safety.  The query parsing mechanism is structurally designed to avoid creating a situation where a bad actor can create an arbitrary pipeline, but it felt appropriate to have an extra layer of explicit protection for commands that should never be available from the web.  Also, I should say that the use of clap does have some inherent level of potential concern about escaping/etc. for the arguments, although I believe I addressed that during the time when I addressed the memory backstop, etc.

asutherland (approved):
This is super great!  The introduction of the frame mechanism and shortening the timeouts I think provides all the functionality we need or want from tests at this time, and the familiar mochitest (browser-ish) style I think is extremely ergonomic and readable!

So I think it makes sense to land this as-is and there's no need for me to experiment with the snapshot mechanism at this time.  In the event we run into a situation where we think it could be useful, the way this is structured I think makes it quite feasible for us to add that support in the future, but if I'm going to put time into testing in the short term it's probably more useful for me to hack on making the existing test log/explanations help people understand what's going on with the backend.

r=asuth if you want to add that CommandSafetyLevel too.  r=asuth also if you want to add a section after https://github.com/mozsearch/mozsearch?tab=readme-ov-file#instant-fun-with-the-test-repo that mentions how to run the UI tests.

PR #778, opened 2024-06-28
Bug 1905328 - Robustify against symbol parsing errors

staktrace:
I got an error like:

```
thread 'main' panicked at src/bin/scip-indexer.rs:1123:81:
called `Result::unwrap()` on an `Err` value: InvalidLocalSymbol("local izedMessage.")
```

I need to investigate the root cause (seems like bad scip data) but in the meantime it would be great if this didn't completely abort the indexer.


asutherland (approved):
That's a little funny.

PR #779, opened 2024-06-29
Emit structured info for EnumDecl

theres-waldo:
This is a first attempt at emitting structured information for a new type of C++ entity, enumerations.

It's currently causing the below difference in the output of an existing test (as well as a similar one in the same directory), which I haven't accepted yet because I'm not sure yet if everything in this change is ok.

The appearance of the new `"meta"` field is expected (that's the point of the patch), but the following are open questions in my mind:

 1. Why is the value of the `"pretty"` key next to `"meta"` changing from `"T_field_layout::field_type::Enum1"` to `"field_layout::field_type::Enum1"`?
 2. In terms of the contents of the structured record, the patch is currently only filling out the `"kind"`, because that's the only thing needed by the semantic highlighting code that I'm using this for. Is that all right for now, or will other things get confused by the other fields in the record being null or empty?

```diff
$ diff -u8 tests/tests/checks/snapshots/fancy/format-symbol/field-layout/field-type.cpp/check_glob@field_layout__field_type__json.snap tests/tests/checks/snapshots/fancy/format-symbol/field-layout/field-type.cpp/check_glob@field_layout__field_type__json.snap.new
--- tests/tests/checks/snapshots/fancy/format-symbol/field-layout/field-type.cpp/check_glob@field_layout__field_type__json.snap 2024-06-28 02:10:30.059569389 -0400
+++ tests/tests/checks/snapshots/fancy/format-symbol/field-layout/field-type.cpp/check_glob@field_layout__field_type__json.snap.new     2024-06-28 19:52:25.530934810 -0400
@@ -1,10 +1,11 @@
 ---
 source: tests/test_check_insta.rs
+assertion_line: 143
 expression: "&to_value(sttl).unwrap()"
 ---
 {
   "tables": [
     {
       "jumprefs": {
         "F_<T_field_layout::field_type::S>_hash_map_field": {
           "sym": "F_<T_field_layout::field_type::S>_hash_map_field",
@@ -536,17 +537,36 @@
           "sym": "T_field_layout::field_type::Container2",
           "pretty": "T_field_layout::field_type::Container2",
           "jumps": {
             "def": "field-layout/field-type.cpp#25"
           }
         },
         "T_field_layout::field_type::Enum1": {
           "sym": "T_field_layout::field_type::Enum1",
-          "pretty": "T_field_layout::field_type::Enum1",
+          "pretty": "field_layout::field_type::Enum1",
+          "meta": {
+            "structured": 1,
+            "pretty": "field_layout::field_type::Enum1",
+            "sym": "T_field_layout::field_type::Enum1",
+            "type_pretty": null,
+            "kind": "enum",
+            "subsystem": null,
+            "implKind": "",
+            "sizeBytes": null,
+            "ownVFPtrBytes": null,
+            "bindingSlots": [],
+            "ontologySlots": [],
+            "supers": [],
+            "methods": [],
+            "fields": [],
+            "overrides": [],
+            "props": [],
+            "variants": []
+          },
           "jumps": {
             "def": "field-layout/field-type.cpp#19"
           }
         },
         "T_field_layout::field_type::S": {
           "sym": "T_field_layout::field_type::S",
           "pretty": "field_layout::field_type::S",
           "meta": {
```

### DIFF
--- a/clang-plugin/MozsearchIndexer.cpp
+++ b/clang-plugin/MozsearchIndexer.cpp
@@ -1271,6 +1271,10 @@ public:
     J.arrayEnd();
     J.attributeEnd();
   }
+
+  void emitStructuredEnumInfo(llvm::json::OStream &J, const EnumDecl *ED) {
+    J.attribute("kind", "enum");
+  }
   
   void emitStructuredFunctionInfo(llvm::json::OStream &J, const FunctionDecl *decl) {
     emitBindingAttributes(J, *decl);
@@ -1426,6 +1430,8 @@ public:
 
     if (const RecordDecl *RD = dyn_cast<RecordDecl>(decl)) {
       emitStructuredRecordInfo(J, Loc, RD);
+    } else if (const EnumDecl *ED = dyn_cast<EnumDecl>(decl)) {
+      emitStructuredEnumInfo(J, ED);
     } else if (const FunctionDecl *FD = dyn_cast<FunctionDecl>(decl)) {
       emitStructuredFunctionInfo(J, FD);
     } else if (const FieldDecl *FD = dyn_cast<FieldDecl>(decl)) {
@@ -1916,6 +1922,12 @@ public:
           findBindingToJavaClass(*AstContext, *D3);
           findBoundAsJavaClasses(*AstContext, *D3);
         }
+        emitStructuredInfo(Loc, D2);
+      }
+    }
+    if (EnumDecl *D2 = dyn_cast<EnumDecl>(D)) {
+      if (D2->isThisDeclarationADefinition() && !D2->isDependentType() &&
+          !TemplateStack) {
         emitStructuredInfo(Loc, D2);
       }
     }
--- a/tests/tests/checks/snapshots/fancy/format-symbol/field-layout/field-type.cpp/check_glob@field_layout__field_type__json.snap
+++ b/tests/tests/checks/snapshots/fancy/format-symbol/field-layout/field-type.cpp/check_glob@field_layout__field_type__json.snap
@@ -541,7 +541,26 @@ expression: "&to_value(sttl).unwrap()"
         },
         "T_field_layout::field_type::Enum1": {
           "sym": "T_field_layout::field_type::Enum1",
-          "pretty": "T_field_layout::field_type::Enum1",
+          "pretty": "field_layout::field_type::Enum1",
+          "meta": {
+            "structured": 1,
+            "pretty": "field_layout::field_type::Enum1",
+            "sym": "T_field_layout::field_type::Enum1",
+            "type_pretty": null,
+            "kind": "enum",
+            "subsystem": null,
+            "implKind": "",
+            "sizeBytes": null,
+            "ownVFPtrBytes": null,
+            "bindingSlots": [],
+            "ontologySlots": [],
+            "supers": [],
+            "methods": [],
+            "fields": [],
+            "overrides": [],
+            "props": [],
+            "variants": []
+          },
           "jumps": {
             "def": "field-layout/field-type.cpp#19"
           }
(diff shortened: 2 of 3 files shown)


asutherland:
> This is a first attempt at emitting structured information for a new type of C++ entity, enumerations.
> 
> It's currently causing the below difference in the output of an existing test (as well as a similar one in the same directory), which I haven't accepted yet because I'm not sure yet if everything in this change is ok.
> 
> The appearance of the new `"meta"` field is expected (that's the point of the patch), but the following are open questions in my mind:
> 
>     1. Why is the value of the `"pretty"` key next to `"meta"` changing from `"T_field_layout::field_type::Enum1"` to `"field_layout::field_type::Enum1"`?

There's [fallback logic here for generating the jumprefs that just uses the symbol](https://github.com/mozsearch/mozsearch/blob/c6a8cff709eb9b5aec8e48bfc50ec38a514e30fa/tools/src/file_format/crossref_converter.rs#L84-L88) so the presence of a structured record stops us from using the crappy fallback, I suspect.

>     2. In terms of the contents of the structured record, the patch is currently only filling out the `"kind"`, because that's the only thing needed by the semantic highlighting code that I'm using this for. Is that all right for now, or will other things get confused by the other fields in the record being null or empty?

It should be fine, although I think if you check out the [rust serde serialization labels](https://github.com/mozsearch/mozsearch/blob/c6a8cff709eb9b5aec8e48bfc50ec38a514e30fa/tools/src/file_format/crossref_converter.rs#L84-L88) you may see that in some cases we skip including fields in the output if we don't have them, so in some cases it might make more sense to emit omitting the fields if they don't make sense?  Like if the rust code wouldn't emit them, ideally don't have the C++ code emit them either.

theres-waldo:
> I think if you check out the [rust serde serialization labels](https://github.com/mozsearch/mozsearch/blob/c6a8cff709eb9b5aec8e48bfc50ec38a514e30fa/tools/src/file_format/crossref_converter.rs#L84-L88) you may see that in some cases we skip including fields in the output if we don't have them, so in some cases it might make more sense to emit omitting the fields if they don't make sense? Like if the rust code wouldn't emit them, ideally don't have the C++ code emit them either.

As far as I can tell, the C++ code already has this behaviour (it doesn't emit fields that the Rust code wouldn't emit). For example, for the field `"parentsym"` (one of the fields of `AnalysisStructured` marked [`skip_serializing_if`](https://github.com/mozsearch/mozsearch/blob/c6a8cff709eb9b5aec8e48bfc50ec38a514e30fa/tools/src/file_format/analysis.rs#L528)), the places where this is emitted in C++ code (like [this one](https://github.com/mozsearch/mozsearch/blob/c6a8cff709eb9b5aec8e48bfc50ec38a514e30fa/clang-plugin/MozsearchIndexer.cpp#L1392)) are all conditional on there being a valid a decl whose name to mangle.

Indeed, in the original analysis file emitted by `MozsearchIndexer`, a structured record for an enum looks like this:

> {"loc":"00019:11-16","structured":1,"pretty":"field_layout::field_type::Enum1","sym":"T_field_layout::field_type::Enum1","kind":"enum"}

The reason that additional fields appear in the final output for the `field-layout/field-type.cpp` testcase seems to be that these go through a `merge-analyses` pass (because of [this](https://github.com/mozsearch/mozsearch/blob/ad0d2e6fab18b51798559bf9ec496d0c15540070/tests/tests/build#L80-L82)), which then writes them out using the Rust serialization code.

Note that, while this causes new fields to appear in the output, the fields marked `skip_serializing_if`, such as `"parentsym"`, do not appear.

So, I don't _think_ there's anything to do here, but it's possible I'm overlooking something.

(Submitting PR for review.)

asutherland (approved):
Thanks for digging more into the JSON fields; hopefully it wasn't too much of a wild goose chase.  That's something where I think the automatically derived explanations could be useful/informative in the future when they're more readable, and if properly scoped.  (Like https://github.com/mozsearch/mozsearch/blob/master/tests/tests/checks/inputs/crossref/merge/merge__big_cpp provides coverage, but it's a whole giant file so the log explanation is basically useless.)

PR #780, opened 2024-06-29
Bug 1905491 - Add webtest for Symbol and Copy as Markdown in panel

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1905491

This does the following:
  * Use DEBUG type for logging for TestUtils internal and make it less appealing than other INFO
  * Add no-cache header to webtest files
  * Apply color to logging in web console [1], this helps debugging tests on browser
  * Add tests for Copy as Markdown menu items and accel keys
  * Add tests for Symbol section
  
[1] colored messages, compatible with both light and dark theme
log

### DIFF
--- a/tests/webtest/test_CopyAsMarkdown.js
+++ b/tests/webtest/test_CopyAsMarkdown.js
@@ -143,7 +143,7 @@ add_task(async function test_CopyAsMarkdown() {
      "[copy_as_markdown::CopyAsMarkdown::SomeMethod](BASE_URL/tests/source/webtest/CopyAsMarkdown.cpp#19)",
      "Method symbol is copied instead of local variable symbol");
 
-  // Select the local variable
+  // Shift-select from the start of the class to the currently selected local variable.
   selectLine(11, { bubbles: true, shiftKey: true });
 
   ok(!filenameButton.disabled, "Filename Link should always be enabled");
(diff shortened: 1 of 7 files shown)


tests/webtest/test_CopyAsMarkdown.js
```diff
@@ -0,0 +1,246 @@
...
+  ok(!filenameButton.disabled, "Filename Link should always be enabled");
+  ok(!symbolButton.disabled, "Symbol Link should be enabled when the selected line is inside a nesting line");
+  ok(!codeButton.disabled, "Code Block should be enabled when a line is selected");
+
+  TestUtils.click(symbolButton);
+  is(sanitizeURL(copiedText),
+     "[copy_as_markdown::CopyAsMarkdown::SomeMethod](BASE_URL/tests/source/webtest/CopyAsMarkdown.cpp#19)",
+     "Method symbol is copied instead of local variable symbol");
+
+  // Select the local variable
```
asutherland, line 146: Nit: Maybe this comment could be something like:

Shift-select from the start of the class to the currently selected local variable.

asutherland (approved):
This is so exciting to be gaining test coverage for all of this functionality!

I very much appreciate the care with which you've described the test assertions in their description strings and provided contextual comments like the rationale behind hooking the clipboard functionality.  You've also struck a really nice balance in terms of defining some support functions like that in the test file which I think is very nice for understanding how the test works!

asutherland:
Also, the screenshots are great, and as someone whose brain is so used to see syntax highlighting that looking at un-highlighted walls of log text becomes harder than it otherwise would be, I very much love having the various colors to make it easier to scan the log byproducts!  It's great!

PR #781, opened 2024-06-30
Bug 1905525 - Add test for page title and line numbers, and fix the "no line selected" case

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1905525

This does the following:
  * Add a text field in the webtest HTML to show the frame's title
  * Add test for selected symbols and sticky lines in the title
    * Add test-only event (`"titlechanged"`) to static JS files to help testing
  * Add test for line numbers in URL
  * Fix the hash handling to reflect the "no line selected" case to the URL (use the pathname instead of passing empty string)


asutherland (approved):
Nice!  It's great to have test coverage for this too!  (Especially since this is a thing I broke once! ;)

PR #782, opened 2024-07-07
Bug 1858453 - support using an objdir for clang-plugin via VPATH

asutherland:
In order to do a try run of indexing mozsearch, this change needs to be landed so that we can use a VPATH to do an objdir-based build.  This does work locally both for indexing mozsearch (by clobbering the Makefile into place) as well as for building the clang-plugin for that indexing.

The stanza in mozsearch/build ends up looking like:
```make
CLANG_PLUGIN_DIR=$FILES_ROOT/clang-plugin
make -f $CLANG_PLUGIN_DIR/Makefile VPATH=$CLANG_PLUGIN_DIR CC="$CC" CXX="$CXX"
```

### DIFF
--- a/clang-plugin/Makefile
+++ b/clang-plugin/Makefile
@@ -11,6 +11,10 @@ build: libclang-index-plugin.so
 %.o: %.cpp
 	$(CXX) $(CXXFLAGS) -c $^ -o $@
 
+from-clangd/%.o: from-clangd/%.cpp
+	mkdir -p from-clangd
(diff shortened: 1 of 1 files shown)


asutherland:
I'm going to land this patch immediately so I can do the try run and because I believe local testing means it's safe, but I'm happy to make changes.

https://stackoverflow.com/questions/43786731/makefiles-vpath-doesnt-work-when-searching-prerequisites-with-wildcards generally covers the need here; without this the VPATH works for the root clang-plugin dir but not the subdir.

PR #784, opened 2024-07-08
Add the tree switcher and the debugging UI, with adding the breadcrumbs and the navigation panel to all pages

arai-a:
For the following bugs:
  * https://bugzilla.mozilla.org/show_bug.cgi?id=1906146
  * https://bugzilla.mozilla.org/show_bug.cgi?id=1672721
  * https://bugzilla.mozilla.org/show_bug.cgi?id=1906223

This does the following:
  * [bug 1906146](https://bugzilla.mozilla.org/show_bug.cgi?id=1906146)
    * Add breadcrumbs and navigation panel to the following pages (navigation panel is collapsed by default for most cases):
      * Top page (breadcrumbs is hidden) : https://arai.searchfox.org/
      * Settings page (breadcrumbs is hidden) : https://arai.searchfox.org/mozilla-central/pages/settings.html
      * Search result: https://arai.searchfox.org/mozilla-central/search?q=EmitterScope&path=&case=false&regexp=false
      * Query result: https://arai.searchfox.org/mozilla-central/query/default?q=field-layout%3A%27js%3A%3Afrontend%3A%3AEmitterScope%27
      * Directory listing: https://arai.searchfox.org/mozilla-central/source/js/src/frontend
      * commit view: https://arai.searchfox.org/mozilla-central/commit/dea459eb01a4c38b696ae3d31c1540e86365a937
      * diff view (navigation panel was already there): https://arai.searchfox.org/mozilla-central/diff/b7d7259f7dbd6b171292f569eba66133ce352612/js/src/frontend/EmitterScope.cpp
    * Align the gear icon to the right side of the navigation panel header
      * This makes the position of the gear icon consistent across expanded and collapsed state
    * Add tests
      * Add config for webtest, which builds both tests and searchfox repositories, in order to test the behaviors with commits and histories
  * [bug 1672721](https://bugzilla.mozilla.org/show_bug.cgi?id=1672721)
    * Add tree switcher UI next to the tree name in the breadcrumbs [1], [2]
  * [bug 1906223](https://bugzilla.mozilla.org/show_bug.cgi?id=1906223)
    * Add debugging-purpose items to the navigation panel
      * Link to raw analysis records file in source listing [3]
      * Show debug log for query (`&debug=true` parameter for query) [4]
      * Show `results` JSON for the liquid template in query [4]
    * Add setting for the debug UI [5]
      * Enabled by default on non-production case

[1] Tree switcher button next to the tree menu
1-tree-switcher-button

[2] Tree switcher menu
2-tree-switcher-menu

[3] Debugging UI setting
3-debug-setting

[4] Raw analysis records link at the end of navigation panel
4-raw-analysis-link

[5] debug log and results JSON items in query page
5-query-debug


static/js/context-menu.js
```diff
@@ -866,3 +899,144 @@ function getTargetWord() {
 
   return string.substring(start, end);
 }
+
+var TREE_LIST = [
```
asutherland, line 903: Requesting changes: We should extract the TREE_LIST out into a JS file that can live in the root of the config directory.  This is not only a win for any other mozsearch instances out there, but also will make it easier for people making changes to mozsearch-mozilla to not miss the need to update the tree list.

It might makes sense to defer making this change until after getting some accessibility feedback if it ends up making more sense to let this just be an HTML snippet or something.

asutherland (changes requested):
This is super-great!  And the test coverage is amazing!

For the tree switcher UI, I think it would be great to try and solicit feedback from the `#accessibility` channel on slack.  My impression of the channel is that it's fairly active and people are responsive to requests there.

Currently when I look at the accessibility tree it looks like each column is a single list with the headers seeming to be homogeneous with the tree links other than the trees being links.  Like in the second column where we have "Firefox other" and "Thunderbird" in the same list.  I understand this is following the existing context-menu idiom, but I think the access patterns are different as the tree-switcher will have rarely changing contents and has a massive number of entries inside.

asutherland (approved):
Thanks!  I see you asked in the accessibility channel about this, so I'll consider that in progress.  If you don't hear anything back in a timely fashion I think it should be okay to land this and to address anything as a follow-up.  I did validate the tab process made sense and keyboard interaction made sense and the label for the combobox icon and the functionality does not appear to actively intrude, so I think it's unlikely to break existing work-flows (especially since the bread-crumbs are already variable length.)  r=asuth for this and the mozsearch-mozilla PR that will include the tree-list.js.

asutherland:
Woooo! Thank you!

static/js/context-menu.js
```diff
@@ -933,39 +950,180 @@ var TreeSwitcherMenu = new (class TreeSwitcherMenu extends ContextMenuBase {
...
+    }
 
-      lists.push(list);
+    // Fallback
+    return "mozilla-central";
+  }
+
+  focusCurrentTree() {
+    const tree = this.getCurrentTree();
+    const item = this.menu.querySelector(`a[data-tree="${tree}"`);
```
asutherland, line 1047: I think this selector is missing a closing-bracket, although it seems like our implementation doesn't care and this works fine either way.  (That said, the [CSS grammar does want the closing square brace](https://drafts.csswg.org/selectors-4/#typedef-attribute-selector).)

asutherland:
Really amazing test coverage for the great accessibility fixes, by the way!

static/js/context-menu.js
arai-a: Thanks! filed https://bugzilla.mozilla.org/show_bug.cgi?id=1909096. will fix it shortly

PR #786, opened 2024-07-15
Emit structured info for EnumConstantDecl

theres-waldo:


theres-waldo:
In this case, the patch has no impact on existing testcases.

I did verify that it has an effect on semantic tokens with my updated draft of https://github.com/mozsearch/mozsearch/pull/757, but if you would prefer I can also add a test case here that makes structured analysis data for an enum constant show up somewhere.

asutherland (approved):
> In this case, the patch has no impact on existing testcases.
>
> I did verify that it has an effect on semantic tokens with my updated draft of https://github.com/mozsearch/mozsearch/pull/757, but if you would prefer I can also add a test case here that makes structured analysis data for an enum constant show up somewhere.

This is fine without a test case given how straightforward and minimal this is and that we don't have existing behavior to meaningfully regress.

PR #789, opened 2024-07-21
Bug 1908197 - Add role for items inside tree switcher menu

arai-a:
https://bugzilla.mozilla.org/show_bug.cgi?id=1908197

This does:
  * Add `role=group` for `ul` for each group in Tree Switcher  (Firefox, Thunderbird)
  * Add `role=menuitem` for `a` for each item
  * Add `role=none` for `li` for each item


asutherland (approved):
Thanks!

PR #790, opened 2024-07-21
Bug 1906580 - Use consistent padding for .content box

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1906580

Instead of applying wider padding for non-source-listing cases, add padding/margin for items inside it when necessary, so that the breadcrumbs is positioned in the same place across pages.

Spacing for the following are kept by adding extra padding/margin:
  * directory listing table
  * field layout table

Spacing for the top page and the settings page are reduced, but they should already have sufficient padding:

Spacing for other queries (call graph, use graph) are also reduced, but they have spacing in its own.

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -276,6 +276,10 @@ table {
   width: 100%;
   overflow: auto;
 }
+table.folder-content {
+  margin: 0.5rem 2rem;
+  width: calc(100% - 4rem);
+}
 table#file {
   width: 100%;
 }
@@ -603,9 +607,6 @@ tr.after-context-line + tr.before-context-line {
   font-weight: bold;
 }
 .content {
-  padding: 1rem 2rem;
-}
-.content.source-listing {
   padding: 0.5rem 0rem;
 }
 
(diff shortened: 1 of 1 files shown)


static/css/mozsearch.css
```diff
@@ -276,6 +276,10 @@ table {
   width: 100%;
   overflow: auto;
 }
+table.folder-content {
+  margin: 0.5rem 2rem;
+  width: calc(100% - 4rem);;
```
asutherland, line 281: nit: extra semicolon

asutherland (approved):
It definitely feels correct to have the breadcrumb positions be consistent; before we had them on all the different pages it wasn't as clear to me what the right thing to do was.

Thanks!

asutherland:
I vaguely remembered some case I think that might have led me to leave things the way they were, it might have been this:

image

rs=asuth for fixing that (and presumably it can happen on this existing bug, etc.)

asutherland:
to be clear, the "Number of results" is right up against the left side of the page with no padding/margin, and we probably want some.  The screenshot's context might make that less clear.

PR #791, opened 2024-07-21
Bug 1908125 - Add keyboard navigation to context menu

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1908125
depends on https://github.com/mozsearch/mozsearch/pull/789

This does:
  * Add role for context menu
  * Move the keyboard navigation from `TreeSwitcherMenu` to `ContextMenuBase`
  * Add keyboard navigation for context menu (up/down/pageup/pagedown/home/end/esc)

Compared to the tree switcher, the context menu focuses the menu itself when opened,
and on macOS, hitting down focuses the first item, and hitting up focuses the last item,
this patch follows the behavior.

### DIFF
--- a/static/js/context-menu.js
+++ b/static/js/context-menu.js
@@ -5,6 +5,7 @@ function atUnescape(text) {
 class ContextMenuBase {
   constructor() {
     this.menu = null;
+    this.columns = [];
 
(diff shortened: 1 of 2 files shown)


asutherland (approved):
Woo!

PR #795, opened 2024-07-26
Bug 1910106 - Make introducedIn field for settings required, and verify the consistency on testing env.

arai-a:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1910106

This does:
  * Add `introducedIn` property for each setting
  * Make `introducedIn` property required
  * Verify the consistency (existence of `introducedIn`, and make sure `introducedIn` is within the `SETTINGS_VERSION`), on non-production env


asutherland (approved):
Thank you!  I'd considered the hash approach initially but it lost out on not being able to establish an ordering.  This is a great approach since it adds useful information that can help us improve the UI!

PR #796, opened 2024-07-26
Bug 1297090 - Make all reserved words contextual.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1297090

This does the following:
  * Skip analysis records for the file itself (`loc: "00001:0"`) when formatting the source
  * Search for analysis record for all keywords, and if analysis records are found, make it regular identifier with `data-symbols`
    * This effectively makes all reserved words contextual
  * Add rust subdirectory and make it possible to add rust test files
  * Enable python indexer in tests repo

There's one unintended change from this, which is `self` in Rust code.
Without this patch, `self` is treated as keyword, but apparently rust's analyzer generates analysis record for `self`, which makes `self` an identifier instead of reserved word while formatting.
I suppose this is not too much problematic, but if it's not okay, I'll add exception.


tools/src/format.rs
```diff
@@ -77,6 +77,15 @@ pub fn format_code(
 
     let mut cur_datum = 0;
 
+    while cur_datum < analysis.len() {
```
asutherland, line 80: It would be nice to have a comment here to provide context to someone newer to the codebase.  And/or if you want to extract `loc.lineno == 1 && loc.col_start == 0 && loc.col_end == 0` into a self-descriptive helper, that might make it more obvious what's going on.

asutherland (approved):
Really nice and elegant refinement here!  And with truly excellent test-case coverage!  Thanks!

> There's one unintended change from this, which is self in Rust code.
Without this patch, self is treated as keyword, but apparently rust's analyzer generates analysis record for self, which makes self an identifier instead of reserved word while formatting.
I suppose this is not too much problematic, but if it's not okay, I'll add exception.

This seems fine; I imagine this is something where once Botond's changes to support semantic styling in C++ it would be more clear how to best implement this for rust.

PR #797, opened 2024-07-27
Bug 1910230 - Do not try to access blame data when no blame info is provided.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1910230

This does:
  * skip blame info popup handling when there's no blame info
  * minor style fix


asutherland (approved):
Thanks!

PR #799, opened 2024-07-27
Bug 1906144 - Show class layout menu item for empty subclass.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1906144

This does:
  * Show "Show class layout of" menu item for class which has superclasses
    * this is shown regardless of the number of superclass fields for simplicity


asutherland (approved):
There were several times I wanted this enhancement this week, thank you!

PR #800, opened 2024-07-27
Bug 1910254 - Support generated files in fetch_raw_source

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1910254

The issue is caused by accessing missing file while fetching the source lines, due to mis-handling `__GENERATED__/*` path.

This patch stack does:
  * Ignore missing file in field layout table, so that similar case results in empty line cell, instead of error page
  * Support fetching generated files


arai-a:
table with class in generated files: https://arai.searchfox.org/mozilla-central/query/default?q=field-layout%3A%27mozilla%3A%3Adom%3A%3AClientManagerParent%27

asutherland (approved):
Great investigation, fix, and tests!  Thank you!

I'm going to land this now so it can get in the imminent indexer run.

PR #801, opened 2024-07-28
Bug 1904320 - Support interface mixin

arai-a:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1904320

This does:
  * Add testcase for existing WebIDL bindings, with pseudo binding code
    * Add testing-only config entry `webidl_binding_local_path` to override the webidl binding path, to simplify the automated tests
  * Support indexing interface mixins, with associating to the binding code for the interfaces that include the mixin


asutherland (approved):
Nice!  And it's really fantastic to have this level of test coverage through to the context menus!

PR #802, opened 2024-07-28
Bug 1910266 - Do not index files with non-ASCII characters in filename

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1910266

Most of the indexer scripts and server scripts don't actually expect non-ASCII (especially JS indexer), and each of them needs to be tweaked to properly support non-ASCII.
This makes the indexer simply ignore files with non-ASCII filename, to suppress warnings, as a first step.

### DIFF
--- a/scripts/find-repo-files.py
+++ b/scripts/find-repo-files.py
@@ -48,6 +48,12 @@ for line in lines:
         continue
     path = path.decode()
 
+    if path.startswith('"') and path.endswith('"'):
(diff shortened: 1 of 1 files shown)


arai-a:
I'll land this as a quick workaround, given those files are not indexed properly anyway.
I'll address the actual indexing in [bug 1910279](https://bugzilla.mozilla.org/show_bug.cgi?id=1910279) etc

asutherland:
Thanks!

PR #803, opened 2024-07-28
Bug 1910277 - Support spaces in filename for JS files

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1910277

This does:
  * in JS indexer, properly split the lines passed from `parallel` command, so that all characters after the first space becomes the filename.


asutherland (approved):
Yay!

PR #804, opened 2024-07-28
Bug 1910279 - Properly handle non-ASCII filenames in JS indexer

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1910279

This does:
  * Handle non-ASCII filenames in `js-analyze.js`:
    * input: decode the raw bytecode sequence from `readline()`
    * output: set `LC_TYPE` environment variable to make `std::wcsrtombs` called inside `os.file.redirect` work
  * Put raw non-ASCII paths into `*-files` files
    * things required escape/quote are re-enabled (they're skipped by [bug 1910279](https://bugzilla.mozilla.org/show_bug.cgi?id=1910279))
    * paths with special characters are still skipped

### DIFF
--- a/tests/webtest/test_UTF8InFilename.js
+++ b/tests/webtest/test_UTF8InFilename.js
@@ -1,0 +1,36 @@
+add_task(async function test_UTF8InFilename_Search() {
+  await TestUtils.loadPath("/");
+  TestUtils.shortenSearchTimeouts();
+
+  const query = frame.contentDocument.querySelector("#query");
+  TestUtils.setText(query, "SymbolInFilenameWithUTF8");
+
+  const content = frame.contentDocument.querySelector("#content");
+
+  await waitForCondition(
+    () => content.textContent.includes("Core code (1 lines") &&
+      content.textContent.includes("Definitions (SymbolInFilenameWithUTF8) (1 lines") &&
+      content.textContent.includes("var SymbolInFilenameWithUTF8"),
+    "symbol in file with space in filename matches as definition");
+});
+
+add_task(async function test_UTF8InFilename_DirAndFile() {
+  await TestUtils.loadPath("/tests/source/js");
+
+  const link = frame.contentDocument.querySelector(`.folder-content a[href*="with-UTF8-"]`);
+  ok(!!link, "UTF-8 filename is listed");
+  is(link.getAttribute("href"), "/tests/source/js/with-UTF8-ファイル.js",
+     "UTF-8 path is used with raw text");
+
+  const loadPromise = TestUtils.waitForLoad();
+  TestUtils.click(link);
+  await loadPromise;
+
+  ok(frame.contentDocument.location.href.includes(
+       "with-UTF8-%E3%83%95%E3%82%A1%E3%82%A4%E3%83%AB.js"),
+     "Navigated to the file page with URL-encoded path");
+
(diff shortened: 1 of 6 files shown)


tests/webtest/test_UTF8InFilename.js
```diff
@@ -0,0 +1,36 @@
...
+    "symbol in file with space in filename matches as definition");
+});
+
+add_task(async function test_UTF8InFilename_DirAndFile() {
+  await TestUtils.loadPath("/tests/source/js");
+
+  const link = frame.contentDocument.querySelector(`.folder-content a[href*="with-UTF8-"]`);
+  ok(!!link, "UTF-8 filename is listed");
+  is(link.getAttribute("href"), "/tests/source/js/with-UTF8-ファイル.js",
+     "UTF-8 path is used with raw text");
```
arai-a, line 23: Although raw text is working right now, we might want to URL-encode the path when embedding into HTML.

asutherland: It seems like these are [legal URL code points](https://url.spec.whatwg.org/#url-code-points) and so this might actually be the most readable thing we can do?  The chain of HTML spec definitions leading to that is:
- [The href attribute on a and area elements must have a value that is a valid URL potentially surrounded by spaces.](https://html.spec.whatwg.org/multipage/links.html#attr-hyperlink-href)
- [A string is a valid URL potentially surrounded by spaces if, after stripping leading and trailing ASCII whitespace from it, it is a valid URL string.](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#valid-url-potentially-surrounded-by-spaces)
- https://url.spec.whatwg.org/#valid-url-string
- https://url.spec.whatwg.org/#absolute-url-with-fragment-string
- https://url.spec.whatwg.org/#absolute-url-string
- https://url.spec.whatwg.org/#scheme-relative-special-url-string
- https://url.spec.whatwg.org/#path-absolute-url-string
- https://url.spec.whatwg.org/#path-relative-url-string
- https://url.spec.whatwg.org/#url-path-segment-string
- https://url.spec.whatwg.org/#url-units
- https://url.spec.whatwg.org/#url-code-points

asutherland: oh, right, but maybe we would want to encode the spaces, which is maybe what you meant

asutherland: Also, I should say, I'm not opposed to just encoding the URLs, but I often will check where URLs go in the Firefox URL target hover thing in the lower left, and obviously a fully encoded URL is much more inscrutable than the non-encoded version.

asutherland (approved):
Thanks!

arai-a:
I meant all non-ASCII range (so that the URL becomes all ASCII), but I didn't know most of them are actually valid.
Thank you for checking the spec!

I'll look into encoding spaces etc in follow.

arai-a:
confirmed the mozilla-mobile indexing works

https://arai.searchfox.org/mozilla-mobile/source/firefox-ios/firefox-ios/Client/Assets/Images.xcassets/Enhanced%20Tracking%20Protection/crossSiteTrackersImage.imageset
https://arai.searchfox.org/mozilla-mobile/search?q=crossSiteTrackersImage.imageset&path=&case=false&regexp=false

PR #805, opened 2024-07-29
Update log analyzer script for replace-aliases.sh, extend indexer runtime limit to 12 hours from 10 hours

asutherland:
I will regenerate the lambda tarball for config4 and upload it, as I believe that's necessary for us to see this change actually take effect in production.  As noted in the bug, we could do more to make this something that's specified by the config file.  Although there's always a trade-off in fail-safes like this where the more complexity you add, the more possible it is to fail in a way that the fail-safe never gets put in place.

### DIFF
--- a/infrastructure/aws/trigger_indexer.py
+++ b/infrastructure/aws/trigger_indexer.py
@@ -8,7 +8,7 @@ from trigger_common import TriggerCommandBase
 # dynamically generated hard-coded python script thing.
 class TriggerIndexerCommand(TriggerCommandBase):
     def __init__(self):
-        super().__init__('indexer', 'index.sh', 10)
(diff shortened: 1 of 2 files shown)


asutherland:
I'm going to land this now because the changes are fairly trivial and it's best practice to generate the lambda tarball from the HEAD.

asutherland:
I ended up updating them all using the automation I'd introduced for that and as documented at https://github.com/mozsearch/mozsearch/blob/master/docs/aws.md#automated-ish-updates.

I will keep an eye on whether the config1 lambda has any problems launching things tonight.

Note that I discovered we are failing to provision `zip` inside the docker image, so I will put up a patch momentarily to correct that issue.  (We actually upload zips, not tarballs.)

PR #807, opened 2024-08-01
Bug 1911051 - Fix replace-aliases.py not to filter out all records with URL_ substring match

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1911051

`replace-aliases.py` tries to replace records with `URL_*` symbols and `RELPATH` symbols,
but I forgot to actually test the symbol after a preliminary test to avoid JSON parse.

This patch does:
  * Add the test for the symbol to actually filter out non-target


asutherland (approved):
Makes sense!  Nice to have the additional coverage!  And nice catch, Nicolas!

PR #808, opened 2024-08-01
Bug 1911131 - Forward all relevant jumpref data to SYM_INFO.

nicolas-guichard:
When populating SYM_INFO, only the first symbol of merged symbols was output to HTML. This led to missing “Go to definition” entries in some cases involving macros or templates.

For templates, see the new test files templates6.h/templates6.cpp and the corresponding webtest.

For macros this is especially visible in the atom_list.h use-case, except in our test repo because atom_list.h doesn't go through the merge-analyses step.

### DIFF
--- a/tests/tests/build
+++ b/tests/tests/build
@@ -66,7 +66,8 @@ CPP_FILES=$(
 find . -name '*.cpp' | \
   grep -v /ipdl/ | \
   grep -v /field-layout/ | \
-  grep -v /macro.cpp
+  grep -v /macro.cpp | \
+  grep -v /templates6.cpp
 )
 
 ## C++ Build Stuff
@@ -81,6 +82,7 @@ done
 MULTI_CPP_FILES=$(
   find ./field-layout/ -name '*.cpp'
   echo "macro.cpp"
+  echo "templates6.cpp"
 )
 
 for PLATFORM in linux64 macosx64 win64
--- a/tests/tests/checks/snapshots/web/dirs/check_glob@root_listing__html.snap
+++ b/tests/tests/checks/snapshots/web/dirs/check_glob@root_listing__html.snap
@@ -630,6 +630,18 @@ Spread over multiple lines.
         </tr>
 
         <tr>
+          <td><a href="/tests/source/templates6.cpp" class="mimetype-fixed-container mimetype-icon-cpp">templates6.cpp</a></td>
+          <td class="description"><a href="/tests/source/templates6.cpp" title=""></td>
+          <td><a href="/tests/source/templates6.cpp">142</a></td>
+        </tr>
+
+        <tr>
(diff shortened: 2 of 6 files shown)


asutherland:
I've triggered your branch on asuth.searchfox.org [1] on config1.json to check if there's any potential fallout related to my original XXX comment.  I'll report back when it's done.

1: I used asuth instead of kdab in case you wanted to run some other run and I don't have anything else I need on my channel right now.

asutherland (approved):
Thanks for addressing this longstanding issue and adding a test!

We will probably want a follow-up on this, as the extra "Search for" entries aren't useful without the expansion context; it might make sense to fold those together, [arbitrary example line].
image

Also we end up with some ~related weirdness on [generated header file includes](https://asuth.searchfox.org/mozilla-central/source/xpcom/ds/nsGkAtoms.h#104) where we do actually have different `FILE_PLATFORM@PATH` symbols where I suppose our merge step potentially needs to understand how to rewrite `FILE_` symbols when we are able to consolidate them (as I believe we are in this case.)
image

But I do think this represents a nice step forward and it looks like the places where

nicolas-guichard:
@asutherland You did not finish that last sentence “it looks like the places where…”?

For the context menu, would you rather see the expansion between goto and search:
- goto def 1
- expansion 1
- search for 1
- goto def 2
- expansion 2
- search for 2

or below goto and search?
- goto def 1
- search for 1
- expansion 1
- goto def 2
- search for 2
- expansion 2

asutherland:
> @asutherland You did not finish that last sentence “it looks like the places where…”?

Oops; I got too focused on putting specific example links in!

That should have fully been: But I do think this represents a nice step forward and it looks like the places where this is happening are macro/#include-specific, so it's not like it's happening everywhere.
 
> For the context menu, would you rather see the expansion between goto and search:

In at least the cases I link above, I think we actually just want all the searches for a given pretty identifier merged into a single search.  The search endpoint's processing of `symbol:` supports the value being a comma-delimited list of symbols, so this works.  So instead of directly pushing search items we'd populate a Map from "pretty" keys to "array of symbols", and then flatten that out at the end.  Right now for the `GK_ATOM` case, the distinct searches end up showing effectively all the same uses, just different defs for each search.  Historically/before I changed us to SYM_INFO, I think this merged search is actually what we would have done.

If there are cases where we do think we want distinct search items, we should probably be differentiating the pretty identifier anyways, which still works with this approach.  But unless there are some really interesting macro cases we find, I think that will probably only become an issue as we improve our template handling.

Specifically, for templates I think we could have multiple C++ overloads called from a single token, where our pretty identifiers not including the full method signature would potentially make us merge multiple symbols that maybe there could be some value to not folding.  That's a case where we might want to dynamically compute an explanation of the difference between the overloads.  But since searchfox searches historically have been over-inclusive, it also isn't the end of the world if we're merging things.  (And ideally with the "query" endpoint's search behavior, we'll get faceting sorted out which will make it easier to slice and dice the results.)

PR #810, opened 2024-08-08
Bug 1912078 - Failed attempts to make config5 / wubkat happy

asutherland:
I think the basic situation is that the macro expansion magic is interacting badly with webkit in a way that causes the processes memory usage to bloat to really extreme levels.  This stack provides swap space, and the webkit indexing job had consumed about 190G of swap before I terminated the index.  The swapping was pretty bad, I think, to the extent that basically of the 16 cores, we only ended up with about 1 core's worth of progress moving forward.

The analysis files on disk did not seem to be large; only 222 MiB was written when I shut things down.

I'm landing this stack primarily because I think we do want all these changes[1] and because the IMDS change will break future indexing runs without the last patch in this stack (because the new AMIs got opted in to IMDSv2 with required token).

I have disabled the "everyday-5am" rule from Amazon EventBridge (busses/rules, but accessible from cloudwatch/rules where I guess it used to be) because there's no point in scheduling webkit jobs when we think they're going to fail.

One disclaimer is that the 8 GiB swap logic for non-config5 servers has not been tested and it's conceivable I made an error there.  I can take a look tomorrow if any jobs fail and re-trigger them.

1: In particular, I do think it makes sense to run the webkit build jobs on a more powerful machine.  This potentially might also make sense for our LLVM job too for the same reason.

### DIFF
--- a/infrastructure/aws/index.sh
+++ b/infrastructure/aws/index.sh
@@ -20,7 +20,7 @@ CONFIG_REPO_URL=$4
 CONFIG_REPO_PATH=$(readlink -f $5)
 CONFIG_FILE_NAME="$6"
 
-EC2_INSTANCE_ID=$(wget -q -O - http://instance-data/latest/meta-data/instance-id)
+EC2_INSTANCE_ID=$(ec2metadata --instance-id)
 
 # Find the revisions we actually checked out so that we can pass these to the
 # web server because indexer and web server are a matched pair and we don't
--- a/infrastructure/aws/mkscratch.sh
+++ b/infrastructure/aws/mkscratch.sh
@@ -8,40 +8,94 @@ set -o pipefail # Check all commands in a pipeline
 # byproducts live while the indexing is ongoing.  To do this, we need to figure
 # out what the device's partition is.
 #
-# If we run `nvme list -o json` we get output like the following:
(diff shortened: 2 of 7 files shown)


asutherland:
And indeed an unexpected thing happened with the jq processing pipeline for the m5d.2xlarge case that was definitely not what I was expecting.  (And I'd tested the behavior of the `nth` predicate using the jqplay website on what should have been the right canned JSON data.)  I'll fix now.

asutherland:
The problem is our provisioned version of jq is 1.6 which apparently has a broken `nth` predicate or which otherwise diverges from the current jq 1.7.1 and which it appears jqplay.org is using (which makes sense), as well as what my machine is using.  I will update our version of jq... I could swear I had already done this, but maybe that was on a WIP branch.

asutherland:
I'm going to quickly see what happens if I bump our Ubuntu release to the ~new 24.04 LTS release, first on docker, then provisioning.  That gets us a modern jq and in general we don't have any reason to prefer 22.04.  ([Ubuntu package info for jq](https://packages.ubuntu.com/search?keywords=jq))

asutherland:
Updating to 24.04 ends up being pretty involved; I think I've nearly got it there, but I'll be in transit soon, so I'm just going to change us to pull jq from the github release in the provisioning stage rather than trying to get it via apt.

PR #813, opened 2024-09-01
Bug 1913921 - Attach event listeners before search result table is added to the document

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1913921

This does:
  * search for the "expando" buttons in the `table` node instead of `document`, given the table is not yet added to the document at the point


emilio (approved):

asutherland:
Thanks!

PR #814, opened 2024-09-05
Bug 1916972 - Fix lifetime annotations to avoid warnings

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1916972

This adds lifetime annotation to avoid the warnings introduced in rust nightly compiler.


nicolas-guichard (approved):
Coincidentally I also got that when trying to build MC with rust-analyzer nightly.

PR #815, opened 2024-09-05
Bug 1912078 - Avoid huge memory usage by setting clang-format ColumnLimit to 0.

nicolas-guichard:
Searchfox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1912078

### DIFF
--- a/clang-plugin/MozsearchIndexer.cpp
+++ b/clang-plugin/MozsearchIndexer.cpp
@@ -2534,17 +2534,37 @@ public:
   }
 
   void endMacroExpansion() {
-    const auto replacements = clang::format::reformat(
-      clang::format::getMozillaStyle(),
-      MacroExpansionState->Expansion,
-      {tooling::Range(0, MacroExpansionState->Expansion.length())}
-    );
-    auto formatted = clang::tooling::applyAllReplacements(MacroExpansionState->Expansion, replacements);
-    if (formatted) {
-      for (auto &[k, v] : MacroExpansionState->TokenLocations) {
-        v = replacements.getShiftedCodePosition(v);
+    // large macros are too slow to reformat, don't reformat macros larger than those arbitrary thresholds
+    static constexpr auto includedFileExpansionReformatThreshold = 20'000;
+    static constexpr auto mainFileExpansionReformatThreshold = 200'000;
+
+    const auto expansionLocation = SM.getExpansionLoc(MacroExpansionState->MacroNameToken.getLocation());
+    const auto expansionFilename = SM.getFilename(expansionLocation);
+    const auto includedExtensions = std::array{".h", ".hpp", ".hxx", ".inc", ".def"};
+    const auto isIncludedFile = std::any_of(includedExtensions.begin(), includedExtensions.end(), [&](const auto *extension) {
+      return expansionFilename.ends_with_insensitive(extension);
+    });
+    const auto expansionReformatThreshold = isIncludedFile ? includedFileExpansionReformatThreshold : mainFileExpansionReformatThreshold;
+
+    if (MacroExpansionState->Expansion.length() < expansionReformatThreshold) {
+      // large macros are too memory-hungry to reformat with ColumnLimit != 0
+      // see https://github.com/llvm/llvm-project/issues/107434
+      auto style = clang::format::getMozillaStyle();
+      if (MacroExpansionState->Expansion.length() > includedFileExpansionReformatThreshold)
+        style.ColumnLimit = 0;
+
+      const auto replacements = clang::format::reformat(
+        style,
+        MacroExpansionState->Expansion,
+        {tooling::Range(0, MacroExpansionState->Expansion.length())}
+      );
+      auto formatted = clang::tooling::applyAllReplacements(MacroExpansionState->Expansion, replacements);
+      if (formatted) {
+        for (auto &[k, v] : MacroExpansionState->TokenLocations) {
+          v = replacements.getShiftedCodePosition(v);
+        }
+        MacroExpansionState->Expansion = std::move(formatted.get());
       }
-      MacroExpansionState->Expansion = std::move(formatted.get());
     }
 
     IdentifierInfo *Ident = MacroExpansionState->MacroNameToken.getIdentifierInfo();


nicolas-guichard:
My test indexer run failed with:
```
-- Checking for module 'sysprof-capture-4'
--   Found sysprof-capture-4, version 3.44.0
-- Could NOT find SysProfCapture (missing: SysProfCapture_INCLUDE_DIR) (found version "3.44.0")
CMake Error at Source/CMakeLists.txt:40 (message):
  system libsysprof-capture-4 not found, consider using
  USE_SYSTEM_SYSPROF_CAPTURE=NO


-- Configuring incomplete, errors occurred!
```

nicolas-guichard:
I ran an indexer run with this and https://github.com/mozsearch/mozsearch-mozilla/pull/261, currently available on kdab.searchfox.org.
Without reformatting them, any multiline macros are basically unreadable…

What do you think about bailing out on super long macros? Something like `if (MacroExpansionState->Expansion.length() < 10000) { reformat } else { don't }`. The main offending macro is 2.2M chars long.
I'm unsure of the right threshold. I'll do some local testing.

nicolas-guichard:
(I still think it would be more efficient to move the reformatting after the merge step, but it's a lot of work compared to just setting a threshold like this.)

asutherland:
I agree a threshold is probably the most pragmatic tradeoff for now.  I agree there's a lot more we could do about being smarter and not redoing work but that we should defer.  In particular, I think there's real potential for having the indexer plugin connect over a local socket to give all the lines it generates to a daemon instead of the current file locking approach.  This could fundamentally involve asking the daemon if it should generate data for some things, similar to how kythe has a "claim" mechanism to avoid redundant work, although I was initially picturing it as unidirectional for simplicity.

One thing I might say in terms of heuristics is that it could make sense to perhaps apply a multiplier to the length heuristic for `.cpp` files.  Specifically, my team has a bunch of `QM_TRY*` macros that can get quite large, but they almost entirely only exist in .cpp files.

I quickly ran `jq '.expandsTo[]?[]? | length' FILENAME | sort -n | uniq -c` against the large dom/indexeddb and dom/quota ActorsParent.cpp files and got (manually extracting the tail end of the lists).  Based on my very limited exploration here, I might suggest we have a 20k limit for header files and multiply that by 10 to 200k for cpp files.

for IDB:
```
      5 6670
      5 6688
      6 6902
      6 7155
      1 7231
      5 7239
      1 9711
      5 9743
      1 9858
      5 9930
      1 10310
      5 10420
      1 11069
      1 11137
      5 11165
      5 11169
      6 11263
      1 13175
      5 13181
      6 13575
      6 20777
      1 21691
      5 21723
      1 27834
      5 27882
      6 28318
      1 48848
      5 48887
      1 69692
      5 69795
```

for QM:
```
      6 2964
      6 3413
      6 3592
      1 4157
      5 4165
      6 4297
      6 4534
      6 4577
      1 4635
      5 4651
      5 6238
      1 6294
      6 6321
      6 6806
      6 8172
      1 9504
      5 9520
      6 12335
      1 15685
      5 15701
      6 24044
      6 43530
      1 116297
      5 116403
      1 192494
      5 192638
```

And then for some stuff from dom/base:

nsGlobalWindowInner.h:
```
     10 743
      5 2649
      1 2791
```

nsGlobalWindowInner.cpp:
```
      5 2428
      1 2460
      5 2508
      1 2540
      5 6550
      1 6606
```

Document.h:
```
      5 2689
      1 2823
      5 3809
      1 4086
```

Document.cpp:
```
      5 6954
      5 7414
      1 7416
      5 7539
      5 7565
      1 7585
      1 7612
```

Element.h:
```
      1 978
      5 986
      6 2542
```

Element.cpp:
```
      5 1257
      1 1447
      5 1455
      1 1512
      5 1520
```

clang-plugin/MozsearchIndexer.cpp
```diff
@@ -2534,17 +2534,27 @@ class IndexConsumer : public ASTConsumer,
...
-      for (auto &[k, v] : MacroExpansionState->TokenLocations) {
-        v = replacements.getShiftedCodePosition(v);
-      }
-      MacroExpansionState->Expansion = std::move(formatted.get());
+    static constexpr auto includedFileExpansionReformatThreshold = 20'000;
+    static constexpr auto mainFileExpansionReformatThreshold = 200'000;
+
+    const auto expansionLocation = SM.getExpansionLoc(MacroExpansionState->MacroNameToken.getLocation());
+    const auto expansionFileId = SM.getFileID(expansionLocation);
+    const auto isIncludedFile = SM.getIncludeLoc(expansionFileId).isValid();
```
asutherland, line 2542: It seems like this might get tricked by unified builds where the cpp files are `#include`d?

nicolas-guichard: Right, should I just filter by filename then?
I'm thinking of filtering out *.h, *.hpp, *.hxx and *.inc.

asutherland: Yeah, I think treating the most common include extensions as includes makes sense (given the horrible contortions used to make C++ builds fast).

asutherland:
I wonder if this is https://bugs.webkit.org/show_bug.cgi?id=275110 https://github.com/WebKit/WebKit/pull/30263 (but which doesn't seem to affect upstream builds)

nicolas-guichard:
With `USE_REV="b36cbce6"` the indexing failed with a panic in [blame-related code](https://searchfox.org/mozsearch/source/tools/src/format.rs#726):
```
File 'Source/WebCore/html/track/TextTrack.idl'
[2024-09-12T16:27:29Z INFO  tools::file_format::analysis] Error trying to open analysis file [/mnt/index-scratch/wubkat/analysis/Source/WebCore/html/track/TextTrack.idl]
  Analysis load duration: 21us
  Per-file info load duration: 678us
  File contents read duration: 155us
thread 'main' panicked at src/format.rs:726:65:
index out of bounds: the len is 56 but the index is 56
stack backtrace:
   0: rust_begin_unwind
             at /rustc/8d6b88b168e45ee1624699c19443c49665322a91/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at /rustc/8d6b88b168e45ee1624699c19443c49665322a91/library/core/src/panicking.rs:74:14
   2: core::panicking::panic_bounds_check
             at /rustc/8d6b88b168e45ee1624699c19443c49665322a91/library/core/src/panicking.rs:276:5
   3: <usize as core::slice::index::SliceIndex<[T]>>::index
             at /rustc/8d6b88b168e45ee1624699c19443c49665322a91/library/core/src/slice/index.rs:302:10
   4: core::slice::index::<impl core::ops::index::Index<I> for [T]>::index
             at /rustc/8d6b88b168e45ee1624699c19443c49665322a91/library/core/src/slice/index.rs:16:9
   5: <alloc::vec::Vec<T,A> as core::ops::index::Index<I>>::index
             at /rustc/8d6b88b168e45ee1624699c19443c49665322a91/library/alloc/src/vec/mod.rs:2924:9
   6: tools::format::format_file_data
             at ./mozsearch/tools/src/format.rs:726:65
   7: output_file::main
             at ./mozsearch/tools/src/bin/output-file.rs:448:15
   8: core::ops::function::FnOnce::call_once
             at /rustc/8d6b88b168e45ee1624699c19443c49665322a91/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Indeed `Source/WebCore/html/track/TextTrack.idl` shrunk from 59 lines to 56 lines between b36cbce6 and now, and the incremental `build-blame` probably doesn't like traveling back in time.

nicolas-guichard:
In summary, i'm confident this PR solves the initial issue (the build doesn't hang anymore), but for 2 new different reasons Webkit doesn't build on the indexer:
1. the `undefined reference to `WebKit::GPUProcessMain(int, char**)'` issue
2. the dependency install issue

And my initial attempt at traveling back in time to a known good commit failed.

asutherland (approved):
Thanks for all the investigations and troubleshooting here!

Agreed that this can land without waiting for the webkit build to be clean.  Please propagate these changes to the m-c version too and I can r+ there too.

asutherland:
I'm going to merge this because I'm going to do a local m-c run to quickly try doing what I proposed in https://bugzilla.mozilla.org/show_bug.cgi?id=1893181#c24 and figure I might as well run with these changes, but I am not planning to push the changes to m-c at this time.

PR #816, opened 2024-09-14
Switch from tree-sitter-kotlin to tree-sitter-kotlin-ng to address incorrect nesting in kotlin due to bad tree-sitter parses

asutherland:
More details in the commits and https://bugzilla.mozilla.org/show_bug.cgi?id=1918801#c0, but the tl;dr is the old grammar would get confused quite which results in bad nesting which hurts diagram quality.  Also, we had effectively forked the old grammar because they backed out the introduction of some added fields (see https://bugzilla.mozilla.org/show_bug.cgi?id=1885069), whereas the new parser has good parses and sufficient labels.  However, it didn't have body labels so we slightly enhance scip-indexer to support node-based queries[1].

1: I had speculated I might move the cst_tokenizer query approach over, but ended up punting on that.


asutherland:
I merged this as part of #817, closing this PR.

PR #818, opened 2024-09-24
Bug 1920754 - Stop using global variable and mutable static in url_map_handler.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1920754

This does the following:
  * instead of using global variable to keep the url map JSON file path, associate it to `Config` and pass it to `linkify_comment` function
  * instead of `lazy_static`, use `OnceLock` to lazily construct the `URLMap` with passed config's url map JSON path

Some places are not passing the config, thus the URLs in comments are not linkified, but that's pre-existing.

### DIFF
--- a/tools/src/file_format/config.rs
+++ b/tools/src/file_format/config.rs
@@ -164,6 +164,7 @@ pub struct Config {
     pub trees: BTreeMap<String, TreeConfig>,
     pub mozsearch_path: String,
     pub config_repo_path: String,
+    pub url_map_path: Option<String>,
 }
 
 impl Config {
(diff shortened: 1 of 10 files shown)


tools/src/file_format/config.rs
```diff
@@ -164,6 +164,7 @@ pub struct Config {
     pub trees: BTreeMap<String, TreeConfig>,
     pub mozsearch_path: String,
     pub config_repo_path: String,
+    pub url_map_path: Option<String>,
```
asutherland, line 167: Structurally this is a bit weird since the url_map_path is conceptually tree-specific ([URL_MAP_PATH is defined using INDEX_ROOT](https://searchfox.org/mozsearch/rev/51ae44c9d9a80a6a53628248f5e5184766c83f71/scripts/mkindex.sh#69) which [is extracted from the tree config](https://searchfox.org/mozsearch/rev/51ae44c9d9a80a6a53628248f5e5184766c83f71/scripts/load-vars.sh#15)), so I think it would make more sense for this to be attached to the tree config.

That said, I think it does make sense to land this as-is (but maybe with a TODO or XXX) to quiet the warnings since in practice it doesn't matter because output-file is only ever used at the granularity of a single tree at a time[1].  And then we can address this in some future enhancement as part of a more comprehensive cleanup where it might matter.

1: It's really only the web-server.rs and pipeline-server.rs that ever want the fully populated config structure.

asutherland (approved):
Thanks for addressing these warnings so proactively!

PR #819, opened 2024-09-29
Bug 1921635 - Add "Go to use of" for diagram edges.

asutherland:
A notable correctness fix is that the `ensure_edge_in_graph` method was using a lookup map to find already extant edges but never actually populating the lookup map.  I believe this may have not affected anything because calls-to/uses diagrams used the traversal suppression to only add the edge once, and calls-from diagrams used callees edges which were pre-aggregated during crossref.

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -1353,6 +1353,18 @@ g.hovered-in-node > polygon {
   fill: #bfb;
 }
 
+g.hovered-cur-edge > path:not(.clicktarget),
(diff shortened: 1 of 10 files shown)


asutherland:
This is up at asuth.searchfox.org and https://asuth.searchfox.org/mozilla-central/query/default?q=calls-to%3ANS_InitXPCOM is one of the motivating diagrams.

arai-a (approved):
This is really useful feature :)

PR #820, opened 2024-10-04
Bug 1833695: reset DeclContext stack when analysing out-of-line templated function

nicolas-guichard:
Link to bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1833695

Without this, OutOfLineTemplateShouldntHaveContextSym had two definition target records that disagreed on the contextsym:
1. The first one was generated when traversing the declaration, because it forwards to the definition if out-of-line (see comment in TraverseFunctionDecl). We didn't reset CurDeclContext, so it looked as-if the definition appeared inside the declaration, leading to a bogus contextsym (pointing to the function itself).
2. The second one was generated later on when traversing the definition itself. This one had a correct contextsym (ie. none).

This introduces a ValueRollback helper to make sure we don't forget to rollback CurDeclContext afterwards. (inspired by QScopedValueRollback)

### DIFF
--- a/clang-plugin/MozsearchIndexer.cpp
+++ b/clang-plugin/MozsearchIndexer.cpp
@@ -154,6 +154,38 @@ struct RAIITracer {
 
 #define TRACEFUNC RAIITracer tracer(__FUNCTION__);
 
+// Sets variable to value on creation then resets variable to its original
+// value on destruction
+template<typename T>
+class ValueRollback {
+public:
+  template<typename U = T>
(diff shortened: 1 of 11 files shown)


asutherland (approved):
Thanks!  The ValueRollback helper is definitely a nice RAII helper!

Thanks to Botond for the initial reduced test-case and thank you for populating the full set of permutations and giving them names that make the very nice test case checks even easier to read and understand!

PR #821, opened 2024-10-04
Bug 1856762 - Visit implicit code inside lambdas

nicolas-guichard:
Link to bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1856762

This reuses the AutoSetContext machinery initially implemented for constructors to visit the implicit code generated for lambdas too, giving us the data required to understand the template-dependent code in lambdas.

AutoSetContext is adapted to support AutoSetContext::Decl == nullptr because, contrary to CXXConstructorDecl, LambdaExprs aren't NamedDecls.

### DIFF
--- a/clang-plugin/MozsearchIndexer.cpp
+++ b/clang-plugin/MozsearchIndexer.cpp
@@ -888,6 +888,11 @@ public:
     return Super::TraverseCXXDestructorDecl(D);
   }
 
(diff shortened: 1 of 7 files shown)


asutherland (approved):
Thanks!  This makes sense and I very much appreciate the minimal test case with the checks that make it clear what the net result is!

PR #823, opened 2024-10-07
Upgrade to Ubuntu 24.04 (multiple commits), double release4 indexer for time reasons

asutherland:
I just kicked off a provisioning pass for this.

For validation I fully purged my docker setup via `docker system prune --volumes -af` (after closing running docker instances) before invoking `./build-docker.sh`.  (One shouldn't do that if they use docker for things other than searchfox, of course, as it clears everything, but is an easy way to make sure there's no cached state problems.)

`make build-test-repo` and `make webtest` both passed with this current set of changes.

I'll plan to land this if the provisioning passes.  Then I'll also update the lambda jobs and re-trigger release1.


asutherland:
Provisioning completed, merging now and going to test the lambda job works on release 1.

asutherland:
The release1 lambda job operated correctly (and I'd previously validated the zipball has the updated release4 and release5 logic) and moved onto index-log stage.  I'll make sure it completed okay tomorrow.

PR #824, opened 2024-10-08
Bug 1923517 - Filter out the warning for protobuf.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1923517 and https://bugzilla.mozilla.org/show_bug.cgi?id=1923516

This does:
  * delete the default "ubuntu" user while setting up the docker image, to avoid the UID conflict with the host user
  * filter out the protobuf's warning, specific to the current version, so that the warning with newer versions will be reported even if we forget to remove the filter


asutherland (approved):
Thank you!!

PR #825, opened 2024-10-10
Bug 1912189 - Detect clang and clang-plugin version mismatch

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1912189

This makes the clang-plugin Makefile verify the clang version for the previous build (written into `clang-version.cache` file, which is not version-controlled), and rebuild from scratch if the version mismatch is detected.

I was about to add the `clang-version.cache` file to the dependency of each rule, but that results in putting the file in the compile/link command lines, so I made it to just perform `make clean` to perform full rebuild.

PR #826, opened 2024-10-11
Bug 1912189 - Part 2: Use the version check only in docker

arai-a:
for the failure in https://github.com/mozsearch/mozsearch/pull/825

This makes the commands used during testing in docker perform the version check, but the command used by server don't


asutherland (approved):
Thanks!

PR #827, opened 2024-10-11
Bug 1924227 - Extract root block device from nvme list rather than hardcoding it

asutherland:
In some cases the EBS root volume may not be the previously hardcoded `/dev/nvme0n1` so we should extract it via jq.  The idiom we've been using in cases like this is to run the command that gives us a JSON blob once into an INFO var and then use separate jq invocations that use `<<< "$FOO_INFO"` to then pick the relevant data out of there rather than issue multiple commands which might change dynamically.  I've done that here.

This successfully ran on both provisioning jobs I just kicked off, although the hardcoded values would have worked too, but I did consult the logs and we are using the new logic.

### DIFF
--- a/infrastructure/aws/indexer-provision.sh
+++ b/infrastructure/aws/indexer-provision.sh
@@ -61,12 +61,14 @@ date
 # command.  Note that the select constraint here is intended more as a check
 # that our assumption about partition sizes hasn't changed, as when provisioning
 # there should only be this single EBS mount.
-ROOT_VOL_ID=$(sudo nvme list -o json | jq --raw-output '.Devices[] | select(.PhysicalSize < 9000000000) | .SerialNumber | sub("^vol"; "vol-")')
+ROOT_DEV_INFO=$(sudo nvme list -o json | jq --raw-output '.Devices[] | select(.PhysicalSize < 9000000000)')
+ROOT_VOL_ID=$(jq -M -r '.SerialNumber | sub("^vol"; "vol-")' <<< "$ROOT_DEV_INFO")
+ROOT_DEV=$(jq -M -r '.DevicePath' <<< "$ROOT_DEV_INFO")
+
 AWS_REGION=us-west-2
 # The size is in gigs.
 aws ec2 modify-volume --region ${AWS_REGION} --volume-id ${ROOT_VOL_ID} --size 20
-# Re: hardcoded devices: The devices should currently be stable.
-#
+
 # We use an until loop because it can take some time for the change to
 # propagate to this VM.  The error will look like:
 #   "NOCHANGE: partition 1 is size 16775135. it cannot be grown"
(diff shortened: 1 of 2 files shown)


asutherland:
Here's an excerpt from the indexer provisioning run:

```shell
+ date
Fri Oct 11 20:42:06 UTC 2024
++ sudo nvme list -o json
++ jq --raw-output '.Devices[] | select(.PhysicalSize < 9000000000)'
+ ROOT_DEV_INFO='{
  "NameSpace": 1,
  "DevicePath": "/dev/nvme0n1",
  "GenericPath": "/dev/ng0n1",
  "Firmware": "1.0",
  "ModelNumber": "Amazon Elastic Block Store",
  "SerialNumber": "vol0d8371bb7169d78ea",
  "UsedBytes": 8589934592,
  "MaximumLBA": 16777216,
  "PhysicalSize": 8589934592,
  "SectorSize": 512
}'
++ jq -M -r '.SerialNumber | sub("^vol"; "vol-")'
+ ROOT_VOL_ID=vol-0d8371bb7169d78ea
++ jq -M -r .DevicePath
+ ROOT_DEV=/dev/nvme0n1
+ AWS_REGION=us-west-2
+ aws ec2 modify-volume --region us-west-2 --volume-id vol-0d8371bb7169d78ea --size 20
{
    "VolumeModification": {
        "VolumeId": "vol-0d8371bb7169d78ea",
        "ModificationState": "modifying",
        "TargetSize": 20,
        "TargetIops": 3000,
        "TargetVolumeType": "gp3",
        "TargetThroughput": 125,
        "TargetMultiAttachEnabled": false,
        "OriginalSize": 8,
        "OriginalIops": 3000,
        "OriginalVolumeType": "gp3",
        "OriginalThroughput": 125,
        "OriginalMultiAttachEnabled": false,
        "Progress": 0,
        "StartTime": "2024-10-11T20:42:06+00:00"
    }
}
+ sleep 5
+ sudo growpart /dev/nvme0n1 1
CHANGED: partition=1 start=2099200 old: size=14677983 end=16777182 new: size=39843807 end=41943006
+ sudo resize2fs /dev/nvme0n1p1
resize2fs 1.47.0 (5-Feb-2023)
Filesystem at /dev/nvme0n1p1 is mounted on /; on-line resizing required
old_desc_blocks = 1, new_desc_blocks = 3
The filesystem on /dev/nvme0n1p1 is now 4980475 (4k) blocks long.

+ date
```

arai-a (approved):
Looks good! Thanks!

asutherland:
This is so weird, I'm getting a problem generating a rebase when the commit is against trunk:
image

I'll permute the commit with an amend in the hopes it makes things happy.

asutherland:
That worked.  Presumably an automated check that was cached was made yesterday during system problems.

PR #828, opened 2024-10-13
Bug 1910314 - URL-encode link path, and support URL-encoded path in the server

arai-a:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1910314

This does:
  * URL-encode the `href` attribute when the path contains the files in the target repository
  * Decode URL-encoding on the server, for dynamically-generated pages

example:
https://arai.searchfox.org/mozilla-central/source/testing/web-platform/tests/tools/wave/tests/WAVE%20Local.postman_environment.json

Links in the blame popup, the navigation bar, and the breadcrumbs are now URL-encoded, and also the
annotated diff link target shows the correct file (it has been "File not found" error page)

### DIFF
--- a/static/js/blame.js
+++ b/static/js/blame.js
@@ -240,19 +240,19 @@ var BlamePopup = new (class BlamePopup {
       rendered += json[i].header;
 
       let diffLink = `/${tree}/diff/${revList[i]}/${revPath}#${linenoList[i]}`;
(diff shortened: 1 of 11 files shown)


asutherland (approved):
Thanks for the fantastic patch stack structuring and diligent test coverage throughout!

PR #829, opened 2024-10-13
Bug 1924329 - Show error when uncommitted changes found while building the searchfox repo

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1924329

This adds a check to the setup process of the searchfox repo (used both by build-searchfox-repo and webtest) to check
uncommitted changes, and show error if any, to avoid errors during the output process caused by the inconsistency between the git history and the working tree.


asutherland (approved):
Thanks!  This has been a longstanding footgun; it's great to have this addressed!

PR #832, opened 2024-10-17
Bug 1925207 - Convert the livegrep warning for the file size to INFO

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1925207

This adds a filter for the `run_showing_output` function, and pass a filter to convert "WARN:" to "INFO:" for the filesize warning.


asutherland (approved):
Thanks!  This was on my queue but it was going to be a few more days; I very much appreciate how proactive you are about keeping the warning emails as actionable instead of spammy!

PR #833, opened 2024-11-04
Bug 1699048 - Index StaticPrefs.

arai-a:
https://bugzilla.mozilla.org/show_bug.cgi?id=1699048

This does:
  * Index each item in `StaticPrefList.yaml`, with `PREFS_{id}` symbol
  * Create structured record for the `StaticPrefList.yaml` items, adding the C++ binding to the getter binding slot.
  * Support tokenizing `StaticPrefList.yaml`, with treating the pref name as identifier, supporting comments and strings

Example in `StaticPrefList.yaml`:
https://arai.searchfox.org/mozilla-central/source/modules/libpref/init/StaticPrefList.yaml#213

pref-menu

Example in the consumer of the binding:
https://arai.searchfox.org/mozilla-central/source/dom/base/FocusModel.h#36

binding-menu

### DIFF
--- a/scripts/staticprefs-analyze.py
+++ b/scripts/staticprefs-analyze.py
@@ -39,11 +39,18 @@ def read_cpp_analysis_one(path, cpp_symbols):
 
         sym = j['sym']
         m = re.match('.+StaticPrefs[0-9]+(.+)Ev$', sym)
-        if not m:
-            continue
+        if m:
+            id = m.group(1)
+        else:
+            m = re.match('.+JS5Prefs[0-9]+(.+)Ev$', sym)
+            if m:
+                id = f'javascript_options_{m.group(1)}'
+            else:
+                continue
 
-        id = m.group(1)
-        cpp_symbols[id] = sym
+        if id not in cpp_symbols:
+            cpp_symbols[id] = set()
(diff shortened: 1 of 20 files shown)


arai-a:
Added support for JS::Prefs (not yet reflected to the server)

pref-js-menu

there's unnecessary menu item ("StaticPrefs prefs StaticPrefs ..."). that could be removed later

scripts/staticprefs-analyze.py
```diff
@@ -0,0 +1,197 @@
...
+                    'syntax': 'def',
+                    'pretty': pretty,
+                    'sym': sym,
+                })
+
+                records.append({
+                    'loc': loc,
+                    'target': 1,
+                    'kind': 'def',
+                    'pretty': pretty,
```
asutherland, line 158: I think we want just the `name` here (without the "StaticPrefs " prefix) so that the identifiers table can find the synthetic `PREFS_` symbol.  That should improve the default search results a lot.

(still reviewing this great patch but I thought I should mention this sooner since I expect it will be the only feedback I have ;)

asutherland (approved):
Yet another fantastic quality-of-life improvement!  Thank you!

I very much appreciate the effort made to create a mini test corpus.

PR #834, opened 2024-11-04
Bug 1453506 - Forward type-dependent function calls to template call locations.

nicolas-guichard:
This provides a way to make type-dependent statements appear as-if they were located at the template call site.
    
During the analysis of the template, this first registers all type-dependent constructor and function call locations in ForwardedTemplateLocations (we don't have more information at this point).

Then when analyzing the template instantiations, we link the actual type-resolved statements to the template instantiation.
    
Then when emitting the analysis data for a template call, we also (recursively) emit all the analysis data for the forwarded items at the call site.


nicolas-guichard:
I've pushed this to [Try](https://treeherder.mozilla.org/jobs?repo=try&revision=37d07f2be16cd72ff129c820c4fb399d52e2471b) then will run a test indexer run on config1 when it passes.

nicolas-guichard:
I think we will want to filter some results out. For instance in the test I provide, opening the context menu on a `MakeUnique` call not only offers a go to entry for the `StructUsedInTypeDependentNew` constructor, but also for the intermediate `MakeUniqueWithIndex` template. That would be solved by only forwarding leaves.

Maybe it should also consider the data path and only forward the symbol use only if it can be returned?
```
template <typename T>
void MakeAndUseButDontReturnUnique() {
  const auto _ = std::unique_ptr<T>{new T()};
}

MakeAndUseUnique<Struct>(); //< appears as a Struct::Struct() call
```

asutherland:
Woo!  Very exciting!

For your MakeUnique test case, I agree it seems desirable to avoid emitting `MakeUniqueWithIndex` if it's straightforward.  But the extra context menu entry isn't a big deal; the major problem right now is searchfox being blind to MakeRefPtr so I'm fine with iterating.

For your hypothetical `MakeAndUseButDontReturnUnique` would the heuristic be that we require the forwarded values to be type compatible with the return value of the outermost template function (and where obviously nothing will match a void return)?  (It seems like anything more clever would tend to require data not available to the AST rep without getting into clang static analyzer stuff or MLIR/ClangIR/etc.)

I think it probably makes sense to err on the side of showing more of the nightmares hiding within the template invocation and then we can address what we find out from that for specific use cases.  [MozPromise.h](https://searchfox.org/mozilla-central/rev/387f3edbef37d31b2e91fb0812c74b54729e86ff/xpcom/threads/MozPromise.h) might be an interesting example where we'd have a bunch of interesting dependent new expressions.

@theres-waldo I expect you have more informed thoughts than I...

nicolas-guichard:
This new version skips over intermediary templates and doesn't end up in an infinite loop with recursive templates. The Try run passed and a config1 indexer run is currently running to be deployed on kdab.searchfox.org.

I haven't touched the `MakeAndUseButDontReturnUnique` case yet. It seems I could indeed guard the `ForwardingTemplates.insert(Template, Expression)` calls to check that the `Template` return type actually depends on the type of `Expression`.

nicolas-guichard:
This is up on kdab.searchfox.org, a few examples:
- [gfxContext built via MakeUnique](https://kdab.searchfox.org/mozilla-central/search?q=symbol:_ZN10gfxContextC1EPN7mozilla3gfx10DrawTargetERKNS1_10PointTypedINS1_12UnknownUnitsEfEE&redirect=false)
- [nsDisplayOwnLayer via MakeDisplayItemWithIndex](https://kdab.searchfox.org/mozilla-central/search?q=symbol:_ZN7mozilla17nsDisplayOwnLayerC1EPNS_20nsDisplayListBuilderEP8nsIFramePNS_13nsDisplayListEPKNS_18ActiveScrolledRootENS_22nsDisplayOwnLayerFlagsERKNS_6layers13ScrollbarDataEbb&redirect=false)
- [nsDisplaySolidColor via MakeDisplayItem via MakeDisplayItemWithIndex](https://kdab.searchfox.org/mozilla-central/search?q=symbol:_ZN7mozilla19nsDisplaySolidColorC1EPNS_20nsDisplayListBuilderEP8nsIFrameRK6nsRectjb&redirect=false)

In MozPromise.h InvokeAsync does foward a couple of type-dependent `new`s to its call site:
- MozPromise::Private (which is actually returned)
- its ProxyRunnableType and StaticString (that's more or less the MakeAndUseButDontReturnUnique case)

The template member functions such as MozPromize::Then don't forward any `new`s yet.

asutherland:
> This is up on kdab.searchfox.org, a few examples:

Awesome!

> The template member functions such as MozPromize::Then don't forward any `new`s yet.

To be clear, I'm not actually looking to see all the `new`s forwarded, MozPromise is just the first pile of templates that came to mind that seemed to use `new` that I thought might be interesting to see what does happen.

theres-waldo:
Sorry I didn't get a chance to look at this sooner. Generally speaking this looks great, and seems to address the reported use cases pretty effectively.

A couple of details that occur to me, that might be food for thought for future enhancements.

1. The current implementation sweeps up constructor calls that templates might make which are "implementation details", as in the following example:

    ```c++
    struct A { A(); };
    struct B { B(); };
    void DoOtherThings() {}
    template <typename T> void Helper1() {
      // create a T for local use
      auto local = new T;
      delete local;
      DoOtherThings();
    }
    template <typename T, typename U> void Foo() {
      Helper1<T>();
      Helper1<U>();
      DoOtherThings();
    }
    int main() {
      Foo<A, B>();  // links to A::A() and B::B()
    }
    ```

    This is probably not a big deal, but in case we want an easy-ish trick to narrow things down a bit: clangd has a heuristic where it tries to classify function templates as being [likely forwarding functions](https://searchfox.org/llvm/rev/e8c07f7458285c6fb2eddff5b7914519de10474d/clang-tools-extra/clangd/Preamble.cpp#169-188) based on their signature. Filtering what's eligible to be a key into `ForwardingTemplates` to instantiations of likely-forwarding-functions would exclude cases like the above which including the `MakeUnique`-like things that we care about.

    (Clangd doesn't currently have the equivalent of the feature being implemented here; it uses its "likely forwarding function" heuristic for a different purpose (related to inlay hints). But if it were to implement this feature (for which there is an [open issue](https://github.com/clangd/clangd/issues/716)), it would likely make use of the same heuristic.)

 2. The analysis does not see through member functions, for example:

    ```c++
    struct C { C(); };
    template <typename T>
    struct Factory {
      T *construct() {
        return new T;
      }
    };
    template <typename T>
    T* make_using_factory() {
      return Factory<T>().construct();
    }
    void foo() {
      make_using_factory<C>();  //  does not reference C::C()
    }
    ```

    I expect the reason for this is that the form of the `Factory<T>().construct()` expression is `MemberExpr` rather than `DeclRefExpr`; adding similar handling for `MemberExpr` would likely handle this.

theres-waldo:
>     2. The analysis does not see through member functions, for example:
>        ```c++
>        struct C { C(); };
>        template <typename T>
>        struct Factory {
>          T *construct() {
>            return new T;
>          }
>        };
>        template <typename T>
>        T* make_using_factory() {
>          return Factory<T>().construct();
>        }
>        void foo() {
>          make_using_factory<C>();  //  does not reference C::C()
>        }
>        ```

A more compelling use case for this is the "emplace" pattern:

```c++
template <typename T>
struct maybe {
  char storage[sizeof(T)];

  void emplace() {
    new (storage) T();
  }
};

struct A { A(); };

void foo() {
  maybe<A> m;
  m.emplace();  // would be nice if this linked to A::A()
}
```

PR #835, opened 2024-11-15
Bug 1931631- C++ indexer should treat constructor member initializer list field uses as having a contextsym of the constructor

asutherland:
(see specific commit messages; commit 1 adds a check that captures our currently wrong logic and commit 2 then lets us see the effects of the fix.)


arai-a (approved):
Looks good!

PR #837, opened 2024-11-18
Bug 1833552 - Annotate heuristic source records from C++ template analysis

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=1833552

This annotates source analysis records to indicate whether they are concrete results or originate from `HeuristicResolver`.

WIP because:
- this conflicts with https://github.com/mozsearch/mozsearch/pull/843 on some .snap file, and GitHub doesn't support stacked PRs


asutherland:
I think my comment at https://github.com/mozsearch/mozsearch/pull/628#issuecomment-1548575643 (which was the PR discussion that gave rise to this issue) is probably relevant here.  While my proposals frequently tend to tilt towards over-engineered, I think the situation with our JS/TS indexing[1] continues to suggest that we'll want a confidence mechanism for JS/TS symbols and it would be desirable for the approach taken here to also support that situation.

nicolas-guichard:
Is that closer to what you had in mind? The main change is that I replaced the `heuristic` boolean with a `confidence` enum. For now it only has 2 variants, but it should be extensible to your TS/JS use-case.

static/js/context-menu.js
```diff
@@ -513,6 +520,7 @@ var ContextMenu = new (class ContextMenu extends ContextMenuBase {
               href: `/${tree}/source/${jumpref.jumps.idl}`,
               icon: "export-alt",
               section: "jumps",
+              confidence: confidence,
```
asutherland, line 523: nit: like in rust [you can just do "confidence" here rather than "confidence: confidence](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#property_definitions) here and below.  (I think if we adopt "prettier" for formatting it probably would do this automatically too?)

asutherland (approved):
> Is that closer to what you had in mind? The main change is that I replaced the heuristic boolean with a confidence enum. For now it only has 2 variants, but it should be extensible to your TS/JS use-case.

Yeah, this is exactly what I had in mind.

Edit: here is a screenshot of what this looks like on https://searchfox.org/mozsearch-tests/source/bug1781178.cpp#30
image


I think this works sufficiently well for initial landing, especially since the only things that come to mind right now would potentially be much jarring to people's muscle memory.  Specifically, seemingly worse options are:
- Add some additional icon that conveys the confidence for the entry.  This seems like it would make the menu much busier and could have the problem that the icon is not self-explanatory and slapping a tooltip on the icons produces a bizarre UI experience where we have a tooltip floating over our context menu.
- Vary the single icon we currently used based on concrete/heuristic.  This also seems likely to be non-obvious and add visual noise that might not be helpful.
- Directly integrate the confidence into the text of the menu item.  Like "Go to def of IDL C++ getter Foo" would be "Go to heuristic def of IDL C++ getter foo".  Putting "heuristic" next to the symbol proper in the bolded text zone seems like it would make it harder to scan and require expanding "heuristic" to "heuristically discovered".
- Add labeled sections to the menu that are visually distinct from the menu items.  This seems most self-explanatory to me and potentially less bad in terms of visual hierarchy since [preattentive processing](https://www.csc2.ncsu.edu/faculty/healey/PP/) can be used to help rapidly identify those sections as navigation affordances, but there's a real potential that, at least as I mock up below, have too much weight with their high visual contrast.  Note again that this is just a mock-up example where I chose a bad thing to click on for the mock-up; IDL symbols will always be concrete:

image

In any event, I might propose that we plan to land this in the short-term and then have a longer-running discussion about the right long-term approach.  @smaug----'s UX canary input would be, in particular, appreciated and I'll drop a link to this issue in the chat.mozilla.org channel.

Landing-wise, note that we are in soft freeze for mozilla-central so we should probably wait until Monday when the freeze is lifted to land any of the m-c changes, but it's fine to land mozsearch changes before then.

clang-plugin/MozsearchIndexer.cpp
```diff
@@ -2501,7 +2512,7 @@ class IndexConsumer : public ASTConsumer,
     if (SyntaxKind) {
       std::string Mangled = getMangledName(CurMangleContext, ND);
       visitIdentifier("use", SyntaxKind, getQualifiedName(ND), Loc, Mangled,
-                      MaybeType, getContext(SpellingLoc));
+                      MaybeType, getContext(SpellingLoc), Heuristic);
```
theres-waldo, line 2515: nit: the comment above this function can be updated from "These should be distinguished from concrete results ... once bug 1833552 is fixed" to "These are distinguished from concrete results ..."

asutherland:
We had some discussion in the searchfox channel on matrix, but I think we probably still want to land this as-is with the context menu items colored since:
- It's hard to discuss the hypotheticals, especially just from our test-cases.
- The current coloring approach will definitely stand out and when people show up to ask "why is this context menu item colored?" it provides an opportunity to discuss very specific use-cases in context.
- The "go to def" becoming "go to a def" proposed so far is great in terms of being subtle and not messing with muscle memory that subtlety might undercut the benefits of being able to express we're dealing with heuristics.
  - For example, should we also be doing something like emitting a context menu row "look for heuristic misses" that generates some kind of fallback regexp search or something.

PR #838, opened 2024-11-18
Bug 1453506 - Forward constructor calls from template member functions too.

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=1453506

This extends https://github.com/mozsearch/mozsearch/pull/834 by covering the emplace use-case as described by @theres-waldo in https://github.com/mozsearch/mozsearch/pull/834#issuecomment-2480838505.


asutherland (approved):
Woo!

Thanks for the nicely factored stack; this is definitely much easier to review at the individual patch level because of the refactoring.

Restating building on prior restating comment in https://github.com/mozsearch/mozsearch/pull/834#pullrequestreview-2422919019:
- `VisitForwardedStatements` is extracted out of the logic in `VisitDeclRefExpr`.  This is conceptually the "bucket brigade" function that propagates `ForwardingTemplates` up the call stack.
- `VisitForwardedStatements`  also learns to understand `CallExpr`s which lets us propagate upwards for function/method calls that are template invocations in conjunction with the change to VistCallExpr to call it under those conditions.  So the emplace method call which does a dependent new operation gets bubbled up to the method-invocation call-site.

theres-waldo:
> This extends #834 by covering the emplace use-case as described by @theres-waldo in [#834 (comment)](https://github.com/mozsearch/mozsearch/pull/834#issuecomment-2480838505).

Thanks for making this follow-up enhancement!

Is this change available on a dev branch of searchfox to play around with?

nicolas-guichard:
I've just launched an indexer run re-using [that Try run](https://treeherder.mozilla.org/jobs?repo=try&revision=9d3f23bda6b8d44766e3bb8336876da8f7fb0c14). Check back on kdab.searchfox.org in a couple of hours.

nicolas-guichard:
Note this is already up on searchfox.org for other repos, just not m-c. See for instance https://searchfox.org/wubkat/search?q=symbol:_ZN21GrGLBackendFormatDataC1Ej&redirect=false.

theres-waldo:
Thanks! I had a look, and it seems like `mozilla::Maybe::emplace()` is managing to do something that foils the analysis, unlike the toy `maybe` implementation in https://github.com/mozsearch/mozsearch/pull/834#issuecomment-2480838505 or WebKit's `SkAnySubclass::emplace()`. I filed bug 1933325 as a follow-up to investigate that further; I'll try to get a reduced code example posted there.

PR #839, opened 2024-11-22
Fix formatting (rustfmt & clang-format) and clippy lints

nicolas-guichard:
I like using reformat on save in Kate so I don't have to worry about styling, but had to disable it every time I switched from other projects to mozsearch. Some formatting was very off, like merge_files in merger.rs which used 2+4n spaces for indent.

This applies rustfmt in tools, clang-format 19 (with LLVM style) in clang-plugin and also fixes a couple of Clippy lints to make the rust-analyzer output less noisy.

I know this won't hold long without CI enforcement though.

### DIFF
--- a/clang-plugin/.clang-format
+++ b/clang-plugin/.clang-format
@@ -1,0 +1,1 @@
+BasedOnStyle: LLVM
--- a/clang-plugin/BindingOperations.cpp
+++ b/clang-plugin/BindingOperations.cpp
@@ -20,23 +20,24 @@
 
 #ifdef __cpp_lib_optional
 #include <optional>
-template<typename T> using optional = std::optional<T>;
+template <typename T> using optional = std::optional<T>;
 #else
 #include <llvm/ADT/Optional.h>
-template<typename T> using optional = clang::Optional<T>;
+template <typename T> using optional = clang::Optional<T>;
 #endif
 
 using namespace clang;
 
 namespace {
 
-template<typename InputIt>
-bool hasReverseQualifiedName(InputIt first, InputIt last, const NamedDecl &tag)
-{
+template <typename InputIt>
+bool hasReverseQualifiedName(InputIt first, InputIt last,
+                             const NamedDecl &tag) {
   const NamedDecl *currentDecl = &tag;
   InputIt currentName;
   for (currentName = first; currentName != last; currentName++) {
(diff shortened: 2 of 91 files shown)


arai-a (approved):
Looks good!
Thank you!

emilio (approved):
Yeah, +1 to this.

asutherland (approved):
I've added :emilio and :arai as reviewers to have them sign-off on this too and for meta-discussion.

This is great, thank you!  I regularly find myself in situations where I'm afraid of running the formatter from inside VS code because I'm worried about messing up the commit with unrelated changes which potentially worsens the problem.  (I realize there's almost certainly a way to try and just format my outgoing changes, but I don't.)

> I know this won't hold long without CI enforcement though.

I'm definitely on board with getting some kind of CI going.  It [sounds like GitHub actions may be free in our case?](https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions).  The [Mozilla wiki guidance on actions is mainly about security guidelines it seems](https://wiki.mozilla.org/GitHub#GitHub_Actions).

Limiting the CI scope initially to just formatting/linting seems appropriate, as I think the docker provisioning runtime and/or storage if we're caching it might be trickier especially if we're not using docker cloud or something like that.  Taskcluster is definitely an option for running `make build-test-repo` and having automatically built/updated docker images, but it feels like more complexity than we'd want for this.

I think it could also be fine to add IDE configs to the tree for kate/VS code/etc. to help automatically enforce correct format-on-save (by default).

nicolas-guichard:
> I think it could also be fine to add IDE configs to the tree for kate/VS code/etc. to help automatically enforce correct format-on-save (by default).

We extensively use https://pre-commit.com/ to enforce that at commit time. One great thing about it is that the same config can be reused CI-side. We use https://pre-commit.ci/ for Pull Requests to our GitHub-hosted projects.
See for instance [.pre-commit-config.yaml in KDDockWidgets](https://github.com/KDAB/KDDockWidgets/blob/main/.pre-commit-config.yaml) which calls things like clang-format, cmake-format and shellcheck.

> It [sounds like GitHub actions may be free in our case?](https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions).

I think it is yes.

asutherland:
pre-commit looks nice and being able to run the checks both locally on CI seems great.

PR #840, opened 2024-11-22
Bug 1932922 - Rename per_target_function test to avoid c\ onflict with uppercase file 

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1932922

This renames the lowercase `per_target_function` file to `per_target_function_lower`, in order to avoid the conflict with the uppercase `PER_TARGET_FUNCTION` file.

and also updates the test snapshot, from `make review-test-repo`


asutherland (approved):
I guess this is something we could add a pre-commit hook for too! ;)

PR #842, opened 2024-11-25
Nix Flake update

nicolas-guichard:
I've updated my Nix shell along the way but since I'm still the only user (I think?), I never actually pushed it.

### DIFF
--- a/run-docker.sh
+++ b/run-docker.sh
@@ -25,7 +25,16 @@
 # you use; `activate` is for bash but there are different suffixed files for
 # other shell types.
 CONTAINER_NAME=${SEARCHFOX_DOCKER_CONTAINER_NAME:-searchfox}
-IMAGE_NAME=${SEARCHFOX_DOCKER_IMAGE_NAME:-searchfox}
+# Fully qualify the image name. This avoids docker/podman trying to fetch the
+# image from a remote.
+# See https://docs.docker.com/reference/cli/docker/image/tag/#description:
+# - If you don't specify a hostname, the command uses Docker's public registry
+#   at registry-1.docker.io by default.
+# And https://docs.podman.io/en/stable/markdown/podman-run.1.html#image:
+# - The reference can include a path to a specific registry; if it does not,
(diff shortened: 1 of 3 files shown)


run-docker.sh
```diff
@@ -25,7 +25,7 @@
 # you use; `activate` is for bash but there are different suffixed files for
 # other shell types.
 CONTAINER_NAME=${SEARCHFOX_DOCKER_CONTAINER_NAME:-searchfox}
-IMAGE_NAME=${SEARCHFOX_DOCKER_IMAGE_NAME:-searchfox}
+IMAGE_NAME=localhost/${SEARCHFOX_DOCKER_IMAGE_NAME:-searchfox}
```
asutherland, line 28: It would be good to add a comment here explaining what's going on or linking to docs somewhere that explain what's going on.

Restating:
- I understand maybe podman adds this prefix to images it builds which potentially makes it necessary?
- This seems to be harmless for my non-podman use-case based on local testing, presumably because I guess docker understands the pre-slash piece to potentially be a reference to a remote repository (and this explicitly is not remote)

asutherland (approved):
I think you may still be the only nix user; I'm unclear on the podman situation though.  Please do add the requested comment for that because it's not immediately obvious and my understanding is based on brief random googling.

run-docker.sh
nicolas-guichard: > linking to docs somewhere that explain what's going on

I've added that now: linking to both docker and podman docs which describe that syntax and why it avoids querying a remote.

arai-a:
This seems to break the workflow on ubuntu.
The `./run-docker.sh` shows the following error, and reverting the change solves.
I'll file a bug.

```
Unable to find image 'localhost/searchfox:latest' locally
docker: Error response from daemon: Get "http://localhost/v2/": dial tcp 127.0.0.1:80: connect: connection refused.
See 'docker run --help'.
```

arai-a:
filed https://bugzilla.mozilla.org/show_bug.cgi?id=1943509

PR #843, opened 2024-11-25
Initial pre-commit config

nicolas-guichard:
Following the discussion in https://github.com/mozsearch/mozsearch/pull/839, I've put together a .pre-commit-config.yaml with clang-format, rustfmt, clippy and a few extra hooks and made sure that it passes.

To use it:
- install pre-commit (using your preferred package manager)
- run `pre-commit install` to install the hook

When committing those checks will be run on the files that are touched by the commit. You can also run the checks manually using `pre-commit run --all-files`.

It does add a bit of overhead when committing, especially the cargo-clippy hook from a cold cache:
```
➜  mozsearch git:(wip-stack) rm -rf **/target

➜  mozsearch git:(wip-stack) time pre-commit run --all-files                                                                                                                                   
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check for added large files..............................................Passed
check for case conflicts.................................................Passed
check xml................................................................Passed
check yaml...............................................................Passed
check json...............................................................Passed
check for broken symlinks............................(no files to check)Skipped
detect destroyed symlinks................................................Passed
check that executables have shebangs.....................................Passed
clang-format.............................................................Passed
cargo fmt................................................................Passed
cargo clippy.............................................................Passed
pre-commit run --all-files  407,98s user 59,95s system 1163% cpu 40,222 total

➜  mozsearch git:(wip-stack) time pre-commit run --all-files
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check for added large files..............................................Passed
check for case conflicts.................................................Passed
check xml................................................................Passed
check yaml...............................................................Passed
check json...............................................................Passed
check for broken symlinks............................(no files to check)Skipped
detect destroyed symlinks................................................Passed
check that executables have shebangs.....................................Passed
clang-format.............................................................Passed
cargo fmt................................................................Passed
cargo clippy.............................................................Passed
pre-commit run --all-files  7,97s user 1,81s system 327% cpu 2,982 total
```

PR #844, opened 2024-11-25
Bug 1833555 - Mark OverloadExpr locations to be visited in instantiations.

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=1833555

Now that the context menu allows for differentiating between heuristic and concrete template instantiations, mark OverloadExpr location to be visited in AnalyzeDependent mode to have concrete results.

In the added WithOverloads::Overloaded test case, the float and bool overloads are only found heuristically (and are thus marked with cppTemplateHeuristic) while the int overload is also found by visiting concrete template instantiations.

### DIFF
--- a/tests/tests/checks/snapshots/analysis/cpp/bug1781178.cpp/check_glob@WithOverloads_Overloaded.snap
+++ b/tests/tests/checks/snapshots/analysis/cpp/bug1781178.cpp/check_glob@WithOverloads_Overloaded.snap
@@ -1,0 +1,129 @@
+---
+source: tests/test_check_insta.rs
+expression: "&json_results"
+snapshot_kind: text
+---
+[
+  {
+    "loc": "00076:14-24",
+    "source": 1,
+    "syntax": "decl,function",
+    "type": "void (int)",
+    "pretty": "function WithOverloads::Overloaded",
+    "sym": "_ZN13WithOverloads10OverloadedEi"
+  },
+  {
+    "loc": "00076:14-24",
+    "target": 1,
+    "kind": "decl",
+    "pretty": "WithOverloads::Overloaded",
+    "sym": "_ZN13WithOverloads10OverloadedEi",
+    "context": "WithOverloads",
+    "contextsym": "T_WithOverloads",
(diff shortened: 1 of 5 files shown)


nicolas-guichard:
Marked as WIP as https://github.com/mozsearch/mozsearch/pull/843 then https://github.com/mozsearch/mozsearch/pull/837 should be merged first, and GitHub still doesn't support dependent PRs AFAICS.

tests/tests/checks/snapshots/analysis/cpp/bug1781178.cpp/check_glob@WithOverloads_Overloaded.snap
```diff
@@ -0,0 +1,129 @@
...
+    ]
+  },
+  {
+    "loc": "00082:4-14",
+    "source": 1,
+    "syntax": "use,function",
+    "type": "void (int)",
+    "pretty": "function WithOverloads::Overloaded",
+    "sym": "_ZN13WithOverloads10OverloadedEi"
+  },
```
asutherland, line 101: ah, okay, so we see it both as a heuristic confidence and with an (implicit) concrete confidence

asutherland (approved):
Looks good!  (I only reviewed the leaf commit labeled with the bug number here because of the github limitations on stacks which I also understand to still be a thing.)

tests/tests/checks/snapshots/analysis/cpp/bug1781178.cpp/check_glob@WithOverloads_Overloaded.snap
nicolas-guichard: Here yes, because this file doesn't go through merger.rs. (Only the files listed in $MULTI_CPP_FILES in `tests/tests/build` go through merger.rs.)
If it did, we would only keep the second entry (the one without cppTemplateHeuristic). format.rs also has some merge logic and only one entry is emitted in `data-symbols` and `data-confidences`.

PR #846, opened 2024-11-26
Bug 1933325 - Link calls to emplace-like functions to the constructors they invoke.

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=1933325

This adds the missing bits to be able to get std::vector::emplace_back<T> appear as constructor calls for T.

### DIFF
--- a/clang-plugin/MozsearchIndexer.cpp
+++ b/clang-plugin/MozsearchIndexer.cpp
@@ -1181,6 +1181,15 @@ public:
     AutoTemplateContext Atc(this);
     Super::TraverseClassTemplateDecl(D);
 
+    // Gather dependent locations from partial specializations too
+    SmallVector<ClassTemplatePartialSpecializationDecl *> PS;
+    D->getPartialSpecializations(PS);
+    for (auto *Spec : PS) {
+      for (auto *Rd : Spec->redecls()) {
+        TraverseDecl(Rd);
+      }
+    }
+
     if (!Atc.needsAnalysis()) {
       return true;
     }
--- a/clang-plugin/MozsearchIndexer.cpp
+++ b/clang-plugin/MozsearchIndexer.cpp
@@ -2144,11 +2144,6 @@ public:
   }
 
   bool VisitCXXConstructExpr(const CXXConstructExpr *E) {
-    SourceLocation Loc = E->getBeginLoc();
-    if (!isInterestingLocation(Loc)) {
-      return true;
-    }
-
     // If we are in a template and find a Stmt that was registed in
     // ForwardedTemplateLocations, convert the location to an actual Stmt* in
     // ForwardingTemplates
@@ -2159,6 +2154,11 @@ public:
             {getCurrentFunctionTemplateInstantiation(), E});
         return true;
       }
+    }
+
+    SourceLocation Loc = E->getBeginLoc();
+    if (!isInterestingLocation(Loc)) {
+      return true;
     }
 
     return VisitCXXConstructExpr(E, Loc);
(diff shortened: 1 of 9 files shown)


clang-plugin/MozsearchIndexer.cpp
```diff
@@ -1138,6 +1138,33 @@ class IndexConsumer : public ASTConsumer,
...
+  // a definition, to handle cases such as:
+  // ```
+  // template <typename T>
+  // struct Maybe {
+  //   template <typename... Args>
+  //   T *emplace(Args&&... args) {
+  //     defined here
+  //   }
+  // }
+  // ```
```
asutherland, line 1155: Should this example code include the forward declaration from https://bugzilla.mozilla.org/show_bug.cgi?id=1933325#c1 potentially with a comment on the forward like: "This is a canonical declaration, but not a definition."?

(Also this is a really great comment, thanks!)

asutherland (approved):
Thank you!  And many thanks to @theres-waldo for the reduced repro!

clang-plugin/MozsearchIndexer.cpp
nicolas-guichard: While getting the std::vector::emplace_back working, I think I got a better understanding of what is going on here, and settled on isThisDeclarationADefinition for class templates and keeping isCanonicalDecl for function templates. I've updated and completed the comment in that regard.
I'll admit I'm still not 100% confident in my understanding of this though :sweat_smile:

asutherland: Ha, yeah, when it comes to templates I very much appreciate that we have a test corpus with precisely targeted coverage because everything about templates tends to test the limits of my mental stack and my reviews are very much of a "this code does not contradict my mental model of what's going on" where my mental model frequently regresses to "templates are spicy macros".

asutherland (approved):
This is awesome, thank you!  Random thoughts:
- Really great comment on TraverseClassTemplateDecl, thank you!  (Also, the individual contextualizing commit messages are deeply appreciated too.)
- The consolidation of logic into `VisitCallExpr` nicely aids in understanding by reducing code complexity, plus I imagine it will also make it easier to confidently evolve the code for additional expression types in the future because there's a clear, centralized place to do it!

nicolas-guichard:
I've put this in a Try run and it's now live on kdab.searchfox.org. Both examples in the bug report now work as expected:
- this Maybe::emplace call: https://kdab.searchfox.org/mozilla-central/source/dom/animation/AnimationEffect.cpp#83
- this std::vector::emplace_back call: https://kdab.searchfox.org/mozilla-central/source/gfx/layers/apz/src/APZCTreeManager.cpp#552

theres-waldo:
Both examples in the bug report now work as expected:
> 
>    * this Maybe::emplace call: https://kdab.searchfox.org/mozilla-central/source/dom/animation/AnimationEffect.cpp#83
> 
>    * this std::vector::emplace_back call: https://kdab.searchfox.org/mozilla-central/source/gfx/layers/apz/src/APZCTreeManager.cpp#552

This is great to see, thanks very much Nicolas!

jrmuizel:
Yes, this is great.

This is a search that I did today: https://searchfox.org/mozilla-central/search?q=symbol:_ZN7mozilla19AutoProfileRunnableC1ER12nsTSubstringIcE&redirect=false

And this is the result I wanted: https://kdab.searchfox.org/mozilla-central/search?q=symbol:_ZN7mozilla19AutoProfileRunnableC1ER12nsTSubstringIcE&redirect=false

PR #847, opened 2024-11-27
Bug 1671640 - Add the ability to suppress warnings from the warnings email

asutherland:
Run our log file through a `grep --invert-match` against the new file `infrastructure/aws/warning-suppression.patterns` which we then pipe into the grep that looks for warnings.  We don't provide line numbers in the email so the omitted lines shouldn't matter, although we could always just replace those lines with blank lines if we started doing that.

To help with testing this, I made a few minor other changes:
- If the email address is "test" we just dump the warnings to stdout.
- I made the log path into an argument rather than the hardcoded path.

So I was able to download the config5 indexer log and run the following and observe that with an empty suppression file we would print out warnings, and with the new suppression in place that's in this commit we early return from the script before both the "test" mode logic and the email sending:
```
infrastructure/aws/send-warning-email.py infrastructure/aws/warning-suppression.patterns FOO test index-log
```

### DIFF
--- a/infrastructure/aws/index.sh
+++ b/infrastructure/aws/index.sh
@@ -123,7 +123,7 @@ release* )
     ;;
 esac
 
-$AWS_ROOT/send-warning-email.py "$CHANNEL/$BRANCH" "$DEST_EMAIL"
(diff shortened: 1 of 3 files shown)


arai-a (approved):
Excellent!

Also good to know that `grep` command receives patterns as a file :)

PR #849, opened 2024-11-28
Bug 1933960 - Fallback to concrete confidence when data-confidences is missing.

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=1933960

When implementing bug 1833552 I assumed we would always have the data-confidences attribute next to the data-symbols attribute. This isn't the case for diagrams at the moment.

When data-confidences is missing, assume everything is concrete.


arai-a (approved):
Looks good!

nicolas-guichard:
Should I apply that patch on the release1 server? Or just wait for the next indexer run?

PR #850, opened 2024-11-29
Bug 1934174 - Ignore devtools test file that can contain experimental syntax.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1934174

### DIFF
--- a/scripts/js-analyze.js
+++ b/scripts/js-analyze.js
@@ -280,7 +280,13 @@ const FILENAME_INTERVENTIONS = [
     includes_list: ["/wubkat/"],
     severity: "INFO",
     prepend: "Wubkat failsafe: "
-  }
+  },
+  {
+    includes_list: ["devtools/client/debugger/test/mochitest/examples/inline-preview.js",
+                    "devtools/client/debugger/test/mochitest/examples/preview.js"],
+    severity: "INFO",
+    prepend: "It may contain experimental syntax: ",
+  },
 ];
 
 function logError(msg)


asutherland (approved):
Thanks!

Restating:
- We [update the jsshell we use for JS parsing at the start of every indexer run](https://searchfox.org/mozsearch/rev/619ed6102d33f8c2dd8f15650cb1aed6175998a7/infrastructure/indexer-update.sh#46-55) so this isn't a case of something we provision being out-of-date, it's a pref, [javascript.options.experimental.explicit_resource_management](https://searchfox.org/mozilla-central/rev/d4235354139ae0da3b15b89943b140b510a74f2f/modules/libpref/init/StaticPrefList.yaml#8285-8292) which is also ifdef'ed but [seems defined](https://searchfox.org/mozilla-central/search?q=ENABLE_EXPLICIT_RESOURCE_MANAGEMENT&redirect=false) and that seems to happen because [the default moz.configure value is gated on being a nightly build](https://searchfox.org/mozilla-central/rev/d4235354139ae0da3b15b89943b140b510a74f2f/js/src/shell/js.cpp#12978-12979).
  - Meta: We really probably should try and do something to surface definitions from pathkinds that aren't "core code" since they can frequently be off-screen.  Maybe we could add a sort of table-of-contents in the navigation panel which would use LocalStorage to remember whether it's open or closed on search pages.  So we'd have an ul/li summary of the sections of the results, so the defs would be under "Generated code (12)", "Definitions (12)" where we'd use the number of lines.
  - The pref can be enabled at runtime in the jsshell [via a command line flag](https://searchfox.org/mozilla-central/rev/d4235354139ae0da3b15b89943b140b510a74f2f/js/src/shell/js.cpp#12978-12979).  It doesn't look like there's a flag that turns on all the prefs, but there is a (generated) [FOR_EACH_JS_PREF](https://searchfox.org/mozilla-central/source/__GENERATED__/__win64__/js/public/PrefsGenerated.h#128-182) which does include that pref which [powers --list-prefs](https://searchfox.org/mozilla-central/rev/d4235354139ae0da3b15b89943b140b510a74f2f/js/src/shell/js.cpp#13016-13019) so we could potentially do a regexp thing on there to turn on all the prefs that way if we thought that the backwards compatible nature of JS meant that this would only harmlessly change how parses behave.  But since I see a bunch of WASM prefs in there and we use WASM, maybe that's not a good idea / worth it.
- Bug https://bugzilla.mozilla.org/show_bug.cgi?id=1933139 recently introduced these new test files.

arai-a:
Enabling all feature is somewhat risky, given we don't test that configuration and in some case the experimental code isn't stable enough.
So, I think it's better just rejecting those experimental syntax, or at least enable with manual allowlist.

Having "core code" vs "test code" would be nice indeed.

PR #852, opened 2024-12-03
Bug 1934697 - Replace Raw and Log links with GitHub links.

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=1934697

generate_commit_info in format.rs already built link to GitHub for a specific commit.

This re-uses the same data source to build URLs to the GitHub log and GitHub raw content for the navigation pane.

The Raw link is replaced with the GitHub link when available _instead_ of the Mercurial link.

The Mercurial “Log” link is renamed to “Mercurial Log”.

A “Git Log” link is addded to point to GitHub when available.

### DIFF
--- a/static/js/panel.js
+++ b/static/js/panel.js
@@ -13,8 +13,6 @@ var Panel = new (class Panel {
 
     this.permalinkNode = this.findItem("Permalink");
     this.unpermalinkNode = this.findItem("Remove the Permalink");
-    this.logNode = this.findItem("Log");
-    this.rawNode = this.findItem("Raw");
 
     this.selectedSymbol = null;
 
--- a/static/js/panel.js
+++ b/static/js/panel.js
@@ -198,6 +198,10 @@ var Panel = new (class Panel {
     return this.panel.querySelector(`.item[title="${title}"]`);
   }
 
+  findAccel(key) {
+    return this.panel.querySelector(`.item[data-accel="${key}"]`);
+  }
+
   maybeHandleAccelerator(event) {
     if (!this.acceleratorsEnabled) {
       return;
@@ -213,22 +217,22 @@ var Panel = new (class Panel {
       switch (event.key) {
         case "y":
         case "Y":
-          return this.permalinkNode;
+          return this.findAccel('Y');
         case "l":
         case "L":
-          return this.logNode;
+          return this.findAccel('L');
         case "r":
         case "R":
-          return this.rawNode;
+          return this.findAccel('R');
         case "f":
         case "F":
-          return this.markdown.filename.node;
+          return this.findAccel('F');
         case "s":
         case "S":
-          return this.markdown.symbol.node;
+          return this.findAccel('S');
         case "c":
         case "C":
-          return this.markdown.block.node;
+          return this.findAccel('C');
       }
     })();
 
--- a/tools/src/bin/output-file.rs
+++ b/tools/src/bin/output-file.rs
@@ -335,22 +335,56 @@ fn main() {
                     accel_key: None,
                     copyable: false,
                 });
-                if let Some(ref hg_root) = tree_config.paths.hg_root {
+
+                let gh_log_link = tree_config
+                    .paths
+                    .github_repo
+                    .as_ref()
+                    .map(|gh_root| format!("{}/commits/HEAD/{}", gh_root, encoded_path));
+                let hg_log_link = tree_config
+                    .paths
+                    .hg_root
+                    .as_ref()
+                    .map(|hg_root| format!("{}/log/tip/{}", hg_root, encoded_path));
+                if let Some(link) = gh_log_link {
                     vcs_panel_items.push(PanelItem {
-                        title: "Log".to_owned(),
-                        link: format!("{}/log/tip/{}", hg_root, encoded_path),
+                        title: "Git log".to_owned(),
+                        link,
+                        update_link_lineno: "",
+                        accel_key: hg_log_link.is_none().then_some('L'),
+                        copyable: true,
+                    });
+                }
+                if let Some(link) = hg_log_link {
+                    vcs_panel_items.push(PanelItem {
+                        title: "Mercurial log".to_owned(),
+                        link,
                         update_link_lineno: "",
                         accel_key: Some('L'),
                         copyable: true,
                     });
+                }
+
+                let gh_raw_link = tree_config
+                    .paths
+                    .github_repo
+                    .as_ref()
+                    .map(|gh_root| format!("{}/raw/HEAD/{}", gh_root, encoded_path));
+                let hg_raw_link = tree_config
+                    .paths
+                    .hg_root
+                    .as_ref()
+                    .map(|hg_root| format!("{}/raw-file/tip/{}", hg_root, encoded_path));
+                if let Some(link) = gh_raw_link.or(hg_raw_link) {
                     vcs_panel_items.push(PanelItem {
                         title: "Raw".to_owned(),
-                        link: format!("{}/raw-file/tip/{}", hg_root, encoded_path),
+                        link,
                         update_link_lineno: "",
                         accel_key: Some('R'),
                         copyable: true,
                     });
                 }
+
                 if tree_config.paths.git_blame_path.is_some() {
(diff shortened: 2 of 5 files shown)


nicolas-guichard:
I think that should fix https://bugzilla.mozilla.org/show_bug.cgi?id=1847297 as well.

I've only tested this locally (using the amended Searchfox config), do you want me to run a test indexer run?

asutherland:
> I've only tested this locally (using the amended Searchfox config), do you want me to run a test indexer run?

Yes, please.  We'll want to quickly see how people feel about the change in the channel before landing it.

tools/src/bin/output-file.rs
```diff
@@ -335,22 +335,58 @@ fn main() {
                     accel_key: None,
                     copyable: false,
                 });
-                if let Some(ref hg_root) = tree_config.paths.hg_root {
+
+                let gh_log_link = tree_config
+                    .paths
+                    .github_repo
+                    .as_ref()
+                    .map(|gh_root| format!("{}/commits/{}/{}", gh_root, oid, encoded_path));
```
nicolas-guichard, line 343: Here I've pointed to the current commit (`oid`), but actually to match the existing Mercurial behavior that should point to `HEAD`.

tools/src/bin/output-file.rs
```diff
@@ -335,22 +335,58 @@ fn main() {
...
+                        accel_key: (!has_gh_log_link).then_some('L'),
+                        copyable: true,
+                    });
+                }
+
+                let gh_raw_link = tree_config
+                    .paths
+                    .github_repo
+                    .as_ref()
+                    .map(|gh_root| format!("{}/raw/{}/{}", gh_root, oid, encoded_path));
```
nicolas-guichard, line 374: Same here: should this be `HEAD` instead of  the displayed commit `oid`?

nicolas-guichard (commented):
This is up on [kdab.searchfox.org](https://kdab.searchfox.org/mozilla-central/source/README.txt) for testing.

tools/src/bin/output-file.rs
asutherland: Yeah, I think we probably want the branch tip.  So I guess we potentially need to pull the branch name that's checked out from the "git" dir?  We could of course put it in the config, but that's sorta currently implicit from the disk state...

tools/src/bin/output-file.rs
asutherland: Yeah, I think so.  And it's probably more polite load-wise.

tools/src/bin/output-file.rs
nicolas-guichard: The Mercurial version just uses `tip`, I think that's the Git equivalent for `HEAD`?

tools/src/bin/output-file.rs
```diff
@@ -335,22 +335,58 @@ fn main() {
...
+                    .paths
+                    .hg_root
+                    .as_ref()
+                    .map(|hg_root| format!("{}/log/tip/{}", hg_root, encoded_path));
+                if let Some(link) = hg_log_link {
+                    vcs_panel_items.push(PanelItem {
+                        title: "Mercurial log".to_owned(),
+                        link,
+                        update_link_lineno: "",
+                        accel_key: (!has_gh_log_link).then_some('L'),
```
asutherland, line 365: So unfortunately I think we need to make the logic a little weirder than this.  As discussed on chat.mozilla.org, I think if we have both "Git log" and "Mercurial log" we want the hotkey to go to Mercurial, even though we want to visually display "Git log" first to try and bias people who don't have a preference to use Git log[1].

1: Specifically:
- We expect git log to be faster based on hg.mozilla.org's load problems right now.
- I'm hearing that we want to discourage people from using hg.mozilla.org as much as possible
- Because we are moving from hg to git for m-c, we do expect "Git log" to become the only non-searchfox option available in the medium term.  (That said, as I noted in the channel, ideally I'll be adding a new searchfox-based option that maybe could end up being our new default.)

tools/src/bin/output-file.rs
```diff
@@ -335,22 +335,58 @@ fn main() {
                     accel_key: None,
                     copyable: false,
                 });
-                if let Some(ref hg_root) = tree_config.paths.hg_root {
+
+                let gh_log_link = tree_config
+                    .paths
+                    .github_repo
+                    .as_ref()
+                    .map(|gh_root| format!("{}/commits/HEAD/{}", gh_root, encoded_path));
```
asutherland, line 343: Have you seen any docs that say this will end up defaulting to the default branch?  I do see at https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#list-commits--parameters that the Query parameters docs for `sha` says:
> SHA or branch to start listing commits from. Default: the repository’s default branch (usually main).

So maybe this is good enough?

(It does seem hopefully unlikely that if someone pushed a branch other than the default branch that github would be favor that commit over the default branch head.)

asutherland (changes requested):
Unfortunately since https://github.com/mozsearch/mozsearch/blob/master/docs/webtest.md uses the "tests" repo I'm not sure we can easily add a webtest for this, but we should do another round on kdab.searchfox.org before landing with the "L" for "mercuail log if present, otherwise git log", I think.  Thanks.

tools/src/bin/output-file.rs
nicolas-guichard: I've not found any docs for that no, but it works, at least currently. I have no idea what promises GitHub makes for its user-facing URLs.
Note that the documentation you linked to is for api.github.com, which is properly documented and versioned, but not a good place to redirect humans to :wink:

tools/src/bin/output-file.rs
nicolas-guichard: The latest version should now give the priority to the Mercurial log for the L hotkey.

I've also removed the hardcoding of hotkeys to nodes in panel.js. In the first version of this PR, I assumed changing `PanelItem::accel` was enough and broke the hotkey. Now changing `PanelItem::accel` should actually be enough.

nicolas-guichard:
I've launched another just-mc indexer run:
```
i-0117e1489f25b045b running indexer-secure 0:05:43 ['branch: replace-raw-log-links-with-github', 'cfile: just-mc.json', 'channel: kdab', 'crepo: https://github.com/nicolas-guichard/mozsearch-mozilla', 'indexer: 2024-12-03 22:53:42.629205', 'mrepo: https://github.com/nicolas-guichard/mozsearch']
```

I'll hopefully be AFK when it finishes.

static/js/panel.js
```diff
@@ -13,7 +13,7 @@ var Panel = new (class Panel {
 
     this.permalinkNode = this.findItem("Permalink");
     this.unpermalinkNode = this.findItem("Remove the Permalink");
-    this.logNode = this.findItem("Log");
+    this.logNode = this.findItem("Git log");
     this.rawNode = this.findItem("Raw");
```
asutherland, line 17: I think `logNode` and `rawNode` can be removed because the only uses were in the accelerator case below.  (But the (un)permalinkNode and `markdown` cases all seem to still be actively used.)

tools/src/bin/output-file.rs
asutherland: Ha, yeah, I just meant to hand-wave that presumably the UI is operating on top of the same logic as the API so ideally it's stable.

asutherland (changes requested):
Thanks, this looks/works good but, as someone who has burned by forgetting our "diff" UI several times in my own patches, I just checked that ([example](https://kdab.searchfox.org/mozilla-central/diff/1000a339cf4672b0f6da29e4116cb26ccbe08255/caps/BasePrincipal.h#14)) and it looks like we need to adjust [the "Log" logic in format_diff too](https://searchfox.org/mozsearch/rev/1434c976ebfc0901b56f9a2fde07d4df16136a0f/tools/src/format.rs#1240-1248).

nicolas-guichard:
I found the other two by grepping for `"Raw"`, but the diff view only had `“Log”` indeed.

asutherland (approved):
Thank you!

PR #855, opened 2024-12-05
Update insta snapshots following the addition of data-accel attributes

nicolas-guichard:
When testing the last few iterations of #852 I only used the searchfox repo for testing and didn't notice I needed to update snapshots for the test repo.


asutherland (approved):
Thanks!

I think it probably makes sense for us to pursue setting up a basic github actions CI for searchfox before looking at Kotlin v2; I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1935484 on this.

PR #856, opened 2024-12-16
Bug 1937445 - increase web-server AMI to 16G, latch mise version

asutherland:
The primary intent here is to boost the size of the web-server AMI to 16G to address running out of disk space during the self-update of the web-server at the start of its spin-up.

However, when running `./build-docker.sh` locally initially to validate that the provisioning process was likely to complete on the server, I encountered a provisioning failure during the installation of `mise` which is the tool we use to create a sort-of `venv` for node.  mise seems to be very much under development at this time and we definitely don't need any of the new enhancements, so I picked the version that was released on the last successful provisioning run of 2024-Oct-11.

As added to the comment, I'm not sure we really need to be using mise. I believe my original rationale was that nvm was deprecated and mise (then rtx) was a rust-based tool that seemed to to support the configs for the (non-rust) asdf tool and it seemed to work.  Given that we've updated to Ubuntu 24.04 and our node.js needs are so limited (and the upstream scip tooling so rarely updated) it might be possible to just use the default Ubuntu node.js packages.  (I believe I had looked at getting .deb files for 22.04 and that had eneded up undesirable.)

### DIFF
--- a/infrastructure/aws/web-server-provision.sh
+++ b/infrastructure/aws/web-server-provision.sh
@@ -49,7 +49,8 @@ popd
 
 date
 
-# Size up our root partition to 12G
+# Size up our root partition to 16G.  This is up from 12G when we were hitting
+# the limit for the self-update process during the rust build.
 #
 # To this end we need to know the volume id in order to issue an EBS resizing
 # command.  Note that the select constraint here is intended more as a check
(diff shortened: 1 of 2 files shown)


asutherland:
The provisioning went okay on the servers after I got the local build-docker to complete.  Note that on this laptop I did have a problem with `run-docker.sh` with the `localhost/` recent change to run-docker.sh but I'm ignoring that for now.

PR #857, opened 2025-01-09
Bug 1935484 - GitHub Actions CI

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=1935484

This implements a simple GitHub Actions workflow for pull requests and pushes to master. It runs the existing provisioning scripts then runs the equivalent of `make build-test-repo` and `make webtest`.
It takes ±50 minutes to run, see a successful run here: https://github.com/nicolas-guichard/mozsearch/actions/runs/12690265272.

I had to:
- reduce the output size of the provisioning scripts, see all the “space savings” commits
- change css-analyze a bit to avoid leaking the index location in a symbol
- make sure the provisioning scripts clone the repo and branch that is being tested and not the master branch of https://github.com/mozsearch/mozsearch

The last point that bothers me is that the graphviz-dependent tests have different numerical results than what I get locally in my Podman test env, see 46352f3a4b83c48df4a2f4822cb2495611cae6e8. Any ideas why and what to do about it?

### DIFF
--- a/infrastructure/common-provision-pre.sh
+++ b/infrastructure/common-provision-pre.sh
@@ -134,6 +134,7 @@ if [ ! -d $HOME/.cargo ]; then
   rustup install nightly
   rustup default nightly
   rustup uninstall stable
+  rustup component add rust-analyzer
 fi
 
 # install ripgrep so we can stop experiencing grep pain / footguns
@@ -220,10 +221,3 @@ fi
 SCIP_VERSION=v0.5.0
 curl -L https://github.com/sourcegraph/scip/releases/download/$SCIP_VERSION/scip-linux-amd64.tar.gz | tar xzf - scip
 sudo ln -fs $(pwd)/scip /usr/local/bin/scip
-
-# Install rust-analyzer
-RUST_ANALYZER_VERSION=nightly
-rm -rf rust-analyzer rust-analyzer-linux-x64.vsix
-wget https://github.com/rust-lang/rust-analyzer/releases/download/$RUST_ANALYZER_VERSION/rust-analyzer-linux-x64.vsix
-unzip -o -d rust-analyzer rust-analyzer-linux-x64.vsix
-sudo ln -fs $(pwd)/rust-analyzer/extension/server/rust-analyzer /usr/local/bin/rust-analyzer
--- a/infrastructure/common-provision-pre.sh
+++ b/infrastructure/common-provision-pre.sh
@@ -175,29 +175,6 @@ if [ ! -d livegrep ]; then
   rm -rf livegrep
 fi
 
-# Install AWS scripts and command-line tool.
-#
-# In order to get the AWS CLI v2 the current options[1] are to use snap or do
-# the curl + shell dance.  We don't have snap support installed by default and are
-# currently intentionally avoiding adding snaps, so we choose curl + shell.
-#
-# 1: https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html
-#
-# awscli can get credentials via `Ec2InstanceMetadata` which will give it the
-# authorities of the role assigned to the image it's running in.  Look for the
-# `IamInstanceProfile` definition in `trigger_indexer.py` and similar.
-#
-# (We check if the install directory already exists because on AWS we install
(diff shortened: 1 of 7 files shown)


asutherland:
> The last point that bothers me is that the graphviz-dependent tests have different numerical results than what I get locally in my Podman test env, see [46352f3](https://github.com/mozsearch/mozsearch/commit/46352f3a4b83c48df4a2f4822cb2495611cae6e8). Any ideas why and what to do about it?

Assuming all cases are indeed using the same graphviz binary release, my guess would be that it's a font problem.  Right now [we set the font in our diagrams [just as "Courier New"](https://searchfox.org/mozsearch/rev/b83a0b650851f4f46750884def30bc35eaa6aad9/tools/src/cmd_pipeline/symbol_graph.rs#384,729,1510,1523) which I arrived on after only a small amount of experimentation.

If I run `fc-match "Courier New"` in my docker VM I get `NimbusMonoPS-Regular.otf: "Nimbus Mono PS" "Regular"` but if I run it outside on my local machine I get `LiberationMono-Regular.ttf: "Liberation Mono" "Regular"` which suggests there's a good chance for problems.

There was an [interesting recent discussion on the graphviz forum about the spacing problems that can result from inconsistent font usage between dot and the browser](https://forum.graphviz.org/t/svgs-are-broken-in-subtle-ways-spacing-issues/2645) which indeed we, or at least I see but haven't addressed other than [making sure we have noto installed for our emoji](https://searchfox.org/mozsearch/rev/b83a0b650851f4f46750884def30bc35eaa6aad9/infrastructure/common-provision-pre.sh#67-68).  In the github issue linked in https://forum.graphviz.org/t/svgs-are-broken-in-subtle-ways-spacing-issues/2645/19 for a doxygen diagram use case, the user found success in using the Roboto font along with web fonts so the browser also uses the exact same font.  We could potentially address our issues by standardizing on Roboto or another license-appropriate monospace font ([Roboto seems to be Apache 2.0](https://github.com/googlefonts/roboto-2)).  I guess we might also need/want to explicitly specify Noto as well; [Wikipedia says it's free](https://en.wikipedia.org/wiki/Noto_fonts).

If we're time-limited, or maybe even if we're not, it's fine to punt on dealing with this and we can just remove the checks that snapshot SVG output.  The "dot" format and "json" format snapshots already provide most of the regression coverage we want and the "svg" snapshots have been more subject to churn than I think I'd initially expected we'd see.  I think I mainly added them because I could and it provided an easy way for someone to inspect what the SVG output looked like from the tree, but whenever I've regressed something, it's the "dot" format snapshot that has been what I pay attention to.

infrastructure/common-provision-pre.sh
```diff
@@ -175,29 +175,6 @@ if [ ! -d livegrep ]; then
...
-# the CLI earlier out of necessity.)
-if [ ! -d awscliv2-install ]; then
-  mkdir -p awscliv2-install
-  pushd awscliv2-install
-  curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
-  unzip awscliv2.zip
-  sudo ./aws/install
-  popd
-fi
-
```
asutherland: Restating: we already install these scripts [in aws/indexer-provision.sh](https://searchfox.org/mozsearch/rev/b83a0b650851f4f46750884def30bc35eaa6aad9/infrastructure/aws/indexer-provision.sh#31-48) and [in aws/web-server-provision.sh](https://searchfox.org/mozsearch/rev/b83a0b650851f4f46750884def30bc35eaa6aad9/infrastructure/aws/web-server-provision.sh#31-48) because we needed it for `aws e2 modify-volume` to increase the size of our AMIs during provisioning.  (As the comment with the dir check here affirms.)

asutherland (approved):
This is super awesome!  Thank you!

I believe we have consensus in https://bugzilla.mozilla.org/show_bug.cgi?id=1935484 that we want to go all-in on nix for sanity/reproducibility purposes, so this r=asuth is for this as well as landing the nix variant as well, however that happens.  I'll comment a few quick notes for the nix stack in the bug though.

PR #859, opened 2025-01-23
Bug 1945159 - Nix-based devenv, CI and deployment

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=1945159

The current script-based provisioning is, to quote @asutherland, “is a pile of hacks combined with foot-guns”. Instead this aims at better defining packages for dependencies and mozsearch components.

This reuses dependencies from nixpkgs when available and introduces new third-party packages for scip-{java,typescript,python}, livegrep, wasm-snip and the webidl and xpidl parsers.

On top of that the mozsearch tooling is split in 4 packages:
- the clang plugin
- the Rust-based tools
- the wasm-css-analyzer
- everything else in mozsearch-scripts

Those are used by the build-index and serve-index scripts (which are just wrappers around the existing indexer-{setup,run}.sh and web-server-{setup,run}.sh scripts).
build-index takes a config in and outputs an index, serve-index takes an index and configures/kills/starts codesearch, router.py, pipeline-server, web-server and nginx.

The tests, searchfox and webtests indexes are built using build-index, and can be served using serve-index. Contrary to my first draft, those are now defined as (pure) Nix packages because their are self-contained. Nix runs checks as part of building a package, so `nix build '.#tests'` actually builds _and checks_ the tests config. `nix build '.#tests.unchecked'` is available to avoid running those checks. Searchfox and webtests use this Git repository itself so require a clever trick to provide the full repository.

The other indexes are built using https://github.com/mozsearch/mozsearch-mozilla/pull/276 and impure/volatile sources such as “the latest successful build on Wekbit's CI” or “the TRYPUSH_REV environment variable”.
In short, to build the graphviz index to `~/graphviz-index` and serve it locally from `~/graphviz-srv`, one can, with only Nix installed:
```
nix run --accept-flake-config 'github:nicolas-guichard/mozsearch-mozilla/nixified#index-just-graphviz' ~/graphviz-index
nix run --accept-flake-config 'github:nicolas-guichard/mozsearch-mozilla/nixified#serve-config' just-graphviz ~/graphviz-index ~/graphviz-srv
```


asutherland:
> I'm marking this as draft because I'm still missing:
> 
>     * `make trypush`, although that seems to be covered by `TRYPUSH_REV=… make build-mozilla-repo`?

Yeah, despite it [being documented](https://searchfox.org/mozsearch/rev/b83a0b650851f4f46750884def30bc35eaa6aad9/docs/manual-indexing.md#133-144) I'd never realized `make trypush` existed or I forgot about it and so have only ever used build-mozilla-repo.  We can ignore and/or remove this.  (The meta-issue of how to only build a subset of the trees in a config file is a separate issue.)

nicolas-guichard:
This should now work _in parallel_ to the existing script-based provisioning.

I switched this from using https://github.com/DeterminateSystems/magic-nix-cache-action to https://github.com/cachix/cachix-action with my experimental https://nicolas-guichard.cachix.org Cachix-hosted cache, which makes the cache available outside of GitHub Actions (ie re-usable by devs and the AWS indexer).

Building and serving the Graphviz index as described above should _not_ rebuild livegrep but just download a cached build.

If you agree with using Cachix (their 5 GiB free tier is plenty for the things we want to cache), I should probably create a mozsearch.cachix.org cache, provide you with admin access and switch to that prior to merging this.
Cachix can be configured to use our own signing keys if we don't trust them to store the binaries securely, but the default (and the current state of that PR) is to let them handle the signing of binaries.
Alternatively we could self-host a cache.

asutherland:
> Alternatively we could self-host a cache.

We probably want/need to self-host due to a combination of minimizing moving parts/auth complications, visibility into the moving parts for anyone ramping up on searchfox in the future (especially any SRE help), and company policy stuff.  Obviously it will be more expensive than free, but that's not the primary factor for searchfox.

nicolas-guichard:
Alright, pushing the cache to an S3 bucket instead of Cachix is simple (I can just replace https://github.com/cachix/cachix-action with its fork https://github.com/nix-ocaml/nix-s3-action).

FYI they both boil down to:
1. note down what's in the store before building
2. build
3. `cachix push` or `nix copy --to s3://...` what's new in the store

Should I go ahead and create an S3 bucket for this?
One thing I have to figure out is how to expire old entries (Cachix provides LRU cache eviction, but AFAIK plain S3 doesn't provide that nor access logs so it would have to be replaced with something time-based, which is good enough in our case because the indexers shouldn't fall behind what GitHub Actions builds).

asutherland:
Yes, please create an S3 bucket.  Expiration isn't a major concern; S3 says we're currently storing 300 GB there and it's costing us $10/month.

I think the primary concern is just that we should make sure that if we are giving the github CI anything resembling AWS credentials that they are limited to just writes to that bucket.  I don't know if S3 can let us directly issue some kind of one-off key for that or if that implies a full IAM user/entity thing.

asutherland (approved):
I know you marked this WIP, presumably because there's some level of inherent bitrot (like us going from jdk 21 to 17, other version/hash drift) due to delay in conclusive review.  I've finished up doing a "that makes sense to me" level of review (with having done more in depth digging previously) and unless @arai-a feels otherwise, I think it probably makes sense to de-bitrot this and land this when you feel like you have some continuous-ish time so that you don't get bitrotted again and can also do any fixups that might shake out after landing/as others update their local dev environments/etc.

To assist the local dev environment thing too, it would be great to propagate the howto commit message explanations for running stuff to make sure that we have either a (temporarily) forked version of the README.md or an updated version/etc.

nicolas-guichard:
I'm breaking this up. We can benefit from caching the livegrep and mozsearch builds right away in https://github.com/mozsearch/mozsearch/pull/1093.

PR #860, opened 2025-01-25
Bug 1942757 - Use the actual list of files in the objdir for indexing

arai-a:
This is for https://bugzilla.mozilla.org/show_bug.cgi?id=1943521 and https://bugzilla.mozilla.org/show_bug.cgi?id=1942757

This patch stack does:
  * Update the test snapshot to reflect the new rust SCIP symbol syntax  (This is a part of the stack because otherwise the test fails on the latest rust nightly)
  * Stop generating the C++ `.o` files in the objdir in the test repo, so that those files are not shown in the output
  * Use the list of files in objdir, for the files reflected in __GENERATED__ files.  The newly added files are listed in the bug 1942757 comment 6

Available in https://arai.searchfox.org/

### DIFF
--- a/scripts/find-objdir-files.sh
+++ b/scripts/find-objdir-files.sh
@@ -4,9 +4,24 @@ set -x # Show commands
 set -eu # Errors/undefined vars are fatal
 set -o pipefail # Check all commands in a pipeline
 
-pushd ${INDEX_ROOT}/analysis/__GENERATED__
-find . -type f | sed -e 's#^./#__GENERATED__/#' > ${INDEX_ROOT}/objdir-files
-find . -mindepth 1 -type d | sed -e 's#^./#__GENERATED__/#' > ${INDEX_ROOT}/objdir-dirs
+# Some repositories don't have objdir.
+mkdir -p ${INDEX_ROOT}/objdir
+
+pushd ${INDEX_ROOT}/objdir
+# All text-like files in objdir are going to be reflected to the output.
+#
+# NOTE: exclude some known binary filename extensions to reduce the cost of
+#       the file type detection.
+set +o pipefail # grep can fail
+find . -type f -not -regex "\.(o|out|so|a|so\..*|scip)$" -exec file --mime {} + \
+    | grep -v 'charset=binary' \
+    | cut -d ":" -f 1 \
+    | sed -e 's#^./#__GENERATED__/#' \
+    > ${INDEX_ROOT}/objdir-files
+set -o pipefail
+find . -mindepth 1 -type d \
+    | sed -e 's#^./#__GENERATED__/#' \
+    > ${INDEX_ROOT}/objdir-dirs
 popd
 
 # This is shuffled for the benefit of the "parallel" invocation in output.sh so that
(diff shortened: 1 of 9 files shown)


scripts/find-objdir-files.sh
```diff
@@ -4,7 +4,7 @@ set -x # Show commands
 set -eu # Errors/undefined vars are fatal
 set -o pipefail # Check all commands in a pipeline
 
-pushd ${INDEX_ROOT}/analysis/__GENERATED__
+pushd ${INDEX_ROOT}/objdir
 find . -type f | sed -e 's#^./#__GENERATED__/#' > ${INDEX_ROOT}/objdir-files
```
asutherland, line 8: I wonder if we should try and filter out files that seem to be binary files here, either by filtering on the file extension and/or using "file".  Like if we did something like `find . -type f -exec file --mime {} + | grep -E 'text|ascii' | cut -d: -f1 | sed -e 's#^./#__GENERATED__/#'` (but that might be missing json?  Maybe we should invert on binary?)

I think in some of our other trees where we do builds on the indexer we potentially will see a bunch of object files otherwise:
- [wubkat](https://searchfox.org/mozsearch-mozilla/source/wubkat/build) where we build on the indexer
- [mozsearch](https://searchfox.org/mozsearch-mozilla/source/mozsearch/build) where we build the clang indexer in the objdir (but everything else should be fine)
  - Ideally [mozsearch-tests](https://searchfox.org/mozsearch-mozilla/source/mozsearch-tests/build) is fine because it just tries to use the test repo config.
- [llvm](https://searchfox.org/mozsearch-mozilla/source/llvm/build) where we run the build on the indexer.
- [graphviz](https://searchfox.org/mozsearch-mozilla/source/graphviz/build) where we run the builds on the indexer.

Having looked briefly at wubkat, checking the contents of the file seemed helpful since it does write out a bunch of dependency files that have extensions that sorta look binary-y.  (Although I guess they're like ".do" and the "d" is significant.

From a performance perspective, it seems that [build-codesearch.py](https://searchfox.org/mozsearch/rev/867abebdb0bc074cbc2876f6c6fafa347e4c5e55/scripts/build-codesearch.py#10-27) will actually end up copying over all the files from objdir-files, so this filtering in this way will help reduce the waste on the copying and also help bootstrap caching a little bit.  (I'm not quite sure why we're copying things over, although obviously it does help with ignoring things we don't want to index.)

asutherland (approved):
This is fine to land but we should expect to see binary files if we don't do the filtering thing, but that's fine since everything but m-c and its trees are tier 2 or lower.

scripts/find-objdir-files.sh
arai-a: Good point! and thank you for checking other repositories.

Indeed it would be better adding a filter instead of modifying the test repository not to generate .o files.

I'll modify the script to exclude binary files.
If all binary files have certain filename extension such as .o and .so, we can apply the filename-based filtering.
I'll check those repositories.

arai-a: Applied the mime-type-based filtering, using "charset=binary", with extension-based pre-filtering,
and also reverted the modification to the test repo.

arai-a: just realized grep cannot be used inside the pipeline, given it hits pipefail if no file is found.
I'll fix that part

PR #861, opened 2025-01-26
Bug 1912952 - Show the class alignment/size in the field layout table.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1912952
depends on https://github.com/mozsearch/mozsearch/pull/860

This does:
  * Update the clang plugin, from https://bugzilla.mozilla.org/show_bug.cgi?id=1943761
  * Add the alignment and the size to the class layout table, after the class name (not base class), as a separate row, but without border

align-size


asutherland:
I presume the WIP is because of the stack of PRs and you'll try to land the other changes first.  r=asuth regardless of how things land.  (I have a response in progress to the others, they should be reviewed shortly.)

Thanks for making this great enhancement to an already great feature (thanks to your amazing prior work)!  I've been using the field layout a lot recently just to be able to quickly and easily understand all the fields that are on a class for pernosco and it's been super helpful.  In particular, the "Line" column and the fact that it's fully semantically hooked up is so great.

PR #863, opened 2025-01-26
CI: Fix cloning repo from pull request

nicolas-guichard:
GitHub merge repos are not available as an actual branch and can't be directly cloned with `git clone -b ... --depth=1`. Instead, perform the manual steps of creating the repo, adding the remote, fetching and switching.


asutherland (approved):
Thanks!

PR #865, opened 2025-01-27
Update update.sh call in trigger_common.py

nicolas-guichard:
In 54490508, I updated the indexer's update.sh arguments to match web-server's update.sh, but missed the call in trigger_common.py.


arai-a (approved):
Looks good! Thanks!

PR #866, opened 2025-01-27
Bug 1942757 - Part 2: Do not snapshot the entire generated files HTML

arai-a:
followup for https://bugzilla.mozilla.org/show_bug.cgi?id=1942757

Given the objdir can contain files that may differ depending on the environment, storing the snapshot in the test case can fail.
For example `.rustc_info.json` can differ.

Assuming the purpose of the test is just to ensure the `__GENEDATED__` directory listing works, I've modified the testcase to just check single item, which is a directory, so that the filesize or content doesn't affect the snapshot.


asutherland (approved):
> Assuming the purpose of the test is just to ensure the __GENEDATED__ directory listing works, I've modified the testcase to just check single item, which is a directory, so that the filesize or content doesn't affect the snapshot.

Yeah, I believe from https://github.com/mozsearch/mozsearch/pull/562 adding the whole directory was just an easy way to ensure we were listing all the files and to lock down any changes in formatting so we'd notice if they changed.

It does seem like if we were to experience a regression in directory listing generation, it would only be a question of whether we generate the listing at all, not if the lower level directory enumeration logic provided by the rust stdlib breaks[1].  So this nicely still verifies that without the churn and while providing a check on the HTML consistency.  It could probably make sense for us to change the root dir listing too since it's a bit noisy for no win.

Thanks!

1: The exception is part of the changes I'd made was for us to move towards supporting git directory listings.  That potentially could be slightly more brittle, but I think it likely would make sense for us to be testing that against some kind of test repo that we don't regenerate every time, and so the git hash would be stable.

PR #868, opened 2025-02-09
Bug 1627778 - Do not float the SVG preview.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1627778

This does:
  * Make the SVG preview not float, so that it's displayed before the source, and it doesn't break the layout of the source
  * Add more background options (explicit light and dark), given it now has more horizontal space in the control part, and the default background depends on the current theme (light/dark)

preview

available in https://arai.searchfox.org/mozilla-central/source/browser/branding/aurora/content/about-logo.svg

This doesn't have testcase because the preview is available only with hg repository.

### DIFF
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -1183,6 +1183,10 @@ input::placeholder {
   }
 }
 
+.svg-preview {
+  padding: 0 1rem;
+}
+
 .svg-preview > h4 {
   margin-bottom: 0;
 }
--- a/static/css/mozsearch.css
+++ b/static/css/mozsearch.css
@@ -1196,8 +1196,6 @@ input::placeholder {
 }
 
 .svg-preview > a > img {
-  float: left;
-  clear: both;
   min-height: 50px;
   max-height: 300px;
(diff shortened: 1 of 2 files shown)


asutherland (approved):
Thanks!

emilio (commented):
https://arai.searchfox.org/mozilla-central/source/browser/themes/shared/weather/flurries.svg is an example where the suggested changes would make a difference.

static/css/mozsearch.css
```diff
@@ -1204,11 +1206,7 @@ input::placeholder {
   margin-bottom: 1em;
 }
 
-#svg-preview-checkerboard {
-  margin-inline-start: 0;
-}
-
-#svg-preview-checkerboard:checked + label + a > img {
+#svg-preview-background-checkerboard:checked + label + input + label + input + label + a > img {
```
emilio, line 1209: Probably better to use `#svg-preview-background-checkerboard:checked ~ a > img` (and same for the others) as it's simpler and not meaningfully slower tbh. But your call

PR #869, opened 2025-02-11
Bug 1947618 - pipeline-server.rs should serve Vary header

asutherland:
I verified this by inspection, specifically looking at the response headers of a diagram via devtools locally.  Our automated test infra right now doesn't quite have the capability to add a header check like this.  It's definitely possible to add but it's also very unlikely that we would regress this specific endpoint again; the likely problem would be adding a new endpoint and most test infra we would add would probably not automatically cover that case.

### DIFF
--- a/tools/src/bin/pipeline-server.rs
+++ b/tools/src/bin/pipeline-server.rs
@@ -6,7 +6,7 @@ use std::{
 
 use axum::{
     extract::{Path, Query},
-    http::{HeaderMap, StatusCode},
+    http::{header, HeaderMap, StatusCode},
     response::{Html, IntoResponse, Response},
     routing::get,
     Extension, Json, Router,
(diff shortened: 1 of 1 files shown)


asutherland:
I'm interested to see what the CI says for this.  I had the pre-commit hooks and they were angry, but they were not angry about my changes to pipeline-server.rs, but instead to other untouched files (summary below which does not name the changes which was `src/file_format/doc_trees.rs` and `src/cmd_pipeline/cmd_format_symbols.rs` and `src/doc_trees_handler.rs` and `src/format.rs`:

```
cargo fmt...............................................................Failed                                                                                                                                        
- hook id: cargo-fmt                                                                                       
- files were modified by this hook                                                                                                                                                                                    
cargo clippy.............................................................Failed                            
- hook id: cargo-clippy                                                                                                                                                                                               
- exit code: 1                                                                                             
```

PR #870, opened 2025-02-15
Bug 1948451 - Ignore more import assertions errors.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1948451

This adds one more pattern for the error message around import assertions.


asutherland (approved):
Thanks!

PR #872, opened 2025-03-08
Bug 1949604 - Linkify moz-src:/// strings.

arai-a:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1949604

This does:
  * Linkify moz-src:/// URL in JS
  * Linkify moz-src:/// URL in CSS
  * Linkify moz-src:/// URL in C++
  * Linkify moz-src:/// URL in comment
  * Add more tests


arai-a:
clang-plugin is copied to https://phabricator.services.mozilla.com/D240771 as well.

arai-a:
Forgot to explain some more.
This doesn't use the mapping, but just generates the FILE_ symbol from the URL's path part.

asutherland (approved):
Nice!  Obviously the point of moz-src is to have a straightforward mapping, but it's very nice to see the benefits of that!

PR #873, opened 2025-04-09
Bug 1959328 - Bump git-cinnabar version.

arai-a:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1959328
  * Bump the git-cinnabar pinning version to 0.7.2
  * Let the script fetch and update to the new revision if necessary

My current understanding is that, after this script is merged, we run the provisioning command, for indexer and web-server (so, run 2 commands), and that solves the warning.  Is it correct?
https://github.com/mozsearch/mozsearch/blob/master/docs/aws.md#provisioning-and-cloud-init

### DIFF
--- a/infrastructure/common-provision-pre.sh
+++ b/infrastructure/common-provision-pre.sh
@@ -189,22 +189,29 @@ fi
 sudo apt-get install -y python3-boto3 python3-rich
 
 # Install git-cinnabar.
+
+# Need mercurial to prevent cinnabar from spewing warnings.
+sudo apt-get install -y mercurial
+
+# We started pinning in https://bugzilla.mozilla.org/show_bug.cgi?id=1779939
+# and it seems reasonable to stick to this for more deterministic provisioning.
+CINNABAR_REVISION=0.7.2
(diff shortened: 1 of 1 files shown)


asutherland (approved):
Thanks!

Yeah, you need to run those scripts and then some emails should ideally show up in ~30 minutes that say that the provisioning succeeded and there are new AMIs.

As noted in matrix, I think the else case probably shouldn't actually end up firing, but if it was handy for manually iterating, that's cool.  Since we have CI now it's very nice to know that the CI ran the provisioning script and it's happy.  (I did expand the run data and scrolled down so I could ctrl-f and verify it thinks it used 0.7.2.  But of course that's not an m-c run, but just knowing the provisioning scripts ran happily is the key thing.)

asutherland:
As an aside, sometimes I will also [update the base AMI we use](https://searchfox.org/mozsearch/rev/fa8075f8546dbab9618f8346013e6c6c36009801/infrastructure/aws/trigger-provision.py#142-148) but that's never essential, it just potentially makes the provisioning go faster because in theory apt has less to do during `apt update`.

PR #874, opened 2025-04-10
Bug 1959582 - Correct mozsearch-mozilla specific revision checkout for web-server

asutherland:
I'm also speculatively trying to leverage the ability to perform a shallow checkout of submodules as well to help the disk space usage for the CI runners.  It sounds like there might have been problems in very old verisons of git if the revision wasn't the branch head but that it's no longer a problem.  We also do try and stick to the tip of the branches, so that also should moot things.


asutherland:
The CI passed after I reverted my change that had us use the revision name to `FETCH_HEAD` since it turned out that the revision for CI was actually a ref / tag.  I'll land now and trigger the re-provisioning.

PR #875, opened 2025-04-10
Bug 1959722 - Ignore warnings about redirecting to hg-edge.mozilla.org

asutherland:
I tested this against the most recent config1 indexer run using `send-warning-email.py`'s test mechanism for when "test" is specified as the email address to use.  Pre-change, I see what got sent, post-change I see no warnings.


arai-a (approved):
Looks good!
Thank you :D

asutherland:
I believe the CI failed here because I deleted the branch here while the CI was still running so when it went to check out the branch again it was gone.

PR #876, opened 2025-04-17
Bug 1938622 - Update test gradle project to Kotlin 2.1.20.

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=1938622

This updates the test gradle project to use Kotlin 2.1.20 and our forked semanticdb-kotlinc.

Like my patch currently on review on Phabricator it's using JitPack for the time being, until we publish the package to Maven Central or maven.mozilla.org.

### DIFF
--- a/tests/tests/checks/snapshots/analysis/java/InlineObject.kt/check_glob@method__json.snap
+++ b/tests/tests/checks/snapshots/analysis/java/InlineObject.kt/check_glob@method__json.snap
@@ -1,41 +1,18 @@
 ---
 source: tests/test_check_insta.rs
 expression: "&json_results"
+input_file: inputs/analysis/java/InlineObject.kt/method__json
 ---
 [
   {
     "loc": "00010:21-33",
     "structured": 1,
-    "pretty": "someFunction",
-    "sym": "S_jvm_src/main/kotlin/sample/InlineObject.kt/#0",
+    "pretty": "sample::<anonymous object at 137>::someFunction",
+    "sym": "S_jvm_sample/<anonymous_object_at_137>#someFunction().",
     "type_pretty": null,
     "kind": "method",
     "subsystem": null,
-    "implKind": "impl",
-    "sizeBytes": null,
-    "alignmentBytes": null,
-    "ownVFPtrBytes": null,
-    "bindingSlots": [],
-    "ontologySlots": [],
-    "supers": [],
-    "methods": [],
-    "fields": [],
-    "overrides": [
-      {
-        "sym": "S_jvm_sample/Interface#someFunction()."
-      }
-    ],
-    "props": [],
-    "variants": []
-  },
-  {
-    "loc": "00016:21-33",
-    "structured": 1,
-    "pretty": "someFunction",
-    "sym": "S_jvm_src/main/kotlin/sample/InlineObject.kt/#2",
-    "type_pretty": null,
-    "kind": "method",
-    "subsystem": null,
+    "parentsym": "S_jvm_sample/<anonymous_object_at_137>#",
     "implKind": "impl",
     "sizeBytes": null,
     "alignmentBytes": null,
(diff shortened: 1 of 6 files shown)


nicolas-guichard:
CI is failing on `sudo apt install -y npm` because of a file conflict between `node-tap` and `containernetworking-plugins`…
> ```
> dpkg: error processing archive /tmp/apt-dpkg-install-NiwdpZ/395-node-tap_16.3.7+ds1+~cs50.9.19-4_all.deb (--unpack):
>  trying to overwrite '/usr/bin/tap', which is also in package containernetworking-plugins 1.6.2-ubuntu24.04u1
> ```

It passed on the source branch a few days ago: https://github.com/nicolas-guichard/mozsearch/actions/runs/14450666988/job/40522308390

tests/tests/checks/snapshots/analysis/java/InlineObject.kt/check_glob@method__json.snap
```diff
@@ -1,41 +1,18 @@
 ---
-source: tests/test_check_insta.rs
+source: src/bin/test-index.rs
```
asutherland, line 2: Is this a new thing from some other stack?

asutherland (approved):
These changes all seem fine but it definitely would be good for the CI to be happy.  Presumably this is hopefully a transient Debian/Ubuntu packaging glitch, as it seems like the alternatives system and related policies should be mooting the problem.

tests/tests/checks/snapshots/analysis/java/InlineObject.kt/check_glob@method__json.snap
nicolas-guichard: Oops yes this is from https://github.com/nicolas-guichard/mozsearch/commit/462cad0e061e5f889f8517bda2103487286b003f in my nixified branch.

nicolas-guichard:
The CI breakage comes from the https://packages.microsoft.com/ubuntu/24.04/prod noble repo, which has containernetworking-plugins 1.6.2-ubuntu24.04u1. (The default Ubuntu repos have 1.1.1+ds1-3ubuntu0.24.04.2 which do not conflict with node-tap.)

nicolas-guichard:
OK so the issue stems from our use of `apt-get --allow-change-held-packages dist-upgrade`, which installs the problematic `containernetworking-plugins` version. (As evidenced by [this passing job](https://github.com/nicolas-guichard/gh-actions-npm-breakage/actions/runs/14593011421/job/40932328941) vs [this failing job](https://github.com/nicolas-guichard/gh-actions-npm-breakage/actions/runs/14593061740/job/40932491230)).

Can we just remove the `--allow-change-held-packages` argument? What was the reason for adding it back in 308974895?

asutherland:
> OK so the issue stems from our use of `apt-get --allow-change-held-packages dist-upgrade`, which installs the problematic `containernetworking-plugins` version. (As evidenced by [this passing job](https://github.com/nicolas-guichard/gh-actions-npm-breakage/actions/runs/14593011421/job/40932328941) vs [this failing job](https://github.com/nicolas-guichard/gh-actions-npm-breakage/actions/runs/14593061740/job/40932491230)).

Thanks for figuring that out!

> Can we just remove the `--allow-change-held-packages` argument? What was the reason for adding it back in [3089748](https://github.com/mozsearch/mozsearch/commit/308974895c596b0de82bcd5d2d8c42c125e1fe96)?

This should probably be fine.  Based on my comments in https://github.com/mozsearch/mozsearch/pull/258 and https://bugzilla.mozilla.org/show_bug.cgi?id=1601451 I think I probably had just cargo-culted something from the internet[1] to make sure that 1) we did make sure our checkout was up-to-date[2] but 2) we didn't hang like in the grub config change prompting I experienced that I mention in the PR.  Especially when we were using vagrant on older hardware, the iteration cycles were fairly long so I probably erred on the side of using the largest hammer available.

1: I usually try and cite specific sources for things like this but don't appear to have done so in this case.

2: I think this was a combination of not wanting to regress our packages being up-to-date-ish via "unattended-upgrades" that I was removing, combined with related general concerns from [trying to attach gdb to a hung python process](https://bugzilla.mozilla.org/show_bug.cgi?id=1498207#c11) where if we're installing packages we want to try and avoid skew where half of the instance is stale and the other half is up-to-date because of an explicitly installed package at provisioning time.

PR #877, opened 2025-04-24
Remove --allow-change-held-packages from apt dist-upgrade

nicolas-guichard:
On GitHub CI, this leads to updating containernetworking-plugins, which would then conflict with node-tap for the file /usr/bin/tap.


asutherland (approved):
Thanks!

PR #880, opened 2025-06-15
Bug 1970807 - Bump release2 EBS size from 300G to 400G.

asutherland:
I'll see what our actual usage is after a successful run, but the rough math based on current usage (~241G excluding nginx-cache) and the size of mozilla-central on config1 (~65G) plus our allocation of 20G for nginx-cache means we expect to hit at least 326G, so ~74G spare is advisable.  (Also, there's the whole GB/GiB units issue.)


asutherland:
I'm re-triggering config2 now.  I'd manually terminated it today and yesterday which is why it hadn't errored out.

PR #881, opened 2025-06-16
Bug 1970807 - Update from m5d to m6id instances for indexer.

asutherland:
This is motivated by config2 hitting the limit of the 300G SSD for m5d.2xlarge. The choice of m6id is being made because it seems to both be faster and gives us a larger 474G SSD which immediately solves our sizing problem.  Also nice is that m6id.4xlarge will have a 950G SSD which means that we won't have to do any clever RAID stuff if we want even more local space than that.

Note that I will need to re-provision the lambda tarballs after this and I will do that.


asutherland:
I ran [the automated-ish lambda update steps](https://github.com/mozsearch/mozsearch/blob/master/docs/aws.md#automated-ish-updates), then launched the release2 lambda job from the lambda web UI and it successfully launched the expected m6id.2xlarge instance type and we're now in the setup phase.  I'll keep an eye on it.

PR #882, opened 2025-06-18
Bug 1970807 - Teach indexer log analyzer about staticprefs-analyze.sh

asutherland:


asutherland:
For the analysis at https://bugzilla.mozilla.org/show_bug.cgi?id=1970807#c7 on the performance changes of the new instances.

PR #883, opened 2025-07-29
Use Firefox/Geckodriver and and Python modules from env

nicolas-guichard:
This is another part of my nixified stack.

It avoids:
- Downloading Firefox/Geckodriver in webtest.sh. Instead webtest.sh expects Firefox and Geckodriver to be available in PATH, which is true on GitHub Actions, and now done in the provisioning step for docker. Firefox and Geckodriver are not used on AWS.
- Downloading webidl, xpidl and ply in idl-analyze.sh. Instead idl-analyze.sh expects them to be in PYTHONPATH and they are downloaded in the provisioning/update steps.

I also patched the livegrep build/install step so that we can `import livegrep` instead of `import src.proto`.

I tested this without re-provisioning on [kdab.searchfox.org](https://kdab.searchfox.org/graphviz/source/) with the Graphviz index. Once the AMI is re-provisioned, some leftover update and env setup steps marked with TODO can be removed.


asutherland (approved):
Thanks!

It's definitely much nicer to have the parser stuff from hg.mozilla.org pulled down during provisioning and update than as a weird side-effect!  At some point we'll want to stop referencing hg.mozilla.org in favor of the new firefox github repo, but since the hg server doesn't have capacity problems anymore and there's no point in changing things that could run into new edge cases, it makes sense to leave us using hg.mozilla.org in this PR.

PR #884, opened 2025-07-29
Update test gradle project to Kotlin 2.2.0

nicolas-guichard:
Mozilla-central updated to Kotlin 2.2.0 almost a month ago with [bug 1975907](https://bugzilla.mozilla.org/show_bug.cgi?id=1975907). This catches up.


asutherland (approved):
Thanks!

I also feel like I should add some kind of joke about us testing in production and only accepting things into the test repo once they've proven themselves there :wink: !

PR #885, opened 2025-07-30
Use mozsearch tools from PATH

nicolas-guichard:
This installs Mozsearch tools to .cargo/bin and updates the scripts to expect the tools to be in `$PATH`.

The mercurial `tip` currently points at `tags-unified` instead of `default`, which failed the previous GitHub Actions run [here](https://github.com/nicolas-guichard/mozsearch/actions/runs/16626148101/job/47043005611#step:4:2799).

Also requires https://github.com/mozsearch/mozsearch-mozilla/pull/281.

This is currently deployed on kdab.searchfox.org with the Graphviz index.

### DIFF
--- a/infrastructure/indexer-provision.sh
+++ b/infrastructure/indexer-provision.sh
@@ -53,12 +53,12 @@ rm -rf ~/.cache/coursier
 PYMODULES=$HOME/pymodules
 mkdir "${PYMODULES}"
 pushd "${PYMODULES}"
-wget "https://hg.mozilla.org/mozilla-central/raw-file/tip/xpcom/idl-parser/xpidl/xpidl.py"
-wget "https://hg.mozilla.org/mozilla-central/raw-file/tip/dom/bindings/parser/WebIDL.py"
+wget "https://github.com/mozilla-firefox/firefox/raw/refs/heads/main/xpcom/idl-parser/xpidl/xpidl.py"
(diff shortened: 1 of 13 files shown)


asutherland (approved):

asutherland:
Ah, so there's an iteration regression in the Makefile-based development loop since the [make rule that rebuilds the rust binaries builds them but does not install them onto the path](https://searchfox.org/mozsearch/rev/f82c282f7f4e42bfcfb1d16708e746b87d6ee4c7/Makefile#22-23)...
```
build-rust-tools:
	cd tools && cargo build --release --all-targets
```

...like this PR changed [web-server-update.sh to do](https://searchfox.org/mozsearch/rev/f82c282f7f4e42bfcfb1d16708e746b87d6ee4c7/infrastructure/web-server-update.sh#29-30):
```
pushd mozsearch/tools
CARGO_INCREMENTAL=false cargo install --path . --verbose
```

Changing the build line to `cd tools && CARGO_INCREMENTAL=1 cargo install --path .` addresses that problem, although even with dropping `--verbose`, I can't tell if it's doing more re-work or if it's just saying more stuff than it used to (or if it's just due to an updated toolchain or if I'm just not realizing it always said the same stuff).

I'll put the fix in my stack since I'm going to try and land the all the firefox-* transition stuff this weekend and using my actually changed code fixed a bunch of mysteries! ;)

PR #886, opened 2025-07-31
Fix WebIDL parsing of async iterables

nicolas-guichard:
The WebIDL definition of async iterables just changed to async_iterable in https://github.com/mozilla-firefox/firefox/commit/92ad1b37c03d428fa91605612d353021f689f025, making the new WebIDL.py incompatible with older WebIDL files.

This:
1. fixes the test when using the latest WebIDL.py
2. restores the feature that uses the in-tree WebIDL.py when available, for non-nightly trees

A test index run of config2 is currently running with this.

### DIFF
--- a/scripts/idl-analyze.sh
+++ b/scripts/idl-analyze.sh
@@ -23,6 +23,25 @@ fi
 PYMODULES="$HOME/pymodules"
 export PYTHONPATH="${PYMODULES}${PYTHONPATH:+:${PYTHONPATH}}"
 
+# Use xpidl and webidl parsers from the tree itself if available.
+# They are unstable and for instance the esr140 and nightly tree disagree on how to define async iterables in WebIDL.
+if [ -f "${FILES_ROOT}/xpcom/idl-parser/xpidl/xpidl.py" -a \
+     -f "${FILES_ROOT}/dom/bindings/parser/WebIDL.py" ]; then
+  TREE_PYMODULES="/tmp/pymodules-${TREE_NAME}"
+  mkdir -p "${TREE_PYMODULES}"
+  pushd "${TREE_PYMODULES}"
+  cp "${FILES_ROOT}/xpcom/idl-parser/xpidl/xpidl.py" ./
+  cp "${FILES_ROOT}/dom/bindings/parser/WebIDL.py" ./
+  mkdir ply
+  pushd ply
+  for PLYFILE in __init__.py lex.py yacc.py; do
+    cp "${FILES_ROOT}/other-licenses/ply/ply/${PLYFILE}" ./ || cp "${FILES_ROOT}/third_party/python/ply/ply/${PLYFILE}" ./
+  done
+  popd
+  popd
+  export PYTHONPATH="${TREE_PYMODULES}${PYTHONPATH:+:${PYTHONPATH}}"
+fi
+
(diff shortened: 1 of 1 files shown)


nicolas-guichard:
The config2 test run failed with “No space left on device” (it probably didn't use the right instance type?), but it passed the idl-analyze stage for mozilla-beta without error (ie it didn't fail when parsing the older `async iterable` syntax).

Relevant excerpt from `index-log`:
```
160550   │ + export PYTHONPATH=/tmp/tree-pymodules:/home/ubuntu/pymodules:/home/ubuntu/mozsearch/scripts
160551   │ + PYTHONPATH=/tmp/tree-pymodules:/home/ubuntu/pymodules:/home/ubuntu/mozsearch/scripts
160552   │ + cat /mnt/index-scratch/mozilla-beta/idl-files
160553   │ + parallel /home/ubuntu/mozsearch/scripts/idl-analyze.py /mnt/index-scratch/mozilla-beta '/mnt/index-scratch/mozilla-beta/git/{}' '>' '/mnt/index-scratch/mozilla-beta/analysis/{}'
<...>
161773   │ + cat /mnt/index-scratch/mozilla-beta/webidl-files
161774   │ + /home/ubuntu/mozsearch/scripts/webidl-analyze.py /mnt/index-scratch/mozilla-beta /mnt/index-scratch/mozilla-beta/git /mnt/index-scratch/mozilla-beta/analysis /tmp null
<...>
162629   │ + echo 0
162630   │ 0
```

asutherland:
> The config2 test run failed with “No space left on device” (it probably didn't use the right instance type?),

Is your local [trigger_common.py up-to-date to have the change to m6id instances](https://searchfox.org/mozsearch/rev/cc540e90a971d70f32eec513192323a535ced60f/infrastructure/aws/trigger_common.py#97) which fixed config2 having this problem as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1970807#c3 ?

nicolas-guichard:
Yes it was an m6id.2xlarge instance.

After further inspection of the `index-log` (which I saved and can share if you want), I see it failed when moving the index from scratch space to EBS (ie during `mv /mnt/index-scratch/mozilla-esr140 /index/mozilla-esr140`).

And indeed I see in `attach-index-volume.sh`:
```
# We've been using 300G for quite some time but with the addition of esr140 to
# config2, we hit the 300G limit; the math indicates we probably need at least
# 305G before the nginx cache which we cap at 20G.  To provide headroom, I'm
# going to use 400 for release2 but we can revisit as we rebalance things.
volumeSize = 300
if channel == 'release2':
    volumeSize = 400
```
…but in my case `channel == "kdab"`.

asutherland:
Ah, yeah, that's the other half of the fix.  I definitely was hand-waving that we almost never run config2 jobs manually to avoid either propagating the original config file name through or to put it in the JSON file.  We could do either, although I think in the short-to-medium-term I am interested in having us move to just having the m6id.2xlarge indexers just turn into web-servers when they are done, storing everything on the SSD and avoiding EBS entirely.  (But that would be too expensive for the m6id.4xlarge indexers where I think we still only need an m6id.2xlarge for web serving.  But also we don't care about perf as much for config4 or config5, or at least I'm not sure we can initially justify the cost, whereas we can for config1-3 for the firefox trees.)

PR #888, opened 2025-08-02
Get clang plugin and wasm analyzer paths from env

nicolas-guichard:
This is another step to be able to point the scripts at prebuilt/prepackaged binaries.

EDIT: forgot to mention this passed a test on kdab.searchfox.org, using the Graphviz tree as usual. It has some html and xml files so html-files and js-files are non-empty and html-analyze.sh and js-analyze.sh actually call js-analyze.js.


asutherland (approved):
Thanks!

PR #891, opened 2025-08-11
Misc changes to enable packaging

nicolas-guichard:
These are the last changes I needed to make to have my nixified tree working:
- use the relative location from include dirs instead of the absolute path when building a macro identifier, so that we get the same hash when the macro comes from `/usr/include` or from `/nix/store` (provided the macro is actually in the same location in the file)
- use the `test-index` binary from `$PATH` instead of running it with `cargo run`
- fix using `grep` from `$PATH` in a Python script
- when self-indexing in the built-in searchfox and webtest configs, make the distinction between `$MOZSEARCH_PATH` which points at the (maybe packaged) mozsearch scripts and `$MOZSEARCH_SOURCE_PATH` which points at the source tree to be indexed
- when generating the `config.json` in `web-server-setup.sh`, generate it in the `$SERVER_ROOT` rather than the `$WORKING` dir (which is the index directory), this way we can make the index directory read-only

This also fixes some mistakes I made with the Dockerfile pointing to wrong directories in https://github.com/mozsearch/mozsearch/pull/888.

As usual I tested this by deploying `just-mozsearch.json` to https://kdab.searchfox.org.

### DIFF
--- a/infrastructure/indexer-setup.sh
+++ b/infrastructure/indexer-setup.sh
@@ -22,7 +22,7 @@ else
     rm -rf $WORKING/*
 fi
 
-$MOZSEARCH_PATH/scripts/generate-config.sh $CONFIG_REPO $CONFIG_INPUT $WORKING
+$MOZSEARCH_PATH/scripts/generate-config.sh $CONFIG_REPO $CONFIG_INPUT $WORKING $WORKING
 CONFIG_FILE=$WORKING/config.json
 
 for TREE_NAME in $(jq -r ".trees|keys_unsorted|.[]" ${CONFIG_FILE})
--- a/infrastructure/reblame-run.sh
+++ b/infrastructure/reblame-run.sh
@@ -26,7 +26,7 @@ else
     rm -rf $WORKING/*
 fi
 
-$MOZSEARCH_PATH/scripts/generate-config.sh $CONFIG_REPO $CONFIG_INPUT $WORKING
+$MOZSEARCH_PATH/scripts/generate-config.sh $CONFIG_REPO $CONFIG_INPUT $WORKING $WORKING
 CONFIG_FILE=$WORKING/config.json
 
 for TREE_NAME in $(jq -r ".trees|keys_unsorted|.[]" ${CONFIG_FILE})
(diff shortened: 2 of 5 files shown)


nicolas-guichard:
Please do tell me if you hit things I broke, the Docker setup is the least tested of the 3 envs we currently have.

asutherland:
Realized I should mention how it went!  No problems.

The only weird thing I experienced is I think an intentional side-effect of the changes to [how we run nginx now](https://searchfox.org/mozsearch/rev/f82c282f7f4e42bfcfb1d16708e746b87d6ee4c7/infrastructure/web-server-setup.sh#95).  If running `make build-test-repo` in /vagrant, and then for something like favicon with is a 404 or a random gibberish URL, nginx errors are showing up in the console, like (reproducing this again now):

```
vagrant@c5d41e43ebff:/vagrant$ 2025/08/15 01:34:57 [error] 19542#19542: *176 open() "/usr/share/nginx/html/favicon.ico" failed (2: No such file or directory), client: 172.17.0.1, server: , request: "GET /favicon.ico HTTP/1.1", host: "localhost:16995", referrer: "http://localhost:16995/tests/source/big_cpp.cpp"

vagrant@c5d41e43ebff:/vagrant$ 2025/08/15 01:36:49 [error] 19542#19542: *175 open() "/usr/share/nginx/html/asdfasdasfas" failed (2: No such file or directory), client: 172.17.0.1, server: , request: "GET /asdfasdasfas HTTP/1.1", host: "localhost:16995"
```

The 404's do show up in the /tmp/searchfox.log our config specifies, so maybe we should drop `-e stderr`?

arai-a:
@nicolas-guichard 

The change to MozsearchIndexer.cpp is not reflected to f-m's [MozsearchIndexer.cpp](https://searchfox.org/firefox-main/rev/b59df383b07ba23db8a6767af402b9e905f4e902/build/clang-plugin/mozsearch-plugin/MozsearchIndexer.cpp).
Is this intentional, or should we reflect it to f-m as well?

asutherland:
I was assuming we would reflect it to firefox-main (eventually).  I don't think there was any urgency to reflect it because we only care about stability in places we do checks and we sorta gave up on that in production.  But it could be handy if we start turning checks back on in production!

PR #892, opened 2025-08-19
Firefox main support

asutherland:


asutherland:
(note that CI is sad because I had merged the branch and it depended on the ref which went away on merge.)

PR #896, opened 2025-08-29
Bug 1985923 - Absorb the Location class difference across versions

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1985923

This does:
  * add a helper function to absorb the `Location` class difference between versions


asutherland (approved):
Thanks!  This is much better than ignoring the warning!

PR #897, opened 2025-08-29
Bug 1986051 - Ignore possibly non-existent webidl files.

arai-a:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1986051

This does:
  * Let the webidl analysis ignore known possibly-non-existent files

There's corresponding change in config repo that adds the renamed file to non-esr branches.

This can be removed once all non-esr branches receive the [bug 1919582](https://bugzilla.mozilla.org/show_bug.cgi?id=1919582) changes.

I'll post update when the test run finishes.
This might require some more fix in the other consumer of the file list.

### DIFF
--- a/scripts/webidl-analyze.py
+++ b/scripts/webidl-analyze.py
@@ -871,6 +871,14 @@ def parse_files(index_root, files_root, analysis_root, cache_dir, bindings_local
         else:
             fname = os.path.join(files_root, local_path)
 
+        if not os.path.exists(fname):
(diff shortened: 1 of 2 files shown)


asutherland (approved):
Thanks!  Feel free to also just land fix-ups as needed to make things happy!

arai-a:
Also added the exception to the rust code.
The same could actually be done with config files, but I'll look into it in followup.

PR #899, opened 2025-09-04
Bug 1986904 - Load blame.js in class field layout table.

arai-a:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1986904

This does:
  * Add blame.js also to scroll_footer.liquid, to make the list in sync with output.rs
  * Add a comment to make them in sync


nicolas-guichard:
CI is right: it seems you need to update the snapshots (you can download them from GitHub Actions [here](https://github.com/mozsearch/mozsearch/actions/runs/17464710462/artifacts/3927602022)).
Otherwise LGTM!

asutherland:
Yay CI!  (Not that this is a case where the expectations are providing a particularly useful service.  Although it is one of those things where I think it is informative and easy to skim, and there are cases where the broad expectations on those files could still be useful in terms of helping us notice unpleasant surprises.)

asutherland:
> CI is right: it seems you need to update the snapshots (you can download them from GitHub Actions [here](https://github.com/mozsearch/mozsearch/actions/runs/17464710462/artifacts/3927602022)). Otherwise LGTM!

Oh wow, that's awesome!  (The updated test artifacts zip ball.)  I'd either forgotten about that or missed that great affordance in the first place, thanks very much for implementing it and then calling it out here!

PR #900, opened 2025-09-08
Bug 1987330 - Ignore CSS-modules-related warnings.

arai-a:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1987330

This does:
  * Add files which uses import declarations for CSS modules to the ignore list


asutherland:
Thanks!

PR #901, opened 2025-09-21
Bug 1989736 - Add an /oldcommit/ endpoint for old m-c /commit/ links

asutherland:
In bug 1890435 I'd punted on this mapping early on in the process assuming we'd be burning comparatively costly and limited Elastic Load Balancer rule resources, but we ended up just having all mozilla-central links sent to the firefox-main tree and having nginx then rewrite the m-c links to f-m links so /rev/ could become /oldrev/ and unambiguously be clear about which type of rev we're dealing with.  This means there's only the additional nginx rule and the marginal increase in web-server logic, which is mainly down to copy-and-paste.  (I opted to "repeat yourself" in the interest of clarity.)

When doing the update I noticed that we didn't have path length guards on hgrev and oldrev to match rev, so I added those.  This changes what would have been a 500 because of a (caught) panic to a 404. Note that the rev-flavored guards rely on rust slicing being fine with slicing at its length to produce an empty slice.

The 2nd commit in the stack adds /diff/ mapping.

The only other revision-accepting endpoint we have in web-server.rs is commit-info which is only used for the blame strip hover purposes and so is only ever same-tree, so we're not adding one for that.  I've added a comment in the file to that effect since it seems like a reasonable question someone might have while reading the code.

### DIFF
--- a/scripts/nginx-setup.py
+++ b/scripts/nginx-setup.py
@@ -294,6 +294,7 @@ for repo in config['trees']:
         oldtree_name = tree_config['oldtree_name']
         print(f'  rewrite ^/{oldtree_name}/rev/(.*)$ /{repo}/oldrev/$1 permanent;')
         print(f'  rewrite ^/{oldtree_name}/commit/(.*)$ /{repo}/oldcommit/$1 permanent;')
+        print(f'  rewrite ^/{oldtree_name}/diff/(.*)$ /{repo}/olddiff/$1 permanent;')
         print(f'  rewrite ^/{oldtree_name}/(.*)$ /{repo}/$1 permanent;')
         print('')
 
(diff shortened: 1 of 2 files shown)


asutherland:
Tested locally with URL http://localhost:16995/mozilla-central/commit/a589ce1e3ff122d53690f101780a3b4747946d71 after a `make serve-firefox-repo` using a stale `make build-firefox-repo` from the [helpful examples in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1989736#c0).

NB: I'm going to wait for a review on this.

asutherland:
Oh, I guess there's the /diff/ endpoint too, and it seems like :robwu has been a fan of using all of these endpoints in links, as I think he was the first example I found of using commit, and [here's a use of /diff/](https://bugzilla.mozilla.org/show_bug.cgi?id=1396684#c17) so I'll quickly add that too.

asutherland:
Okay, and the /diff/ redirection has been tested locally with http://localhost:16995/mozilla-central/diff/d78ea2931e7a7c2e44eaf59e31c1d9e6028d5e56/toolkit/components/extensions/ext-tabs-base.js#376 from [this bugzilla comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1396684#c17).

PR #902, opened 2025-09-23
Bug 1990128  - Point hg.m.o links at "default" instead of "tip"

jcristau:
Since the git migration, "tip" can point at a changeset on a "tags" branch instead of "default", which won't have the files we're looking for.


asutherland:
CI happy, landing and rebasing my local WIP changes.

Thanks again for proactively helping address this searchfox issue!

PR #904, opened 2025-10-06
Bug 1992653 - Sort variants by platform names.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1992653

This does:
  * sort variants by the platform names, to get stable result for each run


arai-a:
Apparently this doesn't yet solve the issue.
https://github.com/arai-a/mozsearch/actions/runs/18272536013/job/52017474498

Perhaps we should sort the result before choosing the best platform.
I'll look into the details.

arai-a:
I misunderstood the automation result.
it was for the previous version where the snapshot was outdated.
it should be working.

PR #905, opened 2025-10-06
Bug 1992698 - Add filter to ssh.py -.

arai-a:
For https://bugzilla.mozilla.org/show_bug.cgi?id=1992698

This does:
  * Add an optional pattern parameter to ssh.py with "-" option, to apply a substring filter to the instance list

For example:
  * `ssh.py - release1` lists only the instances with release1 channel
  * `ssh.py - config1` lists only the instances with config1 config
  * `ssh.py - arai` lists only the instances with arai channel

### DIFF
--- a/infrastructure/aws/ssh.py
+++ b/infrastructure/aws/ssh.py
@@ -22,7 +22,7 @@ from six.moves import input
 ec2 = boto3.resource('ec2')
 client = boto3.client('ec2')
 
-def print_instances(select):
+def print_instances(select, pattern):
     now = None
 
     ids = {}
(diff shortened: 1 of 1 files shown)


asutherland (approved):
Nice!  I'd still been using `channel-tool.py list` and copying and pasting the id because the visual hierarchy of the output there was easier to scan, but I think this tips the balance to ssh.py!

I'm neutral on this, but since the instance identifiers have a specific format that always starts with "i-", it seems like you could probably assume that the first arg is a pattern if it doesn't start with "i-" and avoid needing to type "-" (other than to specify selecting is desired but with no pattern).  But that's mainly me expressing I don't mind if we change the behavior there; I'm also fine with it working this way.

arai-a:
Good point. I'll modify it to use the first arg.
That way the current "-" actually matches the filter behavior, as the line should definitely contain "-".

PR #1027, opened 2026-02-13
Coverage/blame popup follow-up improvements

nicolas-guichard:
This implements accessibility improvements suggested by Andrew in https://github.com/mozsearch/mozsearch/pull/1012#pullrequestreview-3794738418.

I also noticed I broke the .coverage-entry div in my previous PR.


asutherland:
Is it possible to stand this up on kdab.searchfox.org (or is this already happening)?

nicolas-guichard:
I just triggered a run.
Note that this run will also test https://github.com/nicolas-guichard/firefox-codecov which has all (quantized) code coverage data from January.

asutherland:
I was just poking around the historical coverage and https://kdab.searchfox.org/firefox-main/rev/1547677ded5c17d6b71576c42118a248966ff041/docshell/base/nsDocShell.cpp#491 seems weird in terms of when showing the actual data points for the Create method, there's just 2 and then we're interpolating over the whole of the method?

nicolas-guichard:
That data is weird indeed, but that's what we get from the raw coverage-report.json, see the output of:
```
gcloud storage cat gs://relman-code-coverage-prod/mozilla-central/8fe88d73a0faa54cee500ad88deb350cb21f3070/all:all.json.zstd | unzstd | jq '.children["docshell"].children["base"].children["nsDocShell.cpp"].coverage[490:583]'
```
We get 168663 hits for lines 491 and 493 and nothing for the rest…

nicolas-guichard:
The issue with weird coverage data must be https://bugzilla.mozilla.org/show_bug.cgi?id=2002904.

asutherland:
I did a quick test with NVDA and this seems to have broken pushing down to get into the popup from the table cell; it's not immediately clear what's going on.

nicolas-guichard:
I should probably undust my Windows VM and install NVDA on there. I don't see the regression with Orca and can navigate from the cell to the popup with keydown.

nicolas-guichard:
@asutherland I've tested this with NVDA locally and _can_ navigate to the popup using the down arrow after triggering it with enter. 🤔

asutherland (approved):
Thanks for having gotten the windows NVDA setup working.  Presumably I did something wrong in my initial testing; I was on a Windows laptop with an unfamiliar keyboard and may have exited browser mode when I meant to enter it or used the wrong keys or something like that.  I've since wiped the machine to be Ubuntu-only so can't attempt to reproduce on that one in particular.

PR #1080, opened 2026-04-05
Bug 2021777 - Apply pan/zoom to the normal diagram.

arai-a:
for [bug 2021777](https://bugzilla.mozilla.org/show_bug.cgi?id=2021777)

This does:
  * Remove unused code around the diagram
    * For simplicity, removed the "Blurring" code and related style rules
  * Make `graph.js` a place for storing other diagram-related functionalities
    * `graph.js` is always loaded
    * The client-side rendering is performed only when `GRAPH_INPUT` variable is given, which happens only when the "interactive" style is chosen
  * Split pan-and-zoom functionality into a separate class `PanAndZoom`, to make it available for server-side-rendered diagram
  * Move all diagram-related functionalities into `graph.js`, with putting all features into `Diagram` class
    * Move the `Diagram` class from `context-menu.js` to `graph.js`
    * Move the diagram-specific hover functionality into Diagram class
  * Use the `PanAndZoom` also in the server-side-rendered diagram
    * Currently the pipeline interface itself allows having multiple graphs, so there's a code to avoid breaking 2+ graphs case, but the case doesn't happen right now.  we can look into simplification in a followup

Available on the testing server
  * small diagram: https://arai.searchfox.org/firefox-main/query/default?q=calls-between%3A%27Promise_static_all%27+calls-between%3A%27NewPromiseCapability%27+depth%3A10+path-limit%3A1967
  * large diagram: https://arai.searchfox.org/firefox-main/query/default?q=calls-to%3A%27js%3A%3APromise_then%27%20depth%3A4


asutherland:
Exciting!  And yes, it makes sense to remove the unused blurry fun I was having.  I had left it in so I could try it randomly to see if it was useful and... it was not ;)

I think this might be a good opportunity to move to not requiring the ctrl key to zoom if we are going for the standard maps-style interaction.

I personally am not a fan of defaulting to zoomed out when the graph is too large.  I'm willing to accept that I might be in the minority on that, but would want some kind of preference/sticky-UI setting.  But I do think we need to do something better to help show where the depth=0 nodes than we're currently doing with the slightly bolder node border since that's of course much more subtle when zoomed way out; here's what I'm seeing:
image

I would propose that one option would be to auto-sticky the depth=0 nodes when we're zoomed out.  I think combined with the scroll-wheel defaulting to zoom that will make the situation more interesting and intuitive to people if we think defaulting to zoomed out is optimal.  (I'm also fine to make :smaug have to change a setting ;)

arai-a:
Thank you for reviewing!  I'll address the default zoom/focus state, wheel behavior after rebasing on https://github.com/mozsearch/mozsearch/pull/1081 .

Also, the depth=0 node could be rendered more differently especially when zoomed out.

Also we could look into adding a separate navigator view like photoshop, so that you can see where you're when zoomed in, and then starting from zoomed-in state still can show the entire graph.
https://helpx.adobe.com/photoshop/using/viewing-images.html

I'll look into the latter two in followup bugs.

asutherland (approved):
> Also, the depth=0 node could be rendered more differently especially when zoomed out.

Yes, this is definitely an aspect of the client-side layout that's very exciting but where I fear we're more limited for the graphviz layout.  (And where I never particularly found [non-linear lens distortions like fish-eye](https://bost.ocks.org/mike/fisheye/) more useful or at least more intuitive than a mini-map.)

> Also we could look into adding a separate navigator view like photoshop, so that you can see where you're when zoomed in, and then starting from zoomed-in state still can show the entire graph. https://helpx.adobe.com/photoshop/using/viewing-images.html

Yes, an interactive mini-map definitely makes sense (and :smaug has been pushing for one for quite some time too, so he will be very happy!).

I think there's also some exciting potential here to build on the mini-map to provide some contextual up-sells for diagrams in the navigation panel.  Some of my early diagramming experiments were providing local control-flow overview diagrams ([ex 1](https://clicky.visophyte.org/files/screenshots/20190821-164905.png), [ex 2](https://clicky.visophyte.org/files/screenshots/20190821-165907.png)) in context menus.  A major piece of feedback I received was that people did not like giant context menus that blocked out the source code and that still informs my thinking in this space today.  So then that involved experimenting with other things like [having a sort of horizontal navigation panel context area at the top where we could show diagrams and the defs/decls of things](https://clicky.visophyte.org/files/screenshots/20200331-011202.png) (and also trying with [that area having multiple tabs itself](https://clicky.visophyte.org/files/screenshots/20200403-133604.png)).  Which is how I think :kat's proposal of [a super sidebar](https://mozilla.staktrace.com/tmp/mockup.html) with optional panels/widgets I think ended up being the best path forward.

So we could (on an opt-in, experimental basis) display diagrams like that zoomed out in the navigation panel/sidebar but if a user clicked on them we could display them in a light-box overlay.  (Although again there's some interesting potential from the client-side layout to potentially have the overview be laid out differently/with more strict criteria in the sidebar.  Like for [my mockup about potentially showing a contextual diagram with keyboard accelerators in the sidebar to contextualize what the many options are in the context menu](https://bugzilla.mozilla.org/show_bug.cgi?id=1861773#c2).)

> I'll look into the latter two in followup bugs.

Sounds good!

PR #1081, opened 2026-04-07
Bug 2030042 - Implement client-side 'Ignore this node' for interactive graphs

Sameem-baba:
For - https://bugzilla.mozilla.org/show_bug.cgi?id=2030042

This PR implements the client-side **"Ignore this node"** functionality for the interactive graph layout. Previously, ignoring a node required a full server round-trip to re-query and re-render the graph.

### DIFF
--- a/static/js/context-menu.js
+++ b/static/js/context-menu.js
@@ -1923,7 +1923,13 @@ var ContextMenu = new (class ContextMenu extends ContextMenuOrSubMenu {
               icon: "brush",
               section: "diagrams-ignore",
               action: () => {
-                Diagram.ignoreNode(symInfo.pretty);
+                if (window.interactiveGraphInstance) {
+                  window.interactiveGraphInstance.removeNodeByPrettyName(symInfo.pretty);
+                  window.interactiveGraphInstance.updateQueryForIgnoredNode(symInfo.pretty);
+                  this.hide();
+                } else {
+                  Diagram.ignoreNode(symInfo.pretty);
+                }
               },
             }));
           }
--- a/static/js/graph.js
+++ b/static/js/graph.js
@@ -394,6 +394,7 @@ class Graph {
   constructor(viewport, container, input, sources, targets, extraList) {
     this.viewport = viewport;
     this.graphContainer = container;
+    this.extraList = extraList;
 
     const viewportRect = this.viewport.getBoundingClientRect();
     this.viewportSize = {
@@ -411,6 +412,19 @@ class Graph {
     this.edges = [];
 
     this.createBlocksAndEdges(input, sources, targets);
+
+    this.render();
+
+    this.needInitialFitToViewport = true;
+    this.addEventListeners();
+    this.initialFitToViewport();
+    this.addButtons();
+  }
+
+  render() {
+    this.graphContainer.replaceChildren();
+
+    this.resetBlockAndEdgeState();
 
     this.calculateDepth();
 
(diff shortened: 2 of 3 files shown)


static/js/graph.js
```diff
@@ -412,6 +413,26 @@ class Graph {
...
+    this.render();
+
+    this.needInitialFitToViewport = true;
+    this.addEventListeners();
+    this.initialFitToViewport();
+    this.addButtons();
+  }
+
+  render() {
+    this.graphContainer.innerHTML = "";
```
arai-a: This can use `.replaceChildren()` instead

static/js/graph.js
```diff
@@ -412,6 +413,26 @@ class Graph {
...
+    
+    for (const block of this.blocks) {
+      block.depth = -1;
+      block.breadth = -1;
+    }
+
+    for (const edge of this.edges) {
+      edge.path = [];
+    }
```
arai-a: It would be nice to add separate method for them, such as `resetBlockAndEdgeState` for clarity.

static/js/graph.js
```diff
@@ -442,15 +463,92 @@ class Graph {
...
-    this.needInitialFitToViewport = true;
+    // Re-populate interaction data
+    this.extraList.length = 0; 
+    this.populateExtra(this.extraList);
+  }
 
-    this.populateExtra(extraList);
+  removeNodeByPrettyName(prettyName) {
+    const blockToRemove = this.blocks.find(b => b.pretty === prettyName);
+    if (!blockToRemove) return;
```
arai-a: Let's always put braces, especially for new-ish files.

static/js/context-menu.js
```diff
@@ -1923,7 +1923,12 @@ var ContextMenu = new (class ContextMenu extends ContextMenuOrSubMenu {
               icon: "brush",
               section: "diagrams-ignore",
               action: () => {
-                Diagram.ignoreNode(symInfo.pretty);
+                if (window.interactiveGraphInstance) {
+                  window.interactiveGraphInstance.removeNodeByPrettyName(symInfo.pretty);
```
arai-a, line 1927: This should also update the query field and the URL,
so that loading the URL shows the same graph, so that for example you can post the URL to bug comment, or pass it to other people.
Similar thing is done in https://searchfox.org/mozsearch/rev/84b0be4db1e98b99eea7452671078d3cc7044951/static/js/search.js#443,457,459
```js
updateColQuery() {
...
  this.fields.query.value = query;
...
  this.updateHistory(url);
```

static/js/graph.js
```diff
@@ -442,15 +463,92 @@ class Graph {
...
+        block.loopback = null;
+      }
+    }
+
+    this.pruneUnreachableNodes();
+
+    this.render(false);
+  }
+
+  pruneUnreachableNodes() {
```
arai-a, line 499: This algorithm works for "Calls from" and "Calls to", but not for "Calls between".
The "Calls between" should prune also all nodes and edges that doesn't connect between the specified nodes, in the specified direction.

For example, in the tests repository, querying `calls-between-source:'diagram_ignore::F7' calls-between-target:'diagram_ignore::F1' graph-format:mozsearch-interactive` and then ignoring F3 should prune F2 as well.

static/js/graph.js
```diff
@@ -442,15 +463,92 @@ class Graph {
...
+      if (isolatedIds.size > 0) {
+        this.blocks = this.blocks.filter(b => !isolatedIds.has(b.id));
+      }
+      return;
+    }
+
+    const validIds = new Set();
+    focalNodes.forEach(b => validIds.add(b.id));
+
+    const forwardQueue = [...focalNodes];
```
arai-a: This can be `focalNodes.slice()`

static/js/graph.js
```diff
@@ -442,15 +463,92 @@ class Graph {
...
+
+      for (const succ of curr.succs) {
+        if (!forwardVisited.has(succ.id)) {
+          forwardVisited.add(succ.id);
+          forwardQueue.push(succ);
+        }
+      }
+    }
+
+    const backwardQueue = [...focalNodes];
```
arai-a: Here too. slice can be used

static/js/graph.js
```diff
@@ -442,15 +463,92 @@ class Graph {
...
 
-    this.addButtons();
+    this.edges = this.edges.filter(edge => {
+      return edge.pred.id !== blockIdToRemove && edge.succ.id !== blockIdToRemove;
+    });
+
+    for (const block of this.blocks) {
+      block.preds = block.preds.filter(p => p.id !== blockIdToRemove);
+      block.succs = block.succs.filter(s => s.id !== blockIdToRemove);
+      if (block.loopback && (block.loopback.pred.id === blockIdToRemove || block.loopback.succ.id === blockIdToRemove)) {
```
arai-a: loopback connects the same block.
So this block is unnecessary.

static/js/graph.js
```diff
@@ -442,15 +463,92 @@ class Graph {
...
+        this.blocks.filter(b => b.preds.length === 0 && b.succs.length === 0 && !b.loopback).map(b => b.id)
+      );
+      if (isolatedIds.size > 0) {
+        this.blocks = this.blocks.filter(b => !isolatedIds.has(b.id));
+      }
+      return;
+    }
+
+    const validIds = new Set();
+    focalNodes.forEach(b => validIds.add(b.id));
```
arai-a: `validIds` should be empty at this point.
The following loop adds the ids in the queue.

arai-a (changes requested):
Thank you for working on this!
While the basic functionality for "calls from"/"calls to" looks good, there are some issues with the "calls between", and also the query parameter.

Can you address those comments, and also add some testcases?

Sameem-baba:
I have addressed the comments mentioned! I will add the testcase in the next commit

static/js/context-menu.js
```diff
@@ -1923,7 +1923,28 @@ var ContextMenu = new (class ContextMenu extends ContextMenuOrSubMenu {
...
+
+                  const url = new URL(window.location.href);
+                  let currentQuery = url.searchParams.get("q");
+
+                  if (currentQuery) {
+                    currentQuery += ` ignore-nodes:${symInfo.pretty}`;
```
arai-a: `ignore-nodes` takes a comma-separated list of pretty symbols.
So, this should behave differently if there's existing `ignore-nodes`.

The [Diagram.applyOptions](https://searchfox.org/mozsearch/rev/84b0be4db1e98b99eea7452671078d3cc7044951/static/js/search.js#880-898) logic can be reused for adding/replacing the query

that said, let's add a testcase that performs "ignore node" twice.

static/js/context-menu.js
```diff
@@ -1923,7 +1923,28 @@ var ContextMenu = new (class ContextMenu extends ContextMenuOrSubMenu {
...
+                    url.searchParams.set("q", currentQuery);
+
+                    window.history.pushState({}, "", url);
+
+                    const searchInput = document.getElementById("query");
+                    if (searchInput) {
+                      searchInput.value = currentQuery;
+                    }
+                  }
```
arai-a: Can you put this block into a method in Graph class?

static/js/graph.js
```diff
@@ -442,15 +456,97 @@ class Graph {
...
+    };
+
+    const isCallsBetween = sources.length > 0 && targets.length > 0 && !sources.some(s => targets.includes(s));
+
+    let validIds = new Set();
+
+    if (isCallsBetween) {
+      const fromSources = getReachable(sources, "succs");
+      const toTargets = getReachable(targets, "preds");
+      validIds = new Set([...fromSources].filter(id => toTargets.has(id)));
```
arai-a: "calls-between" has two types:
  * directed: a query with `calls-between-source:` and `calls-between-target:`.  this is for finding a path from source to target.  in this case, each symbols are marked "source" or "target".
  * undirected: a query with two `calls-between:`.  this is for finding a path between them, in any direction (A to B, or B to A).  in this case none of them are marked "source" or "target

for example, querying with `calls-between:'diagram_ignore::F7' calls-between:'diagram_ignore::F1' graph-format:mozsearch-interactive`, and then ignoring F2 should remove F3 as well.

let's add testcases for both directed and undirected cases,

arai-a (changes requested):
Thank you for updating!
This is getting really well.
There are some more issues around the query parameter, and the undirected calls-between.

Then, let me know if you have any questions around the functionality.
I think the diagram feature is lacking some documents, which we should add.

arai-a (approved):
Excellent!
Thank you so much!

PR #1083, opened 2026-04-17
Bug 1593367 - Remove Mutex in web-server

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=1593367

web-server had a Mutex which effectively removed any chance of concurrency.

This requires making GitData: Sync, which is achieved with the new ThreadLocalRepository helper.


arai-a (approved):
Looks good!

PR #1085, opened 2026-04-17
Slightly improve update script error handling and reduce disk space usage by removing rust-indexer

nicolas-guichard:
While playing with scip-typescript, I hit an out-of-space error case in the update script which wasn't reported and instead the indexer ran with old code.

An easy target to reclaim some disk space is to remove rust-indexer which is unused.


asutherland (approved):
Thanks!

PR #1089, opened 2026-05-25
Install libarchive-tools to get bsdtar

padenot:
This gives us a way to deal with all archive / compressed files.


emilio (approved):
Is removing `file` intentional?

padenot:
I thought I had checked that it wasn't in use any more, but I had scopped by grep, it's in use by `find-objdir-files.sh`. Restored.

PR #1091, opened 2026-05-29
Misc provisioning-related improvements

nicolas-guichard:
This a diverse mix of changes to the provisioning or update scripts that I came up with while looking at provisioning our scip-typescript fork.

- Really provisioning-related:
  - removed blocks marked with “TODO: remove”
  - switched to stable Rust instead of nightly. We don't need nightly anymore. Except rust-analyzer, downloaded directly from upstream instead of the “Rust stable” version which may lag behind.
  - removed the SCIP tooling
  - fixed the way PATH is updated in ~/.profile after preparing the livegrep venv
- Nix-related: updated flake.lock and remove .envrc
- Removed some Vagrant leftovers
- Updated GitHub Actions as suggested by GitHub


asutherland (approved):
Thanks!  I think it could still be nice to have a way to easily invoke "scip" to debug across all 3 of indexer (in the event of failure), web-server (to figure out why things are weird in production results), in addition to a local dev container (which is easiest because the homedir is sticky via volume), but I agree that having 0.5.0 hardcoded isn't super-helpful since it's at once out-of-date while also only the JSON dump format works (I think the texty format broke), so removing it for now makes sense.

I'm going to merge this now too since in https://bugzilla.mozilla.org/show_bug.cgi?id=2042682 it appears firefox-main has moved to clang-21 and I want to test the phab fixes :glandium put up against our test corpus and also then we probably want to rev provisioning to clang-21 as well.  I'll take a quick pass at doing all that with at least local validation so we can then speculatively land the f-m changes.

PR #1092, opened 2026-06-02
Upstream glandium's clang-21 fixes, switching to clang-21 and fixing a docker rust-analyzer path regression

asutherland:
The main thing here is to upstream https://phabricator.services.mozilla.com/D303283 which is necessary since firefox-main now uses clang-21.  This also moves us to provision clang-21.  While testing locally with docker, I noticed a rust-analyzer filter which is because the Dockerfile sets the PATH which overrides what bash tries to do with the ~/.profile and our changes there, so I update that.  And then there are minor snapshot changes that seem to have more to do with clang-21 than the required change and I've approved those.

Ideally I did the flake stuff correctly; I did not try that yet.

Note that to test/provision locally I needed to force the mozsearch repo and branch to correspond to this one so that provisioning would not fail when trying to compile the clang plugin.  Hopefully the CI approach doesn't run into that.

### DIFF
--- a/clang-plugin/MozsearchIndexer.cpp
+++ b/clang-plugin/MozsearchIndexer.cpp
@@ -681,6 +681,23 @@ private:
       return cast<FunctionDecl>(Decl)->getNameAsString();
     }
 
+#if CLANG_VERSION_MAJOR >= 21
+    // clang 21 (commit 6d00c4297f67) sets the DeclContext of lambdas inside
+    // requires-expression bodies to RequiresExprBodyDecl, but manglePrefix
+    // has no guard for it and crashes. Use a location-based name instead.
+    // Works around https://github.com/llvm/llvm-project/issues/200336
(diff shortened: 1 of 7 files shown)


nicolas-guichard (approved):
LGTM, thanks!

> there are minor snapshot changes that seem to have more to do with clang-21 than the required change

I agree they aren't related to the changes around lambdas.
Could you add a test that does? eg from glandium's upstream bug report https://github.com/llvm/llvm-project/issues/200336:
```cpp
template <class T>
concept C = requires(const T &t) {
    []<class U>(const U &) {}(t);
};
```

asutherland:
I was trying to find cycles to make the test change with an appropriate level of understanding and fidelity, but I think this is complicating the referenced PR, so I'm just going to land this now.  Nicolas, if you have cycles to add a test, it would very much be appreciated.

PR #1093, opened 2026-06-04
Use Nix to install livegrep, scip-python, rust-analyzer, wasm-snip and mozsearch tools

nicolas-guichard:
This is a partial rebase of https://github.com/mozsearch/mozsearch/pull/859, but instead of trying to drive everything from Nix right away, this only builds some dependencies with Nix and makes them available in PATH/MOZSEARCH_WASM_DIR/MOZSEARCH_CLANG_PLUGIN_DIR.

The main gain is the shared cache between the CI here, dev machines and the AWS deployment. When the indexer runs, if the CI has already built and pushed the packages to the cache, then it doesn't need to rebuild anything.

This also means the CI runs much faster because it doesn't have to rebuild our tooling 5 times.


nicolas-guichard:
Note this conflicts with https://github.com/mozsearch/mozsearch/pull/1092 which should probably be merged first?

asutherland (approved):
I gave this more of a cursory review.  I'm happy to fork scip-python from sourcegraph to the mozsearch repo if it helps or do whatever it necessary there to stop referencing my fork.  (Not of particular importance, though.)

I have not yet run locally with the changes recognizing that the use of nix by the default Makefile inside of docker now is notable (but good to get to converging; and I had run using the nix flake via the now-removed direnv stuff on my MoCo machine to good effect, so I have used it a little bit before).

Please feel free to update the PR / land fix-ups needed (not just for the minor bitrot my landing just caused).

nicolas-guichard:
> I'm happy to fork scip-python from sourcegraph to the mozsearch repo if it helps or do whatever it necessary there to stop referencing my fork. (Not of particular importance, though.)

I'm now applying your patch via Nix on top of the latest upstream tag. This removes the reference to your fork and gives us a more recent scip-python.

Another notable change: I changed the build-docker.sh script to bind mount /vagrant early and build our tooling from there. Ideally the docker build shouldn't ever git clone from GitHub (and the GitHub Actions CI also should only clone the repo once), but I'm leaving that for later. (that's https://github.com/mozsearch/mozsearch/pull/1093/commits/49f1ea7aeb006b75f5266ea691ae9cc301263d2f)

asutherland:
Oh neat, that is nice to just go directly from the checkout!

Especially since I did run into the problem of "you need to push to a publicly accessible branch to run provisioning locally via docker because of the update process wanting to pull things down from github" recently when testing the clang-21 changes locally.  (And the double-update check there was fundamentally because the shell based provisioning and update mechanism inherently is non-deterministic versus nix, and it was very easy to accidentally make a provisioning change that would fail on the subsequent update, especially as it related to trying to be clever about conditionally doing things based on the on-disk state.)

asutherland:
Interesting low-concern failure on config1 indexing mozsearch I think as a side-effect of this:

```
clang++ -Xclang -load -Xclang /nix/var/nix/profiles/default/lib/libclang-index-plugin.so -Xclang -add-plugin -Xclang mozsearch-index -Xclang -plugin-arg-mozsearch-index -Xclang /index/mozsearch/git -Xclang -plugin-arg-mozsearch-index -Xclang /index/mozsearch/analysis -Xclang -plugin-arg-mozsearch-index -Xclang /index/mozsearch/objdir -Xclang -fparse-all-comments -fPIC -g -Wl,-R -Wl,'$ORIGIN' -L/usr/lib/llvm-21/lib  -shared FileOperations.o StringOperations.o BindingOperations.o MozsearchIndexer.o from-clangd/HeuristicResolver.o -o /nix/var/nix/profiles/default/lib/libclang-index-plugin.so -lclangASTMatchers
/usr/bin/ld: cannot open output file /nix/var/nix/profiles/default/lib/libclang-index-plugin.so: Permission denied
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [/index/mozsearch/git/clang-plugin/Makefile:46: /nix/var/nix/profiles/default/lib/libclang-index-plugin.so] Error 1
+ handle_tree_error 'tree build script'
```

I believe this is an outgrowth of us doing this in indexer provisioning:
```
echo 'export MOZSEARCH_CLANG_PLUGIN_DIR="/nix/var/nix/profiles/default/lib"' >> ~/.profile
export MOZSEARCH_CLANG_PLUGIN_DIR="/nix/var/nix/profiles/default/lib"
```

I suspect we just need to set/pass-in MOZSEARCH_CLANG_PLUGIN_DIR in https://searchfox.org/mozsearch-mozilla/rev/1dd81d89d22bed342455fe0c3127b30dea741158/mozsearch/build#23-26 (our outer env var here does not propagate in and was not intended to, even though it is named very similarly to the Makefile var inside the [Makefile](https://searchfox.org/mozsearch/rev/3cc5676abd9df2739d8ef8d9eb3f48b532965676/clang-plugin/Makefile)).
```
CLANG_PLUGIN_DIR=$FILES_ROOT/clang-plugin
# Passing a VPATH will let make find the source files despite us using an objdir.
# Explicitly passing CC and CXX propagates the special clang flags.
make -f $CLANG_PLUGIN_DIR/Makefile VPATH=$CLANG_PLUGIN_DIR CC="$CC" CXX="$CXX"
```

PR #1094, opened 2026-06-04
Update scip-typescript to our fork

nicolas-guichard:
This sources scip-typescript from https://github.com/mozsearch/scip-typescript and starts actually using it in the test project.


nicolas-guichard:
Draft because it depends on https://github.com/mozsearch/mozsearch/pull/1093

scripts/js-analyze.js
```diff
@@ -3855,6 +3855,14 @@ while (true) {
   localFile = localRoot + "/" + sourcePath;
   const analysisFile = analysisRoot + "/" + sourcePath;
 
+  // Skip analysis if the output already exists (if it was generated by scip-typescript/scip-indexer for instance).
+  try {
+    os.file.readFile(analysisFile);
+    continue;
```
asutherland, line 3862: It could be nice to print out an informative message here about skipping this file since this could potentially be a footgun for someone who is trying to iterate on changes to the js-analyze behavior (and we haven't yet redirected output to the analysis file) and was assuming that re-running js-analzye would clobber existing files rather than skip them.

asutherland (approved):
Woo!  Very exciting.

Restating:
- In order to experience any changes in behavior for JS indexing, a tree needs to invoke `scip-typescript` and then either directly invoke `scip-indexer` or use the `scip_subtrees` mechanism in the config file which will also invoke scip-indexer.
- We aren't making any changes right now to the firefox trees, and for the test tree we're adding, although we are adding a package.json at the root without any specific paths/related constraints, we explicitly are only invoke scip-typescript on the "ts" directory where we have added all the ts files.  So existing js analysis test output/coverage is unaffected.
- With the change to js-analyze skipping any file it's told about that already has an analysis file, we should ideally be able to point scip-typescript at the things that are well set-up for typescript analysis and skip over things like wpt tests which are not.
  - I had hypothesized about the potential of doing something hybrid as a possibility, but to the extent we can avoid it, it will be nice.  But that's of course all future work.

PR #1095, opened 2026-06-11
Bug 2046800 - Do not give up after sending failure email.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=2046800

Given that these days there are frequent pipeline-server failures, and often we just restart it,
it's better just restart it automatically, while sending an email at some point.


emilio (approved):
Fair enough. I guess the worst thing that could happen is that we hit a bug and keep restarting over and over (thus getting tons of emails very frequently). But that seems not the end of the world? We could have a timer rather than a count I guess.

PR #1099, opened 2026-06-14
Bug 2047544 - Typed CrossrefData and JumprefData.

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=2047544

I started looking into the pipeline-server memory issue and noticed the crossref and jumpref data always used the `serde_json::Value` type. This introduces proper structs `CrossrefData` and `JumprefData` which use interned strings instead.

Now this doesn't fully fix the memory issue, but at least improves the situation, and also fixes a bug caused by a typo in cmd_crossref_expand.rs (`"/meta/overridenBy"`).

This also removes some dead code and renames `CrossrefLookupMap` since it is also used for jumref data.


tools/src/format.rs
```diff
@@ -228,20 +229,20 @@ pub fn format_code(
...
+                            //             "syntax".to_string(),
+                            //             json!(syntax_kind.to_string()),
+                            //         );
+                            //     }
+                            //     obj.insert("type".to_string(), json!(type_pretty.to_string()));
+                            //     if let Some(type_sym) = &a.type_sym {
+                            //         obj.insert("typesym".to_string(), json!(type_sym.to_string()));
+                            //     }
+                            //     generated_sym_info.insert(*sym, json!(obj));
+                            // }
```
nicolas-guichard: Now I'm unsure about what this should be, the other generated_sym_info.insert calls like on line 301 down below clearly pass jumpref data and commenting this doesn't fail any tests. Was this bogus or what am I missing?

tools/src/cmd_pipeline/symbol_graph.rs
```diff
@@ -228,34 +230,31 @@ pub fn semantic_kind_is_class(semantic_kind: &str) -> bool {
...
-            Some(Value::String(pretty)) => is_lambda_pretty(pretty),
+        match self.get_structured().map(|meta| meta.pretty) {
+            Some(pretty) => is_lambda_pretty(&pretty),
             _ => false,
         }
     }
 
     /// Provide the structured rep of this symbol if it has one.  If we use this
     /// a lot we should potentially consider using interior mutability to cache
     /// this or have performed the conversion eagerly upon creation.
```
asutherland: nit: This comment is mooted since we're handing out a ref and are getting the resource usage win.

tools/src/format.rs
asutherland: Some brief context in https://bugzilla.mozilla.org/show_bug.cgi?id=1776522#c0
> Since the landing of the structured record support we actually do have a SYM_INFO dict already providing { syntax, type, typesym } meta-information for locals which are not available in the crossref database as an artifact of the ReactJS-based prototype fancy UI which will never land.

Basically I had been prototyping being able to show the types of any locals you clicked on.  I hadn't landed UI changes for them because the naive UI was bad (those types can be huge) and constituted a change in muscle-memory that wasn't ready yet.  I left the info in at the time to make it easier to iterate on experimenting and prototyping, and because it could be useful at types to be able to inspect a token with devtools to get the node into `$0` and then do `
SYM_INFO[$0.dataset.symbols] ` in the console and be able to see the type info.  (I still do this regularly.)  But I mainly justified leaving it in because I was hoping I would get back to it in a timely fashion.

Unless someone else is actively going to pursue using the info, it probably makes sense to file a bug to remove the emission of the unused data from the clang plugin and eliminate this code and the comment above here but add a comment referencing the bug indicating that we do currently have the data.

asutherland (approved):
This is a really fantastic cleanup, thank you!  It's really great seeing all of the hacks and boilerplate and extra edge-cases related to the consumption of the serde_json Value disappear!  (Also nice to have that typo addressed and be confident such things can't happen again!)

Please associate this with a bug for tracking purposes.

I'd forgotten about the speculative `lazy_crossref` (which really shouldn't have landed in tree but my hobby stack came in without much additional processing).  Quite delightfully the combination of improved machines and :arai's fantastic overhaul of crossref to make it faster and for us to be able to easily re-run just the crossref step has largely mooted the motivating use-case.

tools/src/format.rs
nicolas-guichard: This is now https://bugzilla.mozilla.org/show_bug.cgi?id=2047542

PR #1101, opened 2026-06-15
tokenizer: Add support for <> includes

nicolas-guichard:
`#include <test.h>` was tokenized wrong, this now identifies <test.h> as a string literal just like `#include "test.h"` would.

### DIFF
--- a/tools/src/tokenize.rs
+++ b/tools/src/tokenize.rs
@@ -677,6 +677,34 @@ pub fn tokenize_c_like(string: &str, spec: &LanguageSpec) -> Vec<Token> {
                 kind: TokenKind::Identifier(class),
             });
             next_token_maybe_regexp_literal = false;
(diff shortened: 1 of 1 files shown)


asutherland (approved):
Nice!  Thanks!  I was looking for a place we use `<>` and actually have the headers in-tree; [these protobuf uses maybe work in f-m](https://searchfox.org/firefox-main/rev/4524afcc186654ad357e48fec686249515860ac8/devtools/shared/heapsnapshot/HeapSnapshot.cpp#7-9).

nicolas-guichard:
I actually hit this while building a Qt index (on KDAB time). Qt uses <> includes _a lot_.
See eg https://searchfox.guichard.eu/qt/source/qtbase/src/corelib/kernel/qobject.h

asutherland:
Neat!

PR #1102, opened 2026-06-15
Bug 2047759 - Revert DerivedSymbolInfo::crossref_info to be optional

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=2047759
Example: https://searchfox.org/firefox-main/query/default?q=calls-to%3A%27mozilla%3A%3Adom%3A%3ANavigator_Binding%3A%3AgetBattery_promiseWrapper%27%20depth%3A4

Since it used to be a `serde_json::Value`, `DerivedSymbolInfo::crossref_info` could previously be `Value::Null`. It seems I need to be less strict there and use `Option<CrossrefData>`.


nicolas-guichard:
While I hit this quite easily on firefox-main I didn't manage to reproduce this in our test repo: merging without a test for now.

PR #1104, opened 2026-06-16
crossref: sort identifiers directly in Rust

nicolas-guichard:
Instead of outputing the identifiers file in random order then sorting it using sort, this directly sorts the file from the Rust binary.

Note: the comment was outdated, this was changed from a BTreeMap to a HashMap some time ago.

This is required by https://github.com/KDABLabs/mozsearch-qt which doesn't use the `crossref.sh` wrapper and IMO makes more sense.

+ I'm sneaking in a fixup for https://github.com/mozsearch/mozsearch/pull/1102. It was late and I forgot to run `cargo fmt`.

### DIFF
--- a/scripts/crossref.sh
+++ b/scripts/crossref.sh
@@ -11,9 +11,3 @@ ANALYSIS_FILES_PATH=$3
 echo Root is $INDEX_ROOT
 
 crossref $CONFIG_FILE $TREE_NAME $ANALYSIS_FILES_PATH 16
-
-# Re-sort the identifiers file so that it's case-insensitive.  (It was written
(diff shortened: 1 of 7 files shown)


asutherland (approved):
Agreed that it's nicer to just emit the file sorted directly.

Nice call on the semi-sensitive sorting!

Restating:
- We're consuming the id_table as we sort it and accumulate the lines so our memory usage should look fairly similar before and after the change.  Additionally, the identifiers disk rep right now is only 881M for firefox-main so even if there was a transient memory usage spike it would be fine.
- The insensitive comparator is using iterators so we're lazily uppercasing and not performing full/any allocations.

PR #1106, opened 2026-06-16
Provide mozsearch router via Nix

nicolas-guichard:
This wraps router.py, making it easier to run without having to setup a venv, and removes the livegrep python library shenanigans from the provisioning scripts.

This is the last patch required by my Qt index.


asutherland (approved):
Looking forward to this!

PR #1107, opened 2026-06-16
Resize indexer root filesystem to 50 GiB

nicolas-guichard:
We were already close to the size limit and probably because of my Nix-related changes which cause some duplication (eg we currently have nodejs from both Ubuntu repos and from nixpkgs at this point), I think we now exceed it again.

This apparently caused some silent corruption during the previous run: provisioning seemingly succeeded but in fact many files appeared to be empty once the indexer ran.

### DIFF
--- a/infrastructure/aws/indexer-provision.sh
+++ b/infrastructure/aws/indexer-provision.sh
@@ -49,12 +49,12 @@ popd
 
 date
 
-# Size up our root partition to 20G
+# Size up our root partition to 50G
 #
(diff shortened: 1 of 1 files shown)


asutherland (approved):

nicolas-guichard:
This was not the issue after all: see [bug 2048040](https://bugzilla.mozilla.org/show_bug.cgi?id=2048040).
On the contrary disk usage was quite reduced!
The web-server now uses 10.5 GiB/20 GiB.
The indexer now uses 12.G GiB/20 GiB.

PR #1109, opened 2026-06-17
identifiers lookup: reduce number of temporary allocations

nicolas-guichard:
This avoids creating Strings for the needle and for each line and instead compares the strings case-insensitively using iterators.


asutherland (approved):
Awesome, thanks!

PR #1112, opened 2026-06-18
Limit pipeline-server to handling 4 concurrent requests (& update everything)

nicolas-guichard:
This really does two things:
- update everything to the latest version (except linkify because 0.4+ behaves differently)
- add a concurrency limit to pipeline-server

But I only made 1 PR because they would conflict on the tower version and I would have to rebase one one top of the other anyway.


tools/src/bin/web-server.rs
```diff
@@ -379,10 +355,51 @@ fn main() {
...
+
+        let handler = async |req: http::Request<_>| {
+            if req.method() != Method::GET {
+                return http::Response::builder()
+                    .status(StatusCode::METHOD_NOT_ALLOWED)
+                    .body("Invalid method".to_string());
+            }
+
+            let response = {
+                let _ = SEMAPHORE.acquire().await.unwrap();
```
asutherland, line 376: I believe this doesn't work quite right; from [the rust book section on unused variables as part of the pattern matching stuff](https://doc.rust-lang.org/book/ch19-03-pattern-syntax.html?utm_source=chatgpt.com#an-unused-variable-by-starting-its-name-with-_):
> Note that there is a subtle difference between using only _ and using a name that starts with an underscore. The syntax _x still binds the value to the variable, whereas _ doesn’t bind at all. To show a case where this distinction matters, Listing 19-21 will provide us with an error.

That is, we acquire the semaphore and immediately drop it.

But if we rename `_` to be something like `_permit` (since [the method returns a SemaphorePermit](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#method.acquire)) then this will drop at the end of the block like we expect.

I'll put up a fixup patch for this and land it after this PR since I believe you're away and these are nice changes.

tools/src/bin/web-server.rs
```diff
@@ -379,10 +355,51 @@ fn main() {
...
+                let cfg = cfg.clone();
+                let ident_map = ident_map.clone();
+                tokio::task::spawn_blocking(move || {
+                    let path = req.uri().path();
+                    handle(&cfg, &ident_map, WebRequest { path })
+                })
+                .await
+                .unwrap()
```
asutherland, line 384: So part of why we hadn't really updated hyper ever is because of https://bugzilla.mozilla.org/show_bug.cgi?id=1423345 "Rust panics keep bringing down the server" where I had updated the version of hyper and suddenly things which had not taken down the web-server started taking down the web-server.  https://bugzilla.mozilla.org/show_bug.cgi?id=1422729#c2 had some context; basically we had a case where we did straight-up call panic which I think is gone now (format.rs will panic if you try and give it something binary but its callers check and avoid that), but where we are sometimes cavalier with unwrap and that can cause problems.

I downloaded the patch and ran locally and added a new "REPO/panic" endpoint that would explicitly panic to see if it would take down the web-server, and indeed it does if we have this `unwrap()` here.  :arai's very nice reboot script then kicks in, but it's preferable to not crash at all, and so we can do that if we add a match block that turns join errors into internal error responses.

As with the above, I'll land this and then put up a PR with my changes to land, although I will strip out the `/panic` endpoint since I don't really see upsides to having that in production and once we have this fix in place we're unlikely to regress.

asutherland (approved):
Thanks!  As noted in the action items, there are 2 small changes to make here that I'm going to make and land in a follow-up PR.  My changes are fairly straightforward and the panicking is a concern so I'm going to land without additional review but I am happy for post-landing review (or if someone is very fast and responds after seeing this intention here).

tools/src/bin/web-server.rs
asutherland: Fixed in https://github.com/mozsearch/mozsearch/commit/77d4abe3a7e4ef2c374422f78b6bd8c405a0e705

tools/src/bin/web-server.rs
asutherland: Fixed in https://github.com/mozsearch/mozsearch/commit/a256a8d40415c3b1992861164ead2c241e9e0b60

PR #1113, opened 2026-06-18
 Regression test for Clang 21 crash related to require expressions

nicolas-guichard:
This tests the workaround added in https://github.com/mozsearch/mozsearch/pull/1092.

Also, since this required bumping the C++ version, I found out about [bug 2048859](https://bugzilla.mozilla.org/show_bug.cgi?id=2048859) which this fixes as well.


tests/tests/checks/snapshots/analysis/cpp/ForwardingTemplates.cpp/main@TypeDependentNewInMethodReportedAtCallSite.snap
nicolas-guichard, line 1: It seems the constructor forwarding detection stopped working for `std::vector::emplace_back` when enabling C++ 20+. Indeed https://searchfox.org/firefox-main/search?q=symbol:_ZN5angle15ObserverBindingC1EPNS_17ObserverInterfaceEy&redirect=false doesn't find https://searchfox.org/firefox-main/source/gfx/angle/checkout/src/libANGLE/Context.cpp#525. To be investigated.

asutherland (approved):
Restating:
- The extra UnresolvedLookupExpr logic helps with overload sets for constructors where the selection of which constructor to use is delayed; we want the location of the name (which comes from the super `OverloadExpr`) not the expression loc fallback.
- Upgrades our tests to build with c++26 which gets us all kinds of fun improvements to semantic analysis, including more things being allowed as constexpr, which means more "constexpr" props.  Note that we are gating that off of [clang::FunctionDecl::isConstexpr](https://searchfox.org/llvm/rev/e0cc08dc1121126392df2b832ee060a60cca6dc3/clang/include/clang/AST/Decl.h#2488) (which covers all non-Unspecified values) where we just emit "constrexpr" and is not using [clang::FunctionDecl::getConstexprKind](https://searchfox.org/llvm/rev/e0cc08dc1121126392df2b832ee060a60cca6dc3/clang/include/clang/AST/Decl.h#2494) or mapping the specific [clang::ConstexprSpecKind](https://searchfox.org/llvm/rev/e0cc08dc1121126392df2b832ee060a60cca6dc3/clang/include/clang/Basic/Specifiers.h#36) into differing strings.
- Adds the missing crashtest coverage for the "require" expression that crashed us previously, but we we don't currently extract anything semantically from it.

asutherland:
I'm going to land this now on mozsearch to minimize the number of outstanding mozsearch stacks; the indexer change will still want to land on firefox-main and so I've marked the bug leave-open.  (I'm not planning to land it there right now, it can probably wait until Nicolas' return but there's also no reason not to land it.)

PR #1115, opened 2026-06-22
Have docker bind IPv4 to avoid localhost failures

asutherland:
Without this, on my Ubuntu machine using podman symlinked as docker, browsing to http://localhost:16995/ with Firefox would return an error because it would apparently try and use the IPv6 loopback.  The same thing would happen with curl (but not if explicitly specifying 127.0.0.1).

Note that especially under podman, the container will tend to exist, so we usually end up just using the existing container and so this may not meaningfully take effect without a fresh build-docker.sh or explicitly removing the existing container.

### DIFF
--- a/run-docker.sh
+++ b/run-docker.sh
@@ -35,6 +35,12 @@ THIS_DIR=$(pwd)
 # For consistency we mount the source dir at /vagrant still
 INSIDE_CONTAINER_DIR=/vagrant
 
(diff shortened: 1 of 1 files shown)


asutherland:
I'm going to land this now because I think this is low-risk and is useful / safe (we don't bind nginx to an IPv6 port; the way it works on AWS is that the ELB only talks to us on IPv4).  Happy to discuss or iterate further changes if this breaks anything for anyone.

PR #1116, opened 2026-06-26
Bug 2046479 - Add firefox-esr153 to the tree list.

arai-a:
for [bug 2046479](https://bugzilla.mozilla.org/show_bug.cgi?id=2046479)

This does:
  * Add firefox-esr153 to the tree list

depends on https://github.com/mozsearch/mozsearch-mozilla/pull/322


asutherland (approved):
Thanks!

PR #1117, opened 2026-07-01
Align clang-plugin minimal versions with ESR115.

nicolas-guichard:
This drops support for C++ < 17 (requiring std::make_unique and std::optional), and Clang < 16 which are the versions we use for building ESR 115.
It also stops using our vendored HeuristicResolver with Clang 20+.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ addons
 !/tree-configs/README.md
 .direnv
 .envrc
+/result*
 mozsearch-firefox*
 /tests/tests/checks/target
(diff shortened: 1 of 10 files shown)


asutherland (approved):
Thanks!  Note that I hadn't landed your last set of changes into firefox-main so you may have those changes to land too.

nicolas-guichard:
I was actually looking into that and found out that there were some changes in firefox-main that we didn't have here, so I added them to this PR and when it's merged I'll sync it to firefox-main.

PR #1119, opened 2026-07-07
aws docs: update mount point from /index to /index-ebs

nicolas-guichard:
This fixes the inconsistency with the following command which references /index-ebs/interrupted.


asutherland (approved):
Thanks!

PR #1123, opened 2026-07-09
Bug 2053875 -  Parse binding information from SCIP symbol documentation

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=2053875

This tries to parse binding information from the SCIP symbol docstring:
- binding_to(language, kind, symbol) populates the slot_owner field
- bound_as(language, kind, symbol) populates the binding_slots field

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -64,11 +64,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1780331254,
-        "narHash": "sha256-ZgicGl13cOdZSxb9dLZl1k9XB44oKd5jo6Mjv64lMw4=",
+        "lastModified": 1783513756,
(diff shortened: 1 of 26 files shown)


asutherland (approved):
Thanks!

Restating:
- The binding slot `owner_lang` is not used here and that's fine/appropriate at least for now.  It was introduced for the WebIDL C++ bindings which as compared with XPIDL/XPCOM intentionally don't use C++ inheritance/virtual methods and as compared with JS XPConnect don't hide the underlying binding magic, and the underlying C++ binding impl is in fact quite interesting, so we do have 2 concrete distinct C++ symbols to link in that case.  But not currently here because we only have 1 binding in the lang.

PR #1124, opened 2026-07-09
Bug 1740290 - Make {XP,Web}IDL→js-analyze symbol linking optional

nicolas-guichard:
Bug 2053875 added support for filling the slot owner of TypeScript symbols on the scip-indexer side, and in PR #1134 I'm adding the connection between fully-qualified TypeScript symbols and their matching (but ambiguous) js-analyze symbol. Therefore, once we enable scip-typescript indexing, the direct relation between IDL symbols and js-analyze symbols becomes useless noise and should be disabled.

Bug 2053134 added annotation to autogenerated C++ bindings for XPIDL, so I'm covering them as well here. C++ binding for WebIDL are more complex because the generated binding glue calls the non-generated impl (whereas XPIDL generates a pure-virtual interface that the non-generated impl inherits from), so they are not covered by in-tree annotations yet.


scripts/idl-analyze.py
asutherland: It would be great if you could update this block comment, where update at this point probably could just be starting over from scratch because with the changes you're making to the C++ stuff here and imminently with the scip-typescript changes I don't think any of this is necessary anymore.  (In particular, we don't need the structured binding history/web-server integration progress/notes-to-self anymore.)

asutherland (commented):
This is a great cleanup!

One concern is we can only land this as-is if we've made sure to uplift the firefox-main clang and XPIDL changes to every branch we semantically index unless we do something to fork idl-analyze/condition its behavior on the tree since I don't think we can justify any regressions in our magic there.

nicolas-guichard:
As you pointed out, I've reworked this to make it optional so we don't break non-main branches.

And while my draft only covered C++, my main target was actually TypeScript all along so I've reflected this in the commit/PR message.

asutherland (approved):
Thanks!

PR #1126, opened 2026-07-17
web-server: spawn a task per incoming connection

nicolas-guichard:
When upgrading hyper, I looked at [examples/state.rs](https://github.com/hyperium/hyper/blob/540fff9180ce47ee5fab01b6cc2126eb6c286eda/examples/state.rs) which does not spawn a task for each connection and therefore only handles one at a time.

Looking at [examples/hello.rs](https://github.com/hyperium/hyper/blob/540fff9180ce47ee5fab01b6cc2126eb6c286eda/examples/hello.rs), we see that hyper does not automatically spawn tasks for us.

### DIFF
--- a/tools/src/bin/web-server.rs
+++ b/tools/src/bin/web-server.rs
@@ -365,56 +365,60 @@ async fn main() {
         let (stream, _) = server.accept().await.unwrap();
         let io = TokioIo::new(stream);
 
-        let handler = async |req: http::Request<_>| {
-            if req.method() != Method::GET {
-                return http::Response::builder()
-                    .status(StatusCode::METHOD_NOT_ALLOWED)
-                    .body("Invalid method".to_string());
-            }
-
-            let response = {
-                let _permit = SEMAPHORE.acquire().await.unwrap();
-                let cfg = cfg.clone();
-                let ident_map = ident_map.clone();
-                match tokio::task::spawn_blocking(move || {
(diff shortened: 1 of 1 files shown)


asutherland (approved):
Thanks!

Restating:
- This explains why we only saw a single core being seemingly being saturated in the most recent /rev/ permalink endpoint scraping attack that I had looked into.  The post-upgrade hyper framework nicely isn't hiding anything under magic (compared to the [pre-upgrade handle_threads](https://searchfox.org/mozsearch/rev/d966cd4f2442b61486fb1b72b278c9972163d27d/tools/src/bin/web-server.rs#387) indirection); we can see the bind, the accept loop, etc.
- `spawn_blocking` is still appropriate because we still expect to engage in sync-blocking I/O.  And although [we could cap the number of blocking threads spawned when configuring the tokio builder](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.max_blocking_threads) as a hacky alternative to the semaphore code that's now being used, the blocking workers are used for tons of important stuff and that would really be a footgun if we ever tried to integrate this web-server with any other tokio-based logic (like the pipeline-server).
- All the clones are of Arc and are fine and fast.
- Because we have nginx in front of this we don't have to worry about the tasks stacking up waiting on the semaphore.  Additionally, we should now have a WAF helping too.

PR #1129, opened 2026-08-05
Bug 2059684 - Keep exact hit counts for the last 5 revisions of the coverage repository.

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=2059684

---

This adds the ability for the last commits of the coverage repo to keep
exact hit counts instead of the log10(hits + 1) we've been using since
https://github.com/mozsearch/mozsearch/commit/fa6befb722b6f57f77de450ba3fc840c1e17d303.

In the coverage repo, each branch will now have a tag named
last-quantized/{branch} to point at the last revision which uses the
log10(hits + 1) transform. Commits at the tip of the branch
(descendants of that tag) will have the marker “exact” in their
description.

When importing a report with approximated hit counts, we first reset
the branch to last-quantized/{branch} to ignore past commits with exact
hit counts.

When generating the HTML for the coverage strip, we check whether we
have exact hit counts or now from the commit message. If so we add the
cov-exact class.

blame.js then checks that class to determine if it should display
“hit between {10^x} and {10^(x+1)} times” or just “hit x times”.

The code supports migrating from coverage repos which don't have the
last-quantized/{branch} tags yet.

### DIFF
--- a/static/js/blame.js
+++ b/static/js/blame.js
@@ -203,6 +203,25 @@ var BlamePopup = new (class BlamePopup {
     }
   }
 
+  withSiMultiplier(number) {
+    const multipliers = [
+      [18, "exa",  "E"],
+      [15, "peta", "P"],
+      [12, "tera", "T"],
+      [9,  "giga", "G"],
+      [6,  "mega", "M"],
+      [3,  "kilo", "k"],
+    ];
+
+    for (const [exponent, name, symbol] of multipliers) {
+      const threshold = 10**exponent;
+      if (number >= threshold) {
+        return `${number / threshold}&#x202F;<abbr title="${name}">${symbol}</abbr>`;
+      }
+    }
+    return `${number}`;
+  }
+
   async generateCoverageContent(elt) {
     let content = "<div>";
 
@@ -227,10 +246,11 @@ var BlamePopup = new (class BlamePopup {
         content += `This line was not hit during test runs.`;
       } else {
         if (elt.classList.contains("cov-exact")) {
-          content += `This line was hit ${x} times per coverage instrumentation.`;
+          const hitCount = x.toLocaleString();
+          content += `This line was hit ${hitCount} times per coverage instrumentation.`;
         } else {
-          const hitCountMin = 10 ** (x - 1);
-          const hitCountMax = 10 ** x;
+          const hitCountMin = this.withSiMultiplier(10 ** (x - 1));
+          const hitCountMax = this.withSiMultiplier(10 ** x);
           content +=
             `This line was hit between ${hitCountMin} and ${hitCountMax} times ` +
             `per coverage instrumentation.`;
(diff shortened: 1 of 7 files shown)


static/js/blame.js
```diff
@@ -222,15 +222,19 @@ var BlamePopup = new (class BlamePopup {
...
-        content +=
-          `This line was hit between ${hitCountMin} and ${hitCountMax} times ` +
-          `per coverage instrumentation.`;
+        if (elt.classList.contains("cov-exact")) {
+          content += `This line was hit ${x} times per coverage instrumentation.`;
+        } else {
+          const hitCountMin = 10 ** (x - 1);
+          const hitCountMax = 10 ** x;
+          content +=
+            `This line was hit between ${hitCountMin} and ${hitCountMax} times ` +
```
iainireland, line 255: Drive-by comment: would it also be possible to add some additional formatting here? Distinguishing between, eg, "100000000 and 1000000000" vs "1000000000 and 10000000000" times is not easy. Maybe "100,000,000 and 1,000,000,000" is better? Or "100 million vs 1 billion"?

nicolas-guichard: Good point!

I've settled with:
- toLocaleString for exact hit counts
  “This line was hit 123 456 789 times per coverage instrumentation.” in my fr-FR locale
- replacing the many zeros with SI multipliers for log10-based hit counts
  “This line was hit between 100 M and 1 G times per coverage instrumentation.”

tools/src/bin/codecov2git.rs
```diff
@@ -60,13 +65,26 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
         .success();
 
+    if existing_branch && !metadata.exact {
+        Command::new("git")
+            .args([
+                "reset",
+                &format!("refs/heads/{branch}"),
+                &last_quantized_ref(&branch),
+            ])
+            .arg(&args.output_repo)
```
asutherland: Is this meant to have been `current_dir` rather than `arg`?

asutherland (approved):
Thanks!  Note there may be one change needed unless I'm misunderstanding (see line item).

Restating:
- Branch/tag structure is expected to look like this with 5 exact revs:
```
Q ── Q ── … ── Q ── E ── E ── E ── E ── E   ← refs/heads/{branch} is-a branch
               ↑
       refs/tags/last-quantized/{branch} is-a tag
```
- Our reverse mapping refs look like `refs/tags/reverse/{branch}/{firefox_git_rev}` and we don't have to be concerned about orphaned trees referenced by these refs because we expect an exact tree to be subsequently replaced by a quantized tree and for the ref to therefore be updated to a new tree that's on the branch resulting in the obsolete tree being left behind (and as [mentioned in comment 2 on the bug we use bundles not tarballs](https://bugzilla.mozilla.org/show_bug.cgi?id=2059684#c2)).

tools/src/bin/codecov2git.rs
nicolas-guichard: Right, it should even have been a `git branch -f` instead of a `git reset`.

I've changed the logic a bit to be closer* to what the _gcloud variant does (which is the one actually used in prod, the bare `codecov2git` one is only used in tests now).

*the difference is that `codecov2git` only handles 1 revision at a times, so when it is called with `--exact` multiple times it shouldn't throw away earlier exact revisions as eagerly.

PR #1130, opened 2026-08-10
Update to Clang 22

nicolas-guichard:
https://phabricator.services.mozilla.com/D316715 prompted me to look into updating to Clang 22 here as well. I identified a couple of issues related to qualified names. I will propagate these fixes back to the in-tree version once that's in.


asutherland (approved):
Thanks!  Definitely a lot of potentially subtle/bad regressions without these fixes!

Given there have been non-trivial semantic changes for clang 21 and clang 22 that weren't addressed just by making compilation work as for the firefox-main patches, I wonder what the best option is for us to be prepared for clang 23 and subsequent releases.  Like is there an easy way for us to have an automatic weekly run against the clang development branch?  Or just you put something in a calendar? ;)

nicolas-guichard:
I think manually tracking upcoming releases (and trying to beat Glandium to it ^^) is better than an automated test run against clang-next. I wouldn't want to lose time updating the plugin to some WIP upstream changes that end up superseded or reverted.

For Kotlin RyanVM has been pinging me regularly about upcoming releases and I think that works well enough.
Which reminds me Kotlin 2.4.20-RC﻿ came out yesterday, I should look into that!

PR #1132, opened 2026-08-13
Bug 2063286 - Move multiple overrides to submenu.

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=2063286

The previous logic was:
- no overrides: no context menu entries
- 1 override: 1 “Sole Override” context menu entry
- 2 overrides: 2 “Override” context menu entry
- more overrides: no context menu entries

The new logic is:
- no overrides: no context menu entries
- 1 override: 1 “Sole Override” context menu entry
- 2 to MAX_OVERRIDEN_BY_LINKS overrides: a submenu lists them all
- more overrides: a submenu lists MAX_OVERRIDEN_BY_LINKS and provides a link to the full search

### DIFF
--- a/static/js/context-menu.js
+++ b/static/js/context-menu.js
@@ -1606,46 +1606,81 @@ var ContextMenu = new (class ContextMenu extends ContextMenuOrSubMenu {
         // Helper for cases like showing the recv def when the user is clicking
         // on a call to its send, but where we don't want to crowd the context
         // menu with the decl.
-        const directDefJumpify = (jumpref, pretty) => {
-          if (!jumpref.jumps) {
-            return;
-          }
-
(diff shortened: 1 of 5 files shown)


nicolas-guichard:
image

This is what it currently looks like.

asutherland (approved):
This is fantastic, thank you!  I was just testing this on the recent kdab push and the ergonomics of this seem great (plus it was also easy to be confident that they would be good even as we were discussing things thanks to @arai-a's previous enhancements to give us these submenus and for the possible IDL defs use-case!).

(And many thanks to @arai-a for having created the great front-end testing infra that lets us have the excellent test coverage here!)

PR #1133, opened 2026-08-14
Bug 2063622 - Early-return when Td is null in VisitTemplateSpecializationTypeLoc.

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=2063622

With Clang 21, we don't reach VisitTemplateSpecializationTypeLoc when visiting a dependent template name. With Clang 22 we do, but we don't have enough information for getTemplateName to return an actual value.

In both cases: we never revisit this dependent location, so we do not emit analysis records for them. This could probably be improved but let's just fix the crash first.


asutherland (approved):
Thanks!

PR #1134, opened 2026-08-19
Bug 1740290 - Cross-link scip-typescript and js-analyze symbols. 

nicolas-guichard:
Bugzilla URL: https://bugzilla.mozilla.org/show_bug.cgi?id=1740290

scip-typescript won't entirely replace js-analyze, it can't index inline JavaScript in HTML and many other plain JavaScript files such as web platform tests.

This tries to establish a bidirectional connection between js-analyze and scip-typescript symbols by taing inspiration from the js-analyze and IDL support:
- when emitting the AnalysisStructured of a scip-typescript symbol, we add a js_sym member which holds the expected js-analyze symbol
- then crossref takes those to create the reverse link in js-analyze symbols, just like it already does for IDL links
- those are forwarded to jumpref
- the context menu builds a “Possible TypeScript definitions” submenu just like it does for “Possible IDL definitions”

For the other way around, router.py traverses the new crossref field `AnalysisStructured::js_sym` to add potential js-analyze symbol usage to scip-typescript symbol results. They are added to a new “Possible JavaScript Uses” section.


nicolas-guichard:
This is now deployed on kdab.searchfox.org. Examples:
- https://kdab.searchfox.org/firefox-main/search?q=symbol:S_js_typescript_generated%2Flib.gecko.dom.d.ts%2FIDBCursor%23continuePrimaryKey().&redirect=false
- Context menu on `continuePrimaryKey` in https://kdab.searchfox.org/firefox-main/source/testing/web-platform/tests/IndexedDB/idbcursor-continuePrimaryKey-exception-order.any.js#42

tools/src/bin/crossref.rs
```diff
@@ -1210,9 +1215,14 @@ fn create_js_idl_table(meta_table: &MetaTable) -> JSIDLTable {
                 js_idl_table.insert(js_sym, vec![idl_sym]);
             }
         }
+
+        if let Some(js_sym) = meta.js_sym {
+            let ts_sym = meta.sym;
+            js_ts_table.entry(js_sym).or_default().push(ts_sym);
```
asutherland: I think we need some kind of limiting/saturation logic like we have for the JS idl-syms above.  For [this init() in aboutDialog.js](https://kdab.searchfox.org/firefox-main/rev/a4d4f7ecfae304e10b0f6724595f9c931d7c919b/browser/base/content/aboutDialog.js#184) which was the first hit for [searching on the symbol #init](https://kdab.searchfox.org/firefox-main/search?q=symbol%3A%23init&path=&case=false&regexp=false) the `ts_syms` array has 394 entries and we try and show a context submenu for all of them, although none of them are [for the definition earlier in the file which is the right answer](https://kdab.searchfox.org/firefox-main/rev/a4d4f7ecfae304e10b0f6724595f9c931d7c919b/browser/base/content/aboutDialog.js#27).

tools/src/file_format/jumpref.rs
```diff
@@ -98,6 +102,7 @@ pub fn convert_crossref_value_to_sym_info_rep(
         jumps,
         meta: cross_val.meta,
         idl_syms: cross_val.idl_syms,
+        ts_syms: cross_val.ts_syms,
```
asutherland, line 105: So while I think we probably do want crossref.rs to have accumulated all of these for the router.py expansion use-case since it provides essential glue between the two representation domains, I think we probably want to be applying a limiting heuristic for jumprefs here.

For [this init() in aboutDialog.js](https://kdab.searchfox.org/firefox-main/rev/a4d4f7ecfae304e10b0f6724595f9c931d7c919b/browser/base/content/aboutDialog.js#184) which was the first hit for [searching on the symbol #init](https://kdab.searchfox.org/firefox-main/search?q=symbol%3A%23init&path=&case=false&regexp=false) the `ts_syms` array has 394 entries and we try and show a context submenu that provides a go-to-def and a search for all of them, although none of them are [for the definition earlier in the file which is the right answer](https://kdab.searchfox.org/firefox-main/rev/a4d4f7ecfae304e10b0f6724595f9c931d7c919b/browser/base/content/aboutDialog.js#27).

I think the only real option we can provide to the user is "search for `symbol:#init`" which we already provide as the search option, but arguably there is value in conveying that there were simply too many possible definitions.  While we could accomplish this by continuing to pass-through the entirety of the `ts_syms` and just decide in "context-menu.js", I would argue that the unbounded nature of the `ts_syms` list and the low value of the information when the list gets large means that we should not propagate the entirety of the list.

I think reasonable choices here are:
- Have an empty list here (which then won't be serialized); it's ambiguous why we don't mention typescript stuff, but it avoids doing one of the below that's uglier...
- Add something like an "extra" dict to the jumpref where we could add `ts_syms_overflow: 394` 
- Have the type of `ts_syms` be weird where it can be a number in the overflow case and an array in the non-overflow case.

asutherland (approved):
This is really great and pragmatic, thanks.

I think for the extra_syms propagation we are missing an extra layer of piercing that I think we definitely want.  Right now on kdab [if I click on openCursor here](https://kdab.searchfox.org/firefox-main/rev/a4d4f7ecfae304e10b0f6724595f9c931d7c919b/testing/web-platform/tests/IndexedDB/idbcursor-continuePrimaryKey-exception-order.any.js#30) I get a link here to [openCursor in lib.gecko.dom.d.ts](https://kdab.searchfox.org/firefox-main/rev/a4d4f7ecfae304e10b0f6724595f9c931d7c919b/tools/%40types/generated/lib.gecko.dom.d.ts#21562), but if I click on that openCursor in the .d.ts I properly get a go-to def of [IDBIndex::openCursor](https://kdab.searchfox.org/firefox-main/rev/a4d4f7ecfae304e10b0f6724595f9c931d7c919b/dom/webidl/IDBIndex.webidl#69).  We should be able to go directly there.

tools/src/file_format/jumpref.rs
nicolas-guichard: > Have the type of ts_syms be weird where it can be a number in the overflow case and an array in the non-overflow case.

I don't think that's ugly with a nice Rust type to model this, so I went with this. It's not displayed in the context menu though.

nicolas-guichard:
> I think for the extra_syms propagation we are missing an extra layer of piercing that I think we definitely want.

I've added links to “possible IDL definitions” when we have “possible Typescript definitions”:
image

It's weird to have them separated by the search entries, do you think I should move “Possible TypeScript definitions” down or “Possible IDL definitions” up?
IMO it would make sense to have all definitions on top.

asutherland:
I think the test case in the screenshot is somewhat misleading in that tsClass is unambiguously the right answer, but in general I don't think we expect this heuristic to be that good or useful.  I think @arai-a's original rationale in https://bugzilla.mozilla.org/show_bug.cgi?id=2011660#c0 was accurate for the IDL case then and still holds for the new Possible TS defs now as I think about it more (very happy to hear your opinion in this specific case too, @arai-a !!):
> For JS symbols which matches the WebIDL properties, we show them, but given that they're just a guess, it's better making them less prominent than other items such as regular search.

Specifically, I think the right answer is probably to move the Possible TS Defs down to be below Possible IDL definitions.  That combines the "this is a limited heuristic" downward displacement with our ordering preferences from [jump-menu building of IDL, (glean), def, decl](https://searchfox.org/mozsearch/rev/75f110e1e270631f700cc408ee5aaedb5dcd2ec3/static/js/context-menu.js#1540-1576) which largely match router.py's [router.router::SearchResults::key_precedences](https://searchfox.org/mozsearch/rev/75f110e1e270631f700cc408ee5aaedb5dcd2ec3/router/router.py#289-293) would apply.

This helps too if we're not going to pierce the generated .d.ts symbols in the possible TS defs since people will get to the IDL heuristic first, click that, and be done.

Part of my rationale is that I think it's fine for people to want our non-TS JS processing to be improved and to be motivated to help improve the situation, either by directly letting the typescript server understand those files, or by helping searchfox do custom stuff.  In https://bugzilla.mozilla.org/show_bug.cgi?id=1775130 I'd raised the fact that in making eslint happy with the globals in the files we can't currently typescript we had created the situation where we basically already know what type of globals they have and we could potentially use that to spoon-feed the typescript server enough info for us to be able to process those files.

For example, preprocessing the files to a dual rep which typescript can process and then we could map back to the original source file.  That potentially works well with the web-platform-test ".any.js" files where the file is loaded under multiple globals where we could create a separate transformed rep for each global it can run in and then merge them.

So having the menu feel wrong to people is fine and/or would help them notice when we move to being able to have typescript level data for the file.

nicolas-guichard:
Works for me :+1: 
I've moves them below:
image

I'm going to push that to kdab.searchfox.org for a sanity check, then will merge unless you scream :wink:

asutherland:
That sounds good, but maybe post publicly in the searchfox channel on matrix too if you're around (otherwise I can if I see it's finished) so we can give people an opportunity to help do a quick smoke test or at the very least be primed for the changes.

nicolas-guichard:
I just asked for feedback on Matrix.

I'm merging this one because it shouldn't change anything on searchfox.org. https://github.com/mozsearch/mozsearch-mozilla/pull/321 is the actual switch.

PR #1135, opened 2026-08-20
Mark generated TypeScript definitions as generated

nicolas-guichard:
As @asutherland pointed out in Matrix, this should be marked as generated.
Alternatively we could set `pathkind.generated.heuristics.dir_names = ["generated"]`, but I prefer being conservative here.

### DIFF
--- a/config_defaults/per-file-info.toml
+++ b/config_defaults/per-file-info.toml
@@ -13,6 +13,7 @@ sort_order = 3
 [pathkind.generated.heuristics]
 path_prefixes = [
     "__GENERATED__/",
+    "tools/@types/generated/",
 ]
 
 [pathkind.test]


asutherland (approved):
Thanks!  This will need a few minor changes to effect our router.py results.  In https://github.com/mozsearch/mozsearch/commit/b6d5e2737a4ad27651c30fa47d92a14248c1a95c and as explained in the commit message I set things up so we now have `path_kind` baked into the crossref output, but I didn't hook it up to router.py.  For example, here we are for the first entry in the firefox-main crossref:
> ` {"uses":[{"path":"third_party/libwebrtc/rtc_base/physical_socket_server.cc","path_kind":"Third-party code","lines":[{"lno":1604,"bounds":[8,14],"line":"int n = select(fdmax + 1, &fdsRead, &fdsWrite, nullptr, wait_time);","context":"webrtc::PhysicalSocketServer::WaitSelect","contextsym":"_ZN6webrtc20PhysicalSocketServer10WaitSelectENS_9TimeDeltaEb"}]}]}`

So instead of [using categorize_path in here router.py](https://searchfox.org/mozsearch/rev/8718548f3c1b78e6f3a8f48e299fc4650cfb284a/router/router.py#326,350):
```py
def compile_result(self, kind, qual, pathr, line_modifier):
...
    pathkind = self.categorize_path(path)
```

If we use `pathr.get('path_kind')` we are sorta okay except you'll note we've pre-humanized the string in crossref since we use the [pathkind name](https://searchfox.org/mozsearch/rev/8718548f3c1b78e6f3a8f48e299fc4650cfb284a/tools/src/file_format/repo_data_ingestion.rs#35) as opposed to the [key for its config](https://searchfox.org/mozsearch/rev/8718548f3c1b78e6f3a8f48e299fc4650cfb284a/tools/src/file_format/repo_data_ingestion.rs#30), whereas router.py just returns "thirdparty" and [search.js has a mapping here](https://searchfox.org/mozsearch/rev/8718548f3c1b78e6f3a8f48e299fc4650cfb284a/static/js/search.js#1684-1687) .  It looks like [searchfox-cli just ignores those keys entirely except for ignoring the asterisk prefixed ones](https://github.com/padenot/searchfox-cli/blob/0e37238f185dadd802b532581bbb383dde80c11a/searchfox-lib/src/search.rs#L453-L480), so we could change to pre-humanized if we wanted.  But maybe it makes sense to change the names to just be the keys and we can introduce a separate concept of "label" for the current presentation values, leaving the crossref more condensed and avoiding any change in the apparent scripting interface.

Note that the query endpoint's handling of pathkinds as implemented is currently broken which is why it does weird things [because in this liquid script we were looking for just capitalized slug-type values there](https://searchfox.org/mozsearch/source/tools/templates/query_results/pathkind_group.liquid), so the canonical change to slugs is already considered a win there.

A related concern is [there's a cop-out in the pathkind computation in crossref that needs to stop being a cop-out](https://searchfox.org/mozsearch/rev/8718548f3c1b78e6f3a8f48e299fc4650cfb284a/tools/src/bin/crossref.rs#1297,1303-1310).  Currently the [derivation logic for the files assumes the first entry is the default entry which I think gives us 'normal'](https://searchfox.org/mozsearch/rev/8718548f3c1b78e6f3a8f48e299fc4650cfb284a/tools/src/file_format/repo_data_ingestion.rs#552) but it perhaps could be reasonable to assume that files we don't know about are generated and maybe just hard-code generated there, but I don't have a strong opinion.

nicolas-guichard:
I've switched crossref to using slugs, and using the path category in router.py to using the one from crossref. I've also updated the liquid template and search.js to match.

This is currently being rolled out to kdab.searchfox.org.

asutherland:
Aha, I was just about to comment exactly to the effect of the changes you just force pushed :)

nicolas-guichard:
Yeah sorry about that :sweat_smile: 

For external observers or future us: `router.py` doesn't go through crossref for file search and full-text search, so it still needs `categorize_path` for those.

asutherland (approved):
To elaborate, it seemed like both file matches and full-text matches were missing the mapping, but I assume the fallback gets us our full-text results back as well.  (And my apologies for losing track of those in my suggested changes above; I got tunnel-vision when I realized we already were precomputing the path_kind although I had been thinking about them before that.)

I think this has scope-creeped enough that it makes sense to just stick with the current fallback approach you updated here, as I think the next escalation would be for us to include the contents of the concise-file-info stuff in the crossref database like I'd speculated about doing in a variety of comments and commit messages but we never did.  If the file info is in the crossref then we can pretty easily look it up for the file matches and fulltext matches to get the info that way.

(Alternately, we could load the aggregate concise file info JSON file in router.py just like we currently load it in the query server, although we do that there so that we can perform the file path matching against that loaded file's fully in-memory contents instead of on disk in [FileLookupMap::search_files](https://searchfox.org/mozsearch/rev/8718548f3c1b78e6f3a8f48e299fc4650cfb284a/tools/src/file_format/per_file_info.rs#67-73), but since for router.py we shell out to grep and we limit those results, the O(log n) lookups from crossref could be fine there.  Note that part of the design decision in rust there was that we potentially might be evaluating predicates beyond just the regex match.)

PR #1136, opened 2026-08-21
Move custom gcc/clang symlinks to /usr/local/bin (and remove gcc12)

nicolas-guichard:
/usr/bin is the territory of the package manager, we shouldn't be messing with it.

In the wubkat setup script we call their install-dependencies script which installs the default `clang` package and replaces our symlinks to the Clang version compatible with our compiler plugin.

Instead by installing those symlinks to /usr/local/bin, we ensure the package manager won't replace or delete them and keep higher priority in $PATH.

Also I noticed we still installed gcc-12 for out-of-date reasons, so this removes it as well.


asutherland (approved):
Thanks!

Restating:
- Our high level goal was to make sure we were using our clang of choice and we were doing it in what we thought was vaguely the general appropriate debian way (install packages, use update-alternatives to choose the default bare binary name to be the one you want), but we also never assumed anything else would install packages after provisioning.
- But then we ended up using the webkit helper scripts to make sure the VM was set up appropriately for webkit, and that does do things and it clobbered what we had done above and we do like our clang version that works, so use of local to supersede that makes more sense.

asutherland:
I am going to reprovision with this and retrigger wubkat (once I make sure there are no other outstanding patches that wubkat wants)

PR #1137, opened 2026-08-22
Print types with CleanUglifiedParameters to workaround Clang 23 infinite recursion

nicolas-guichard:
This is a workaround for an infinite recursion case that happens with Clang 23 when indexing gmock [1]. IMO it doesn't hurt either.

[1] https://github.com/llvm/llvm-project/issues/218076


asutherland (approved):
Thanks!  And as always, thanks for reporting to upstream in an actionable way with a reproducer!

PR #1138, opened 2026-09-04
Don't ask clang for the layout of invalid records (clang 23 crash)

rvandermeulen:
`emitStructuredRecordInfo` calls `ASTContext::getASTRecordLayout` for every record it emits, including declarations that failed to parse. clang 23 crashes inside the layout builder (`EmptySubobjectMap`) for such invalid declarations, where clang 22 happened to tolerate them. Firefox's clang-plugin tests contain deliberate errors and segfault the compiler in five test files when the indexer is loaded alongside the checker plugin:

```
 #5 clang::ASTContext::getASTRecordLayout(clang::RecordDecl const*) const
 #6 (anonymous namespace)::EmptySubobjectMap::EmptySubobjectMap(clang::ASTContext const&, clang::CXXRecordDecl const*)
 #7 clang::ASTContext::getASTRecordLayout(clang::RecordDecl const*) const
 #8 IndexConsumer::emitStructuredRecordInfo(llvm::json::OStream&, clang::SourceLocation, clang::RecordDecl const*, IndexConsumer::LayoutHandling)
```

The offending record in that case is `DisallowMemberArgs` from Firefox's `TestCanRunScript.cpp`, an invalid class definition. Skip structured emission for invalid records (alongside the existing dependent-type check), and skip the field-record layout lookup when the field's record is invalid.

Verified with clang 23.1.0 against the Firefox clang-plugin test suite: the five crashing files compile, and the rest are unchanged. Firefox carries the same change in-tree.


rvandermeulen:
Closing: the crashes this was chasing turned out to be the type-printer recursion from llvm/llvm-project#206041 (llvm/llvm-project#218076), which #1137 already works around. The invalid-record guard is not needed.

